PR #12, opened 2022-02-17
Fix cmake build

sjaenick:
cmake didn't work out of the box with v0.5, so I played around with it a little bit (and
learned more about cmake than I ever wanted :-)

This PR fixes building with cmake; I made -O3 the default optimization; also, AVX2
is enabled by default, but can be disabled with `cmake -DENABLE_AVX=0`.


ksahlin:
Excellent thanks. I have to learn about it too.

PR #15, opened 2022-02-21
cmake: add install target

sjaenick:
Hi @ksahlin,

I've now also added an install target to cmake; by default, `make install` would attempt
installation to /usr/local, but this can be adapted with `-DCMAKE_INSTALL_PREFIX:PATH=${PREFIX}`
when invoking cmake.


ksahlin:
Hi @sjaenick,

Great! Before I merge, I wanted to check with you the line `project(strobealign VERSION 0.5)`. Will this line need an update since the current version is 0.6? I assume that this line is some cmake labelling flag not related to installing the actual source. But nevertheless good to check with you.

sjaenick:
yes, this should be updated with each release.

PR #32, opened 2022-08-13
Add continuous integration (CI) test

marcelm:
This adds a simple GitHub Actions workflow that simply builds the project on every push or pull request. The idea is to extend this over time so that actual tests are run that execute the `strobealign` binary.

With this, I would also like to make CMake the primary method of building StrobeAlign, which is why I changed the README.  The previous installation section in the README recommends a hand-crafted `gcc` invocation, but that is not sustainable in the long run. Running `cmake` is a lot easier. The very first time, run `mkdir build && cd build && cmake ..` to create the Makefile, and then only running `make` in the `build/` directory is all that is needed to build the binary.


marcelm:
@ksahlin How shall we do this, do you want to review and merge these PRs when you think they are ok?

ksahlin:
Please take full executive power to merge pull requests. You can ping me if necessary.

PR #45, opened 2022-08-24
Factor out `create_index()` (and some other changes)

marcelm:
This moves index creation into a separate function. There’s more to be done, for example, I think the index should perhaps become a class, but I need to understand the code better before I can do that.

@johan-gson If you want, I think you could start looking into how to write the index to disk after this is merged.


johan-gson (approved):
These look good to me, but I don't fully understand the index part yet, so let's wait for Kristoffer's review.

ksahlin:
@johan-gson this factorization looks good to me. I'm ok with merging.

PR #46, opened 2022-08-25
Fix incorrect MAPQ and and add test that ensures unchanged SAM output

marcelm:
We previously tested only whether the file contains 200 records.

I’m adding this because I noticed that merging #45 changes output (MAPQ is set to 0 for many records).


marcelm:
I’ve added a commit that fixes the incorrect MAPQ values.

PR #47, opened 2022-08-26
Add const to some function calls (and refactor process_flat_vector)

marcelm:
This contains a couple of changes only related by the fact that I did them while trying to understand some functions.

Quite a few functions in strobealign use function arguments not only for passing in parameters, but also for returning values (passed by reference). This makes it a bit hard to understand what’s going on. Some functions even take multiple parameters that are passed by reference, but some of them are inputs and some of them are outputs. At the call site, one cannot see which parameters are modified and which aren’t. One improvement is to mark the inputs as `const`, which I’ve done here in some places.

### DIFF
--- a/src/index.cpp
+++ b/src/index.cpp
@@ -11,7 +11,7 @@
 #include <bitset>
 #include <climits>
 #include <inttypes.h>
-
+#include <cassert>
 
 /**********************************************************
  *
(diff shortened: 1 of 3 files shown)


ksahlin (approved):

ksahlin (approved):

ksahlin (approved):
So I now learned that std::min is as fast as ternary operation (https://stackoverflow.com/questions/33047749/performance-difference-using-stdmin-or-ternary-operator)

marcelm:
> So I now learned that std::min is as fast as ternary operation

You should learn to trust the standard library and compiler writers! Compilers are really good at optimization. Here’s one small example. Clang recognizes this code:
```
int sum(int n) {
    int result = 0;
    for (int i = 0; i < n; i++) {
      result += i;
    }
    return result;
}
```
and [transforms the function body into `return n * (n + 1) / 2`](https://godbolt.org/z/9Yxqvjffc).

It’s usually better to focus on readable, maintainable code and let the compiler do its job. (Although in the example above the simpler version is also more readable.)

PR #48, opened 2022-08-28
Feat/save index to file

johan-gson:
I did two things:
1) Collected the index into a struct.
2) Added the possibility to read/write the index from/to disk. 

I have not created test cases (I don't know how in this project) but have tested the code, it produces identical output for phix compared to previous runs.

I tested to run on the human transcriptome:
Read of reference: 4.3 s
Generation of index: 31.1 s
Writing index: 6.8 s - this is a bit slow, it is probably possible to speed it up but I left it for now.
Reading index: 3.6 s
Index size: 1.5 GB - the kallisto index for the same is 2.4 GB, somewhat the same order of magnitude, the fasta file is 350 MB.

I developed the code in Visual Studio, worked fine. I have not tested to compile with GCC, but it should work. Would be good if someone with that environment up could test it quickly.

There are three things we discovered that should be changed later at some point, but it is better to do these as separate actions:
* idx_to_acc - change this to a vector, no point in having a map with <i,data>, where i is just the index - access will be identical, i.e., index.acc_map[i]
* Remove the first uint64_t in the mers_vector - according to Kristoffer it should not be needed, which will reduce the memory taken up by the index (and file size).
* I think we should rename unsigned int etc. to uint32_t - the number of bits in types like int has traditionally been different between machine architectures to optimize speed, so the size is a bit unclear, although it is pretty safe to assume they will be 32 bits.

I can fix these things once this is approved.

### DIFF
--- a/src/index.cpp
+++ b/src/index.cpp
@@ -311,14 +311,20 @@ void read_index(st_index& index, std::string filename) {
 	index.mers_index.clear();
 	ifs.read(reinterpret_cast<char*>(&sz), sizeof(sz));
 	index.mers_index.reserve(sz);
-	//read everything in one big chunk
-	auto buf_size = sz * (sizeof(kmer_lookup::key_type) + sizeof(kmer_lookup::mapped_type));
+	//read in big chunks
+	const uint64_t chunk_size = 2 ^ 22;//4 M => chunks of ~50 MB
+	auto buf_size = std::min(sz, chunk_size) * (sizeof(kmer_lookup::key_type) + sizeof(kmer_lookup::mapped_type));
 	char* buf2 = new char[buf_size];
-	ifs.read(buf2, buf_size);
-	//Add the elements directly from the buffer
-	for (int i = 0; i < sz; ++i) {
-		auto start = buf2 + i * (sizeof(kmer_lookup::key_type) + sizeof(kmer_lookup::mapped_type));
-		index.mers_index[*reinterpret_cast<kmer_lookup::key_type*>(start)] = *reinterpret_cast<kmer_lookup::mapped_type*>(start + sizeof(kmer_lookup::key_type));
+	auto left_to_read = sz;
+	while (left_to_read > 0) {
+		auto to_read = std::min(left_to_read, chunk_size);
+		ifs.read(buf2, to_read * (sizeof(kmer_lookup::key_type) + sizeof(kmer_lookup::mapped_type)));
+		//Add the elements directly from the buffer
+		for (int i = 0; i < to_read; ++i) {
+			auto start = buf2 + i * (sizeof(kmer_lookup::key_type) + sizeof(kmer_lookup::mapped_type));
+			index.mers_index[*reinterpret_cast<kmer_lookup::key_type*>(start)] = *reinterpret_cast<kmer_lookup::mapped_type*>(start + sizeof(kmer_lookup::key_type));
+		}
+		left_to_read -= to_read;
 	}
 	delete[] buf2;
 };
--- a/src/index.cpp
+++ b/src/index.cpp
@@ -98,13 +98,11 @@ static unsigned char seq_nt4_table[256] = {
 //}
 
 
-uint64_t count_unique_elements(const mers_vector &flat_vector){
-    assert(flat_vector.size() > 0);
-    uint64_t prev_k = std::get<0>(flat_vector[0]);
-    uint64_t curr_k;
+uint64_t count_unique_elements(const hash_vector& h_vector){
+    assert(h_vector.size() > 0);
+    uint64_t prev_k = h_vector[0];
     uint64_t unique_elements = 1;
-    for (auto &t : flat_vector) {
-        curr_k = std::get<0>(t);
+    for (auto &curr_k : h_vector) {
         if (curr_k != prev_k) {
             unique_elements ++;
         }
@@ -113,9 +111,9 @@ uint64_t count_unique_elements(const mers_vector &flat_vector){
     return unique_elements;
 }
 
-unsigned int index_vector(const mers_vector &flat_vector, kmer_lookup &mers_index, float f){
+unsigned int index_vector(const hash_vector &h_vector, kmer_lookup &mers_index, float f){
 
-    std::cerr << "Flat vector size: " << flat_vector.size() << std::endl;
+    std::cerr << "Flat vector size: " << h_vector.size() << std::endl;
 //    kmer_lookup mers_index;
     unsigned int offset = 0;
     unsigned int prev_offset = 0;
@@ -126,14 +124,12 @@ unsigned int index_vector(const mers_vector &flat_vector, kmer_lookup &mers_inde
     unsigned int tot_mid_ab = 0;
     std::vector<unsigned int> strobemer_counts;
 
-    uint64_t prev_k;
-    std::tuple<uint64_t, unsigned int, int> t = flat_vector[0];
-    prev_k = std::get<0>(t);
+    uint64_t prev_k = h_vector[0];
     uint64_t curr_k;
 
-    for ( auto &t : flat_vector ) {
+    for ( auto &t : h_vector) {
 //        std::cerr << t << std::endl;
-        curr_k = std::get<0>(t);
+        curr_k = t;
         if (curr_k == prev_k){
             count ++;
         }
@@ -199,6 +195,9 @@ unsigned int index_vector(const mers_vector &flat_vector, kmer_lookup &mers_inde
 void write_index(const st_index& index, std::string filename) {
     std::ofstream ofs(filename, std::ios::binary);
 
+    //write filter_cutoff
+    ofs.write(reinterpret_cast<const char*>(&index.filter_cutoff), sizeof(index.filter_cutoff));
+
     //write ref_seqs:
     uint64_t s1 = uint64_t(index.ref_seqs.size());
     ofs.write(reinterpret_cast<char*>(&s1), sizeof(s1));
@@ -242,6 +241,9 @@ void write_index(const st_index& index, std::string filename) {
 
 void read_index(st_index& index, std::string filename) {
     std::ifstream ifs(filename, std::ios::binary);
+    //read filter_cutoff
+    ifs.read(reinterpret_cast<char*>(&index.filter_cutoff), sizeof(index.filter_cutoff));
+
     //read ref_seqs:
     index.ref_seqs.clear();
     uint64_t sz = 0;
@@ -838,12 +840,11 @@ static inline void get_next_strobe_dist_constraint(const std::vector<uint64_t> &
 //    return  kmers;
 //}
 
-mers_vector seq_to_randstrobes2(int n, int k, int w_min, int w_max, std::string &seq, int ref_index, int s, int t, uint64_t q, int max_dist)
+void seq_to_randstrobes2(ind_mers_vector& flat_vector, int n, int k, int w_min, int w_max, std::string &seq, int ref_index, int s, int t, uint64_t q, int max_dist)
 {
-    mers_vector randstrobes2;
 
     if (seq.length() < w_max) {
-        return randstrobes2;
+        return;
     }
 
     std::transform(seq.begin(), seq.end(), seq.begin(), ::toupper);
@@ -862,7 +863,7 @@ mers_vector seq_to_randstrobes2(int n, int k, int w_min, int w_max, std::string 
 
     unsigned int nr_hashes = string_hashes.size();
     if (nr_hashes == 0) {
-        return randstrobes2;
+        return;
     }
 
 //        for (unsigned int i = 0; i < seq_length; i++) {
@@ -906,7 +907,7 @@ mers_vector seq_to_randstrobes2(int n, int k, int w_min, int w_max, std::string 
         }
         else{
 //            std::cerr << randstrobes2.size() << " randstrobes generated" << '\n';
-            return randstrobes2;
+            return;
         }
 
 //        uint64_t hash_randstrobe2 = (string_hashes[i] << k) ^ strobe_hashval_next;
@@ -918,8 +919,8 @@ mers_vector seq_to_randstrobes2(int n, int k, int w_min, int w_max, std::string 
         int packed = (ref_index << 8);
 //        int offset_strobe =  seq_pos_strobe2 - seq_pos_strobe1;
         packed = packed + (seq_pos_strobe2 - seq_pos_strobe1);
-        std::tuple<uint64_t, unsigned int, int> s (hash_randstrobe2, seq_pos_strobe1, packed);
-        randstrobes2.push_back(s);
+        std::tuple<uint64_t, uint32_t, int32_t> s(hash_randstrobe2, seq_pos_strobe1, packed);
+        flat_vector.push_back(s);
 //        std::cerr << seq_pos_strobe1 << " " << seq_pos_strobe2 << std::endl;
 //        std::cerr << "FORWARD REF: " << seq_pos_strobe1 << " " << seq_pos_strobe2 << " " << hash_randstrobe2 << std::endl;
 //        std::cerr << "REFERENCE: " << seq_pos_strobe1 << " " << seq_pos_strobe2 << " " << hash_randstrobe2 << std::endl;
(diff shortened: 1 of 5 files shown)


johan-gson:
I just realized there is a weakness in the reading if the index is large - I allocate one big chunk of memory that I load once - this should be split up into smaller chunks - easy to fix, I can fix this later.

ksahlin:
Excellent johan, I had a look at the structure (and some of the details) of the implementation and it looked good to me! I will however turn to @marcelm for final predicament. I did not test the implementation.

Timings also seems reasonable for the human transcriptome. Interesting that the time of reading the index is faster than reading the simple fasta reference file(?). Time for writing the index is less important than reading the index, I guess, as the reading is what will be used in the alignment step. 

I found a small typo: "A pregenerated strobomers index file" -> "A pregenerated strobomers index file"

ksahlin:
About the weakness of "allocate one big chunk of memory" that you mention here. What is the actual implication of this weakness. Much larger peak memory? Crash? I guess the magnitude of this 'weakness' will determine the action of the pull request decided by @marcelm

johan-gson:
About the weakness - let me just fix it - It gives a memory usage peak that you really don't want if the index is large. It is easy to fix - I can commit a fix here soon.

About why reading the fasta is slower - it is read line by line (using getline), which is usually slower than reading the file in big chunks. I cannot guarantee that this is why, but it is generally something you want to avoid.

src/index.cpp
```diff
@@ -199,6 +200,134 @@ unsigned int index_vector(mers_vector &flat_vector, kmer_lookup &mers_index, flo
...
+	index.ref_lengths.resize(sz); //annoyingly, this initializes the memory to zero (which is a waste of performance), but ignore that for now
+	ifs.read(reinterpret_cast<char*>(&index.ref_lengths[0]), sz*sizeof(index.ref_lengths[0]));
+
+	//read acc_map
+	////////////////
+	//TODO: update the code to vector later
+	index.acc_map.clear();
+	ifs.read(reinterpret_cast<char*>(&sz), sizeof(sz));
+	//index.acc_map.reserve(sz);
+	char buf[1000];//potentially risky, but ascension names are usually around 10 chars
```
marcelm, line 293: spelling: ascension → accession

But it would be more accurate to just call it "sequence name" (or sth like that) anyway. Accession numbers are unique identifiers, like a GenBank accession. A FASTA id, which can be anything such as `chr15` is in general not an accession.

src/index.cpp
```diff
@@ -199,6 +200,134 @@ unsigned int index_vector(mers_vector &flat_vector, kmer_lookup &mers_index, flo
...
+		auto to_read = std::min(left_to_read, chunk_size);
+		ifs.read(buf2, to_read * (sizeof(kmer_lookup::key_type) + sizeof(kmer_lookup::mapped_type)));
+		//Add the elements directly from the buffer
+		for (int i = 0; i < to_read; ++i) {
+			auto start = buf2 + i * (sizeof(kmer_lookup::key_type) + sizeof(kmer_lookup::mapped_type));
+			index.mers_index[*reinterpret_cast<kmer_lookup::key_type*>(start)] = *reinterpret_cast<kmer_lookup::mapped_type*>(start + sizeof(kmer_lookup::key_type));
+		}
+		left_to_read -= to_read;
+	}
+	delete[] buf2;
```
marcelm, line 329: (Not super important) If `ifs.read()` fails and an exception is thrown, `buf2` isn’t deallocated. It’s better to use RAII to manage buf2.

src/main.cpp
```diff
@@ -241,7 +241,9 @@ struct CommandLineOptions {
     bool index_log { false };
     bool r_set { false };
     bool max_seed_len_set { false };
-    bool s_set { false };
+    bool s_set{ false };
+	bool only_gen_index{ false };
```
marcelm, line 245: It appears this is indented with tabs. Please use spaces in order to be consistent with the rest of the code.

marcelm (commented):
This is a great start! I’d like you to not merge this for the moment, as I have very little time this week and cannot review in full. I did take a first look and have a couple of review comments already. I also just merged #47, which now causes some merge conflicts for this PR. Sorry I cannot fix them myself this time.

Regarding the command-line interface, my thought was actually to introduce subcommands. That is, you would have `strobealign index` and `strobealign map`. With that, you can clearly separate the command-line options that influence index creation from those relevant for mapping. At the moment, you can provide `-i` to force index generation, and the program will just ignore the other options. But we can discuss this and do that in a followup PR if necessary.

Hm, sorry I haven’t gotten that far, but are the index parameters stored on disk? What happens if you, for example, use `-k 20` at index creation time but `-k 24` at mapping time?

@ksahlin
> I found a small typo: "A pregenerated strobomers index file" -> "A pregenerated strobemers index file"

You can comment directly in the source code for these types of things. Makes it easier to find the problematic line.

src/index.cpp
```diff
@@ -199,6 +200,134 @@ unsigned int index_vector(mers_vector &flat_vector, kmer_lookup &mers_index, flo
...
+	index.flat_vector.resize(sz); //annoyingly, this initializes the memory to zero (which is a waste of performance), but let's ignore that for now
+	ifs.read(reinterpret_cast<char*>(&index.flat_vector[0]), sz*sizeof(index.flat_vector[0]));
+
+	//read mers_index
+	////////////////
+	index.mers_index.clear();
+	ifs.read(reinterpret_cast<char*>(&sz), sizeof(sz));
+	index.mers_index.reserve(sz);
+	//read in big chunks
+	const uint64_t chunk_size = 2 ^ 22;//4 M => chunks of ~50 MB
```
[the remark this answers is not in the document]
johan-gson: Haha, I spent too much time in other programming languages :). Will fix.

src/index.cpp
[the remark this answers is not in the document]
johan-gson: It's just that it comes at a performance cost. But maybe it matters little. I'll test and see if it matters.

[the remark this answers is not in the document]
johan-gson: The performance difference is small, maybe 0.1 s or so, difficult to measure properly. So it is fine.

johan-gson: I couldn't find this in the code? Maybe it disappeared in some other fix?

ksahlin:
> Regarding the command-line interface, my thought was actually to introduce subcommands. That is, you would have `strobealign index` and `strobealign map`. With that, you can clearly separate the command-line options that influence index creation from those relevant for mapping. At the moment, you can provide `-i` to force index generation, and the program will just ignore the other options. But we can discuss this and do that in a followup PR if necessary.

At first this sounded like a good idea but I think I changed my mind because of: 

1) indexing is relatively fast compared to BWA based tools (roughly 4-5 mins for human from what I recall)
2) one cannot call the mapping step with parameters different from the indexing parameters (as your question below). This should probably be clearly explained in the readme when the separate indexing is in place, just as for minimap2.

So 1 and 2 means that most use cases will still be full indexing-aligning "pipeline" (i.e., `strobealign ref.fa q1.fq q2.fq`).  Also, because of point 2, the indexing step needs a chunk of the mapping parameters (related to seeding). The current implementation sneak peeks at the first 500 read lengths to set proper indexing parameters for the given reads, this cannot be done for the indexing-only step. So for separate indexing, the user has to know which read lengths they intend to map and then call the indexing with parameter `-r [rough estimate of read length]` (like `strobealign -i ref.sti -r 150 ref.fa` for 150nt reads)

Looking at the minimap2 interface, we can se that it does not separate these two steps (likely for the same reason). 

So why do we bother with implementing the indexing step? The main benefit, as I see it now, will be to throw out the 64 bit hash from the flat vector which likely leads to mem and speed improvements. Second reason is for code structure, and one can perhaps improve the index (e.g., with some smart data structures for multi k-range indexing, as here https://www.biorxiv.org/content/10.1101/2021.12.06.471436v2). So the separate indexing step makes the modularity nice for future research on this. Third reason is simply for scientific benchmarking - it is annoying that other aligners implement this but not strobealign.

> Hm, sorry I haven’t gotten that far, but are the index parameters stored on disk? What happens if you, for example, use `-k 20` at index creation time but `-k 24` at mapping time?

Mapping 'will not work' for different parameters on index and mapping, as answered above.

johan-gson:
But it may make sense then, as suggested, to save the mapping params in the index file - we could then check that they are the same as the params used in the alignment. But if the indexing is not so important, maybe we should give it low priority.

ksahlin:
Depends on low priority with respect to what. I think it's a nice feature that you have implemented this. Also, if we can add the throwing out of the 64bit hashes in the flat vector we could see some performance gains. Should also be relatively straight forward to implement compared to other 'high priority' tasks such as MAPQ optimisation and supplementary alignment reporting.

ksahlin:
Also, the checking of indexing parameters is a good idea! Didn't think of that but it should definitively be done.

src/index.cpp
```diff
@@ -199,6 +200,134 @@ unsigned int index_vector(mers_vector &flat_vector, kmer_lookup &mers_index, flo
 }
 
 
+void write_index(const st_index& index, std::string filename) {
+	std::ofstream ofs(filename, std::ios::binary);
+
+	//write ref_seqs
+	////////////////
```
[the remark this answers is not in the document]
johan-gson: We can remove them, I just think it is nice to split up the code in sections in some way (how really doesn't matter to me). Maybe a //write ref_seqs: (i.e., the colon) is sufficient.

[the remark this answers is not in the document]
marcelm: Every time I feel compelled to write some kind of "section header" above some lines of code, I know it’s time to factor that code out into a function. Just saying ...

src/index.cpp
```diff
@@ -195,6 +196,126 @@ unsigned int index_vector(const mers_vector &flat_vector, kmer_lookup &mers_inde
...
+    //read acc_map:
+    //TODO: update the code to vector later
+    index.acc_map.clear();
+    ifs.read(reinterpret_cast<char*>(&sz), sizeof(sz));
+    //index.acc_map.reserve(sz);
+    auto& acc_map = index.acc_map;
+
+    for (int i = 0; i < sz; ++i) {
+        ifs.read(reinterpret_cast<char*>(&sz2), sizeof(sz2));
+        char* buf = new char[sz2];//Comes at a slight performance cost to allocate each round in the loop.
```
marcelm, line 286: I don’t think it’s justified to worry about performance here. These are just the sequence names, right? In the worst case, that’s perhaps a couple of thousand comparatively short strings, so I don’t see how this should take more than a couple of milliseconds.

Though it would be interesting to find out whether you could read into the string directly. I don’t like the raw new/delete (manually managing memory is error prone), but leave it as it is for now if you want. We can fix it later.

marcelm:
>     2. one cannot call the mapping step with parameters different from the indexing parameters (as your question below). This should probably be clearly explained in the readme when the separate indexing is in place, just as for minimap2.

I mainly want to ensure a good user interface. If a user says "only create an index" in some way, there should be an error if alignment options are used. They should not be silently ignored. If indexing and mapping functionality is split into two subcommands, this would be straightforward: The indexing subcommand accepts indexing parameters, and the mapping subcommand accepts lots of alignment parameters. So indexing options aren’t available when mapping and vice versa.

> Mapping 'will not work' for different parameters on index and mapping, as answered above.

The question is what 'will not work' means at the moment. IMO, this must either not be possible (for example because there’s no way of providing index parameters when mapping) or at least throw a clear error message. It would be best if the required index parameters were just read from the index when a pre-generated one is used.

src/index.cpp
johan-gson: You're probably right about the sequence names, I can add a shared_ptr there as well, I generally like such solutions as well, except at the very core high performance code (which this I would guess is not as you say). Yeah, I'm not sure how to solve reading directly into the buffer. Maybe a certain string length can be allocated followed by a const cast on the c_str() output, but it's not pretty. But it matters little for performance.

ksahlin:
> I mainly want to ensure a good user interface. If a user says "only create an index" in some way, there should be an error if alignment options are used. They should not be silently ignored. If indexing and mapping functionality is split into two subcommands, this would be straightforward: The indexing subcommand accepts indexing parameters, and the mapping subcommand accepts lots of alignment parameters. So indexing options aren’t available when mapping and vice versa.

This is a solid point.  I guess the tradeoff is a good user interface vs an option to do a one command run (as for minimap2). Perhaps both can be achieved with the argparser we now use? Like `strobealign index` `strobealign align` or `strobealign idx-n-aln`?

> The question is what 'will not work' means at the moment. IMO, this must either not be possible (for example because there’s no way of providing index parameters when mapping) or at least throw a clear error message. It would be best if the required index parameters were just read from the index when a pre-generated one is used.

I meant it as in that there will be little to no correct alignments (but strobealign will likely still run without error). Most reads will probably be unmapped (depending on the magnitude of the difference in settings). @johan-gson pointed out that we could easily check for identical mapping parameters when we load the index and throw an error if they are different. I think that will be a nice addition.

src/main.cpp
```diff
@@ -406,27 +415,25 @@ void adjust_mapping_params_depending_on_read_length(mapping_params &map_param, c
     }
 }
 
-std::pair<mers_vector, kmer_lookup> create_index(mapping_params& map_param, std::vector<std::string>& ref_seqs, uint64_t total_ref_seq_size) {
+void create_index(mapping_params& map_param, std::vector<std::string>& ref_seqs, mers_vector& flat_vector, kmer_lookup& mers_index, uint64_t total_ref_seq_size) {
```
marcelm, line 418: Why did you change this to `void`?

johan-gson: I thought it was more clear this way that the data is not copied. I'm not sure what happens when you use tie like that, perhaps copying is avoided, I'm not sure, and I'm really not sure if that in that case also happens when compiling in debug mode. Sending in references to the vectors is definitely safe. But, we should probably change it to just take a reference to a index structure, I just didn't get to that.

src/index.cpp
```diff
@@ -195,6 +192,121 @@ unsigned int index_vector(const mers_vector &flat_vector, kmer_lookup &mers_inde
...
+    ifs.read(reinterpret_cast<char*>(&index.flat_vector[0]), sz*sizeof(index.flat_vector[0]));
+
+    //read mers_index:
+    index.mers_index.clear();
+    ifs.read(reinterpret_cast<char*>(&sz), sizeof(sz));
+    index.mers_index.reserve(sz);
+    //read in big chunks
+    const uint64_t chunk_size = pow(2,20);//4 M => chunks of ~10 MB - The chunk size seem not to be that important
+    auto buf_size = std::min(sz, chunk_size) * (sizeof(kmer_lookup::key_type) + sizeof(kmer_lookup::mapped_type));
+    std::shared_ptr<char> buf_ptr(new char[buf_size], std::default_delete<char[]>());
```
marcelm, line 296: I think you can use a `unique_ptr` instead (because you’re not actually sharing the pointer), which would allow you to remove the `default_delete` argument.

src/main.cpp
```diff
@@ -444,29 +451,43 @@ std::pair<mers_vector, kmer_lookup> create_index(mapping_params& map_param, std:
...
+    hash_vector h_vector;
+    flat_vector.reserve(ind_flat_vector.size());
+    h_vector.reserve(ind_flat_vector.size());
+    for (std::size_t i = 0; i < ind_flat_vector.size(); ++i) {
+        flat_vector.push_back(std::make_tuple(std::get<1>(ind_flat_vector[i]), std::get<2>(ind_flat_vector[i])));
+        h_vector.push_back(std::get<0>(ind_flat_vector[i]));
+    }
+    uint64_t unique_mers = count_unique_elements(h_vector);
+    std::cerr << "Unique strobemers: " << unique_mers << std::endl;
+    ind_flat_vector.clear(); //deallocate the vector used for sorting
```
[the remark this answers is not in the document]
johan-gson: Good point, I didn't know, I'll look into this.

src/index.cpp
```diff
@@ -728,12 +840,11 @@ static inline void get_next_strobe_dist_constraint(const std::vector<uint64_t> &
 //    return  kmers;
 //}
 
-mers_vector seq_to_randstrobes2(int n, int k, int w_min, int w_max, std::string &seq, int ref_index, int s, int t, uint64_t q, int max_dist)
+void seq_to_randstrobes2(ind_mers_vector& flat_vector, int n, int k, int w_min, int w_max, std::string &seq, int ref_index, int s, int t, uint64_t q, int max_dist)
```
marcelm, line 843: You can also revert this change? So that the function returns a `mers_vector` again?

johan-gson: In this case, however, I think there is a point with my change, although I guess it is a balance between performance and readability? The function is called in a for loop, where previously a new vector was created in each round, followed by an insert into the large vector. This gives unneccesary copying I would think, unless we rely on that the optimizer solves it? Or what do you think?

johan-gson: So, the vector sent in is not empty

marcelm: Hmm, that's right. Perhaps then it's not so easy to get around this.

ksahlin: I have no idea how the optimizer in C++ works, but I remember trying a similar thing in other functions to gain a bit of performance (like for the `find_nams*` functions not deallocating the nams or hits_fw, hits_rc  vectors). Don't remember the outcome, but they are still in the code, which hints me that it probably had a positive effect (at least not negative effect).

johan-gson: Ok, I'll leave it as it is for now then

marcelm:
I think it’s time to merge this now. I can see a couple of things that we may want to improve in subsequent PRs, but the basic feature is in place and by merging it, we can avoid merge conflicts when working on other things.

I have opened #59 for the subcommand/not subcommand discussion. I haven’t actually run the code in this PR, so I’ll do that and open separate issues if I find anything.

Issue #44, opened 2022-08-24
Replace some asserts with proper error handling


marcelm:
There’s a series of asserts in `main.cpp` that should not be asserts, but handled in a different way:

https://github.com/ksahlin/StrobeAlign/blob/e7fc8f5764abd753092d7e781fc06ff9f7128e68/src/main.cpp#L543-547
```
assert(map_param.k > 7 && "You should really not use too small strobe size!");
```
Asserts shouldn’t be used for verifying external input (one reason for this is that they are removed in release builds).

PR #50, opened 2022-09-07
Exception based error handling for parameter requirements.

PaulPyl:
I added some std::exception's to the code for the four assertions that check the mapping parameters. Maybe there are more assertions that could be replaced in a similar way? This has been the first C++ in a good number of years for me so maybe it needs some review on best practices etc.

That being said, I could compile it on my Intel MacBook Pro 16 inch (2019) model without trouble and I could get it to exit with the appropriate message when I used parameters outside the specified ranges.

I think the whole maximum seed length being map_param.max_dist + map_param.k is somewhat confusing. The requirement should probably be that max_dist is < 255 and the message should say that instead of adding k on both side of the equation.


src/main.cpp
```diff
@@ -539,12 +541,27 @@ int main (int argc, char **argv)
...
+	std::cerr << "- the maximum seed length (defined as max_dist + k from parameters -m <max_dist> and -k <k>) should not exceed 255 + k" << std::endl;
+	exit(1);
+    }
 //    assert(k <= (w/2)*w_min && "k should be smaller than (w/2)*w_min to avoid creating short strobemers");
-    assert(map_param.k > 7 && "You should really not use too small strobe size!");
-    assert(map_param.k <= 32 && "k have to be smaller than 32!");
-    assert( ( map_param.s <= map_param.k ) && " s have to be smaller or equal to k!");
-    assert( ( (map_param.k-map_param.s) % 2 == 0) && " k - s have to be an even number to create canonical syncmers. Set s to e.g., k-2, k-4, k-6, k-8.");
-    assert(map_param.max_dist <= 255 && " -m (maximum seed length have to be smaller than 255 + k.");
+//    assert(map_param.k > 7 && "You should really not use too small strobe size!");
```
marcelm, line 560: Just delete this. If we want to see how the code was before the change, we can always see it in the Git history.

src/main.cpp
```diff
@@ -539,12 +541,27 @@ int main (int argc, char **argv)
         << "O: " << opt.O << std::endl
         << "E: " << opt.E << std::endl;
 
+    std::cerr << std::endl << "Verifying parameters:" << std::endl;
+    try
```
marcelm, line 545: The current style in the code is that the `{` is on the same line as the `try`.

src/main.cpp
```diff
@@ -539,12 +541,27 @@ int main (int argc, char **argv)
...
+    {
+	map_param.verify();
+    }
+    catch(std::exception& e)
```
marcelm, line 549: There should be a space after the `catch`

src/exceptions.hpp
```diff
@@ -0,0 +1,29 @@
+//
```
marcelm, line 1: Would you mind leaving out this header? I know Kristoffer did it this way, however, it just contains redundant information. We can look up author and creation date (and much more) in the Git history. And we already know the name of the file because we opened it somehow (worse, if the file is renamed, we need to remember to update the comment).

src/exceptions.hpp~
```diff
@@ -0,0 +1,9 @@
+#include <exception>
```
marcelm, line 1: This is also a backup file.

To prevent this, create a file `~/.config/git/ignore` and add the line `*~` to it (global, not project-specific ignore patterns). But we can also add a `.gitignore` file to the StrobeAlign repo.

src/exceptions.cpp~
```diff
@@ -0,0 +1,33 @@
+#include "exceptions.hpp"
```
marcelm, line 1: You seem to have accidentally added a backup file (extension is `cpp~`)

src/exceptions.hpp
```diff
@@ -0,0 +1,29 @@
...
+// Created by Paul-Theodor Pyl on 2022-09-07
+//
+
+
+#ifndef exceptions_hpp
+#define exceptions_hpp
+
+#include <exception>
+
+class k_exception: public std::exception {
```
marcelm, line 13: I think this could be much simpler. You can just throw a `std::runtime_error` with a message. Or introduce a subclass of `std::runtime_error`, but just one is enough.

src/main.cpp
```diff
@@ -539,12 +541,27 @@ int main (int argc, char **argv)
         << "O: " << opt.O << std::endl
         << "E: " << opt.E << std::endl;
 
+    std::cerr << std::endl << "Verifying parameters:" << std::endl;
```
marcelm, line 544: This may just be here for debugging purposes, but I would suggest to remove this before merging. This just clutters the output which is already quite long (we need to clean this up at some point) and doesn’t add any information: If the parameters are fine, there’s no need to notify the user that they are, and if they aren’t, there’ll be an error message anyway.

src/main.cpp
marcelm, line 549: This is too broad: If you catch `std::exception`, you catch every type of exception of the standard library. If on error occurred unrelated to the verification, you mask that error. Introducing a common base class for the verification exceptions would allow you to be more specific here.

src/main.cpp
```diff
@@ -539,12 +541,27 @@ int main (int argc, char **argv)
...
+	map_param.verify();
+    }
+    catch(std::exception& e)
+    {
+	std::cerr << "An exception occured in parameter verification: " << e.what() << std::endl;
+	std::cerr << std::endl << "Be advised of the following requirements:" << std::endl;
+	std::cerr << "- k should be from the interval [8,32], i.e. at least 8 and at most 32" << std::endl;
+	std::cerr << "- s should be smaller or equal to k" << std::endl;
+	std::cerr << "- (k - s) should be an even number" << std::endl;
+	std::cerr << "- the maximum seed length (defined as max_dist + k from parameters -m <max_dist> and -k <k>) should not exceed 255 + k" << std::endl;
```
marcelm, line 556: The user should get a more specific error message, that is, you should print `e.what()`.

src/main.cpp
```diff
@@ -539,12 +541,27 @@ int main (int argc, char **argv)
...
+	exit(1);
```
marcelm, line 557: I’m not sure, but I think `exit()` may not clean up all resources correctly. Since this is within `main()`, you can just use `return EXIT_FAILURE`.

src/main.cpp
marcelm: Ah, and I also just read http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#e15-throw-by-value-catch-exceptions-from-a-hierarchy-by-reference, which recommends to catch const references.

marcelm:
BTW, you can write 'Closes #44` in your commit message to automatically have GitHub close issue 44 when the PR with that commit is merged. I have now linked that issue manually (by clicking on "Development" on the right of this page).

PaulPyl:
I have addressed the comments in this commit, I chose to inherit from runtime_error and got rid of the backup files etc. Also I changed the name of the pull request since that was just the initial commit message before.

src/main.cpp
```diff
@@ -539,14 +541,14 @@ int main (int argc, char **argv)
         << "O: " << opt.O << std::endl
         << "E: " << opt.E << std::endl;
 
+    try {
+	map_param.verify();
+    }
+    catch (mapping_parameter_exception& e) {
+	std::cerr << "An exception occured in parameter verification: " << e.what() << std::endl;
```
marcelm, line 548: This output is very programmer oriented. I think a better message would be sth. like: "A mapping parameter is invalid: ". A normal user doesn’t need to know that there was an exception, just what the problem is.

src/exceptions.hpp
```diff
@@ -0,0 +1,11 @@
+#ifndef exceptions_hpp
+#define exceptions_hpp
+
+#include <stdexcept>
+
+class mapping_parameter_exception: public std::runtime_error {
```
marcelm, line 6: Can you name this `BadMappingParameter` or so? See https://kevlinhenney.medium.com/exceptional-naming-6e3c8f5bffac for why I think dropping the `_exception` suffix is better.

Also, I’d like us to use CamelCase class names. I realize it’s not consistent with the rest of the code, and I haven’t brought it up, but if we do it, we might as well start here.

src/main.cpp
```diff
@@ -12,18 +12,20 @@
 #include <math.h>
 #include <inttypes.h>
 
+#include <stdexcept>
```
marcelm, line 15: This is not needed anymore, correct?

src/exceptions.hpp
```diff
@@ -0,0 +1,11 @@
+#ifndef exceptions_hpp
```
marcelm, line 1: Hm, now that this file is so small, I wonder whether it actually needs to be separate. You could just put the class into `index.hpp`. It’s an exception that is specific to the `mapping_params` struct anyway.

PaulPyl: I don't have a string opinion on it either way, only that if there would be more exceptions added later (not related to mapping parameters), it might be nicer to have them all in one place instead of attached to whichever class they pertain to (or somewhere else if they are not class specific)?

src/main.cpp
marcelm: Yes, but `main.cpp` doesn’t catch `runtime_error`, only `BadMappingParameter`. (At the moment, it doesn’t really make a difference because `main.cpp` includes `exceptions.hpp` which in turn includes `stdexcept`.)

src/exceptions.hpp
marcelm: Yeah, you’re right. I’m also undecided (both ways make sense), so just leave it where it is now.

src/main.cpp
PaulPyl: I think I misunderstood that comment to mean to remove all dependencies on that o.O

PaulPyl:
Alright, I think this might be resolved at this point.

marcelm:
I agree, a nice step towards better error handling!

PR #53, opened 2022-09-07
Remove expression that is always true

marcelm:
Am I thinking correctly? The `if (i + w_max < nr_hashes)` a couple of lines earlier already ensures that `(nr_hashes <= i + w_max)` is true if we get to the `else`.

### DIFF
--- a/src/index.cpp
+++ b/src/index.cpp
@@ -896,7 +896,7 @@ void seq_to_randstrobes2(ind_mers_vector& flat_vector, int n, int k, int w_min, 
             get_next_strobe_dist_constraint(string_hashes, pos_to_seq_choord, strobe_hash, strobe_pos_next, strobe_hashval_next, w_start, w_end, q, seq_pos_strobe1, seq_end,i);
 
         }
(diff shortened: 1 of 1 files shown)


ksahlin:
I don't know (forgot) how to comment in the code in this pull request, but the same redundant logic is present in line 892 in `seq_to_randstrobes2_read`

marcelm:
Alright, good to have a second opinion on this :-). I’ll fix the second occurrence and then merge.

marcelm:
Regarding how to comment: Just hover of the line you want to comment on and click on the plus "+" that appears.

src/index.cpp
```diff
@@ -999,14 +999,13 @@ mers_vector_read seq_to_randstrobes2_read(int n, int k, int w_min, int w_max, st
             get_next_strobe_dist_constraint(string_hashes, pos_to_seq_choord, strobe_hash, strobe_pos_next, strobe_hashval_next, w_start, w_end, q, seq_pos_strobe1, seq_end, i);
 
         }
-        else if ((i + w_min + 1 < nr_hashes) && (nr_hashes <= i + w_max) ){
```
ksahlin, line 1002: This is my test comment.

PR #55, opened 2022-09-07
Do not convert reference to uppercase in seq_to_randstrobes2

marcelm:
This should not be necessary because the seq_nt4_table array takes care of both upper and lowercase characters. This allows us to mark the `ref` parameter as `const`.

Fix me: This actually changes the output! The reported insert sizes in the SAM file are different. Very weird.


marcelm:
I’ve now verified that the index is identical even if I remove the conversion to uppercase. So I guess the changed mapping qualities are due to a lowercase reference used during local alignment.

I’ve changed the PR now in such a way that uppercase conversion still takes place, but is done before calling `seq_to_randstrobes2` (and before calling `seq_to_randstrobes2_read`). That way, the `seq` parameter can still be marked as const, which is what I wanted in the first place.

ksahlin:
oh I see. I guess it would make sense to convert it to uppercase as soon as it is read in from file then?

PR #62, opened 2022-09-08
Consts and use KSeq

marcelm:
Two changes:
* Change some functions that have separate parameters for sequence, name and qualities to have a KSeq parameter instead
* Add lots of `const`s to help document which parameters are "in" parameters


marcelm:
Ah, and this also changes `Sam::add_unmapped_pair` so that it just calls `Sam::add_unmapped` twice. Should have been in the #61 PR.

Issue #52, opened 2022-09-07
Some reads get MAPQ = 0 in the phiX test dataset


marcelm:
I opened the tiny test dataset (with phiX) alignments in IGV (top in screenshot) and compared it to what BWA-MEM outputs (bottom). Many of the phiX reads are mapped with a mapping quality of zero (these are the alignments in a lighter gray), but BWA-MEM assigns them a nonzero mapping quality.

![mapq0](https://user-images.githubusercontent.com/724864/188870903-54d4ecb9-96b5-4ff9-b9b6-38a3e7aaef25.png)

Read SRR1377138.40 is an example.

ksahlin:
Interesting, cannot say why that is. But great example to start looking into whether the MAPQ values can be better.

marcelm:
I’m going to close this because no normal reads in the phiX test dataset get a MAPQ of zero anymore, in particular not SRR1377138.40.

There is still a problem with reads unexpectedly getting MAPQ=0, but it is covered by #334.

PR #65, opened 2022-09-09
Fix a probable logic error due to a typo

marcelm:
I’m working on a PR to enable compiler warnings with `-Wall` and to get rid of the warnings. One of the warnings was a complaint that variable `r2_r1` was never used, which led me to the typo (`r1_r2 || r1_r2`. I assume it should have been `r1_r2 || r2_r`).

Interestingly, without intending to work on it at the moment, this helps with issue #52 because it changes MAPQ for some of the alignments from 0, 14 or 15 to 60. See #52.

Seeing this, I would like to enable `-Wall` (and probably `-Wextra`) permanently. We may even consider `-Werror`, which turns all warnings into errors.


ksahlin:
Wow, great find! I assume this will change the output a bit, for the better. Good friday!

ksahlin:
Well, let me be more careful in my assessment: It is definitely a bug and good catch. 

What the fix will do is to allow strobealign to output quickly more PE-read alignments that (seem to, based on seeds) have one  'clear best location' - not entering the if-statement will make strobealign consider several other candidate locations. So while the runtime may decrease a tad, the extra mapq scores of 60 may be overly confident. This may be bad. However, if they match with what BWA-MEM is outputting for MAPQ the same reads, it is not a pressing issue. If they turn out to be overly confident, we may want to consider fewer alignments ending up in this if statement - or redesign it together with MAPQ scores.

marcelm:
Merging this now. For any other fixes, we’ll need to open separate issues/PRs.

PR #69, opened 2022-09-12
Let compilation flags be determined by CMake build type

marcelm:
CMake supports something called "build types" and I suggest to use them. A build type determines which optimization and debugging flags are used. There are a couple of predefined ones:
* `Release`: `-O3 -DNDEBUG` (`NDEBUG` disables `assert`s)
* `RelWithDebInfo`: `-O2 -g -DNDEBUG` (good enough optimization for a release, but with debugging symbols)
* `Debug`: `-O0 -g`

The `Release` one is closest to what we had before and should, for example, be used when compiling for Bioconda. To set the build type, pass `-DCMAKE_BUILD_TYPE=Release` when running `cmake`. I have added this to the instructions in the README.

I tried to make it so that the default build type is `RelWithDebInfo` so one gets most of the optimizations even when forgetting to set a build type. There is usually very little difference between `-O2` and `-O3` anyway, but I haven’t measured this.


ksahlin:
Okay I don't know the impact of going from O3 'down to' O2. Might be worth to measure the impact at some point. but if one can anyways specify Release, it may be a non issue.

marcelm:
Ok, I’ll measure this before merging just in case, also because I’m interested.

marcelm:
There seems to be a small difference of perhaps 3%, but the variance of successive runs is higher than that (I didn’t close all other programs running on my machine). I have changed the default to `Release` now. That also simplifies the README again.

(I also removed the outdated notice about zlib from the README.)

PR #70, opened 2022-09-12
Less verbose default logging output

marcelm:
I think the log output is a bit "intimidating" and contains some lines that an end user doesn’t need to see. This PR silences some of these messages by default and adds a command-line option `-v` (verbose) that can be used to re-enable them.

This is achieved by introducing a very simple `Logger` class. Instead of writing to `std::cerr` unconditionally, each message is written to `logger.debug()` or `logger.info()` (`logger.warning()` and `logger.error()` are also available). Only messages at least as high as the current log level are printed, and the default log level is set to `LOG_INFO`. The `-v` option sets the current log level to `LOG_DEBUG`. We could also add a `--quiet` option later that would set the log level to `LOG_WARNING` and therefore hide even the info-level messages.


ksahlin:
great!

PR #71, opened 2022-09-13
Fix missing qualities for unmapped reads

### LINKED ISSUE #56: Unmappable reads lose their qualities in SAM output
If a read is not mappable, only its sequence is written to the SAM output, and the QUAL field is set to `*`.

marcelm:
This fixes #56. As usual, I also did some refactoring so I could work with the code better.


marcelm:
I think I merged #72 and #73 into the wrong branch, this will look a bit weird in the Git history but isn’t an actual problem.

Issue #49, opened 2022-08-29
Add unit tests (using a unit testing framework)


marcelm:
I’ve searched a bit for testing frameworks, and I think [Catch2](https://github.com/catchorg/Catch2) would be a good option. It’s also used by args (which we use), and the [tests seem to be quite readable](https://github.com/Taywee/args/blob/master/test.cxx).

PR #72, opened 2022-09-13
Add a unit testing framework

marcelm:
I went with doctest instead of Catch2 because the compile times are much lower. Also, Catch2 is in its most recent version no longer a single-header library, but doctest still is. doctest is inspired by Catch2, but is apparently even simpler to use.

Tutorial (short, recommended reading for those who want to add tests):
https://github.com/doctest/doctest/blob/master/doc/markdown/tutorial.md

I added a first test that tests the `read_references` function.

To run the tests, just build everything with `make -C build` as usual and then run `build/test-strobealign` (or `tests/run.sh`).

Closes https://github.com/ksahlin/StrobeAlign/issues/49


marcelm:
Whoops, should have merged this one first.

PR #73, opened 2022-09-13
Add a References class

marcelm:
A `References` instance stores the sequences read from a FASTA file. It has fields for the sequences, their names and their lengths. Instead of separately passing sequnce names and lengths to functions, this allows us to pass a single `references` object.

* `read_references` is now the static method `References::from_fasta` that returns a `References` instance.
* Moved sequences, names and lengths out of `st_index`. Instead, the `write_index`/`read_index` functions get a `References` object as additional parameter.
* Get rid of `total_ref_seq_size`. Instead, total reference size is computed within the `References` constructor and can be retrieved with the `total_length()` method.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,8 +23,8 @@ endif()
 
 add_compile_options(-Wall -Wextra)
 
-set(SOURCES
-  src/fasta.cpp
+add_library(salib STATIC ${SOURCES}
+  src/refs.cpp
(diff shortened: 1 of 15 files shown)


marcelm:
I’ve pushed a couple of commits that rename `fasta.hpp`/`.cpp` to `refs.hpp`/`.cpp` and do some cleanups.

PR #74, opened 2022-09-14
Move some unused code to a new unused.cpp file

marcelm:
There is lots of code in StrobeAlign that is unused. It is either commented out or it exists as a function that is never called. Since I would like to focus on the relevant code when working on the program, I would like to somehow get rid of it. I usually just delete old code, knowing that I can find it again in the Git history if the need arises.

However, I can see that you may be reluctant to do that as it appears that you sometimes run experiments by quickly commenting/uncommenting a function call (for example, to switch out the hash function). So to still have the code somewhere more easily accessible, I suggest to move it to a file "unused.cpp" for the moment. This PR adds that file and moves some unused functions from `index.cpp` into it. Additionally, this adds an `unused-code` target to CMake, which is only there to ensure that the code in `unused.cpp` is compiled when running "make". This ensures that the functions are runnable should they be needed.


src/unused.cpp
```diff
@@ -0,0 +1,291 @@
...
+//        }
+//    }
+////    std::cerr << hash_count << " values produced from string of length " <<   seq_length << std::endl;
+////    for(auto t: pos_to_seq_choord){
+////        std::cerr << t << " ";
+////    }
+////    std::cerr << " " << std::endl;
+//}
+
+static inline void make_string_to_hashvalues_closed_syncmers_canonical(std::string &seq, std::vector<uint64_t> &string_hashes, std::vector<unsigned int> &pos_to_seq_choord, uint64_t kmask, int k, uint64_t smask, int s, int t, int startpos) {
```
ksahlin, line 186: Fully agreed with this commit. However, this function is still used in the code.

marcelm: I think you mean `make_string_to_hashvalues_open_syncmers_canonical`? That one is used, but this one isn’t. There’s one mention of it here, but it’s commented out:
https://github.com/ksahlin/StrobeAlign/blob/4d906ca7ba0ec73f740825c71fb824870b238acd/src/index.cpp#L611

ksahlin: You are right! Got confused by my long function names..

marcelm:
I’ve gone through `aln.cpp` as well and moved everything I found that was unused over to `unused.cpp` as well. With this, `aln.cpp` is now down to 3000 lines (from 4000 lines when I started a couple of weeks ago).

Issue #75, opened 2022-09-15
De-duplicate `align_SE` and `align_SE_secondary_hits`


marcelm:
From a comment in `aln.cpp`:
> I created an entire new function here, duplicating a lot of the code as outputting secondary hits is has some overhead to the
> original align_SE function (storing a vector of hits and sorting them)
> Such overhead is not present in align_PE - which implements both options in the same function.

To avoid the code duplication, we can (for example) move the common stuff into helper functions.

Issue #25, opened 2022-06-01
MAPQ scores


ksahlin:
Note to developer: 

It is not clear if strobealign has optimal mapq scores for variant calling. For example, having an accurate MAPQ score is crucial for SNP calling. See this [tweet thread](https://twitter.com/denizkural/status/1531751340699013120?s=20&t=CZV2uCBtmQh5yqY7nIpztw) and [this subthread](https://twitter.com/denizkural/status/1531754106905407490?s=20&t=CZV2uCBtmQh5yqY7nIpztw).

For info, a relevant paper about this is found  [here](https://twitter.com/naechyun_chen/status/1531866777264631810?s=20&t=CZV2uCBtmQh5yqY7nIpztw).

marcelm:
Link to mentioned paper: [A tandem simulation framework for predicting mapping quality](https://genomebiology.biomedcentral.com/articles/10.1186/s13059-017-1290-3)

The current MAPQ computation function in StrobeAlign uses the formula from minimap2, see [Section 2.3.1 in the paper](https://genomebiology.biomedcentral.com/articles/10.1186/s13059-017-1290-3).

PaulPyl:
I don't see any section numbers in the linked paper, both links above point to the same one: [A tandem simulation framework for predicting mapping quality](https://genomebiology.biomedcentral.com/articles/10.1186/s13059-017-1290-3)

marcelm:
> I don't see any section numbers in the linked paper

Sorry, I shouldn’t comment so late at night – I have fixed the link now.

marcelm:
@ksahlin sent me this comment by e-mail and asked me to post it here:
> Good catch with the min_matches bug! Since our “hits” (pairs of linked seeds) are different to minimap2’s minimizer hits, me should probably not use this formula in the future. The heuristic min(M/10,1) is probably derived by Heng from testing. We will not have the same number of matches for our seeds and long reads, therefore the magic constant 10 may be off.
> 
> Just mentioning this for caution.  Again, mimicking what BWA-MEM is doing will probably be better (in case possible). Although I heard that BWA-MEM always tries a second mapping location to compute MAPQ(?). This is probably something we want to avoid for all reads, but use whenever we still have to align twice due to multiple mappings. Or in case WFA alignment is so much faster than SSW that we can “afford it”. 

(`min_matches` bug is fixed in #84).

marcelm:
Copying @PaulPyl’s comment from #79:
[...]

PR #79, opened 2022-09-15
Adressing Issue #75 with some initial refactoring, both functions now…

PaulPyl:
… call a helper that returns a reference to a vetor of alignments and align_PE simply returns the best (by SW score as it did before) while align_SE_secondary_hits does it's usual thing with the vector.

I am relatively certain that this will do the same thing as before, however since we now have unit tests, adding some unit tests that check that alignments look as they should with test data could be really good.


src/aln.cpp
```diff
@@ -850,295 +850,19 @@ static inline bool sort_highest_sw_scores_single(const std::tuple<int, alignment
...
-//                    sw_score = aln_params.match*(read_len-hamming_mod) - aln_params.mismatch*hamming_mod;
-
-                    sam_aln.cigar = cigar_string.str();
-                    best_align_dist = hamming_dist;
-                    //                sam_aln.cigar = std::to_string(read_len) + "M";
-                    sam_aln.ed = hamming_mod;
-                    sam_aln.ref_start = ref_start + soft_left +1; // +1 because SAM is 1-based!
-                    sam_aln.is_rc = is_rc;
-                    sam_aln.ref_id = n.ref_id;
-                    sam_aln.sw_score = aln_score;
```
PaulPyl, line 1050: This is a spot that is now different, since one function set sam_aln.sw_score to sw_score and the other to aln_score and I went with what from the variable names perspective makes more sense, but maybe this is intentional and in that case I need to add logic to switch between both behaviors depending on which function calls the helper.

marcelm:
Oh great, that’s a lot of duplicated code we can get rid of! The tests don’t pass, though, so something is different. It appears to be different mapping qualities. Here’s how I usually run the tests locally:
```
export PATH=$PWD/build:$PATH
make -C build && tests/run.sh
```
The first line is currently necessary; however, I think I should rewrite the `run.sh` script so that it does not require `strobealign` to be in the PATH.

Since this is just a refactor, we definitely need a test that verifies that the output is the same as before when running with `-N 2`. The easiest way to do this would be to add yet more `strobealign` invocations to `tests/run.sh`. Feel free to do so, but we I think that the number of files in `tests/` is getting too big too fast, and we need to find a better way for testing. I’ll open a separate issue about this.

BTW, feel free to use shorter branch names. IMO, they don’t need to be very descriptive because the name disappears once the PR is merged. I think it is more valuable to be able to type them in on the command line without resorting to copy&paste.

PaulPyl:
I will keep on working on this next week, I am off tomorrow and Monday. But there is a get_alignment and get_MAPQ function used by align_PE that I want to use for this, since it would make sense. However right now using those gives me the wrong mapping qualities so I will have to do dinner digging first. I think this might overlap with the issue about MAPQ being 0 sometimes (#52).

I created the branch with a button on GitHub to connect it to the issue and it set the name like that, but indeed it is a bit overkill :)

PaulPyl:
A curious observation here is also, that when I compare my SE result with the test SE result and the test PE result for the same read, the MAPQs are not identical in the test data either:

`(base) paul@Paul-Theodors-MacBook-Pro build % grep "SRR1377138.6\t" test.se.sam                                     
SRR1377138.6	0	NC_001422.1	649	60	1S184=1X115=	*	0	0	[...]
(base) paul@Paul-Theodors-MacBook-Pro build % grep "SRR1377138.6\t" ../tests/phix.pe.sam                            
SRR1377138.6	99	NC_001422.1	649	0	1S184=1X115=	=	719	-371	[...]
SRR1377138.6	147	NC_001422.1	719	0	30=1X83=1X185=1S	=	649	371	[...]
(base) paul@Paul-Theodors-MacBook-Pro build % grep "SRR1377138.6\t" ../tests/phix.se.sam
SRR1377138.6	0	NC_001422.1	649	8	1S184=1X115=	*	0	0	[...]`

So now I get a 60, the SE test data wants and 8 and the PE test data a 0, presumably the mapping quality should not differ between SE and PE mode? at least the CIGAR strings are all the same and the locations as well.

Anyway, I will poke around get_MAPQ for a bit more to see if I can find out what is happening here.

ksahlin:
Without commenting at a detailed level. I do think that MAPQ is different if there is extra confidence for the mated read being aligned consistently nearby. This is how i remember trying to implement it. I think the other aligners would do that too (think I looked at BWA mem for this), although I am not sure.

ksahlin:
Nevermind previous answer (was before coffee :). While my comment still holds, your particular set of observations indeed look suspicious - PE having lower confidence than SE. MAPQ is indeed something we should look at in detail. Your observation goes in the same trend as our [variant calling results](https://github.com/ksahlin/StrobeAlign#variant-calling-benchmark-simulated-sim3), strobealign has high precision but a bit lower sensitivity. I think more consistent MAPQ would improve these results.

PaulPyl:
I didn't have a lot of time this week to check it more clearly but I was wondering about the current implementation of MAPQ as `// MAPQ = 40(1−s2/s1) ·min{1,|M|/10} · log s1` in the `get_MAPQ` function.

Where did this definition arise? As I understand it we would want to put our best guess for the probability that the alignment is wrong in the MAPQ score. So here we take 1 minus the ratio of best to second-best alignment and scale that to 40, i.e. if the alignments are equally good then we end up with MAPQ 0. Since 0*<some other stuff>.

Now here is a fun question, if I have exactly two alignments and they are equally good and not discarded because of bad alignment score, would I assume the probability of a given alignment being wrong as 1, implying a MAPQ of 0, or would I assume the probability is 1/<#equally good alignments> and in this case 0.5 with a resulting MAPQ of around 3?

It seems like there is a bit of an issue with this, in the sense that there are no 'correct' MAPQ scores we can benchmark against, since that really depends on the algorithm and our choices.

For reference:
`-10*log10(c(0.0001,0.5,0.9999,1.0))`
`40 3.01029 0.00043 0`

ksahlin:
Not at a computer so excuse my brevity and guessing.

First, our formula is from minimap2 I believe (see paper) if my memory is correct. If not, it’s from BWA.

Good example. I think this is where MAPQ scores from BWA/Bowtie2/tophat differs. Tophat uses the p=0.5 -> MAPQ=3 while bowtie or BwA use 0 whenever multiple identical. I’m not sure what makes sense. Following the initial interpretation of MAPQ, 3 would perhaps mimic the idea of the formula better.

see paper in issue #25 for some ideas.

PaulPyl:
Thanks, I will have a look over the paper as well. Maybe we need to have a discussion at some point to define what we want to go with. It might just be doing the same as minimap2?

marcelm:
I’ll try to move the discussion over to #25 (discussion in this PR should preferably be only about the refactoring itself).

marcelm:
Paul, your most recent commit made it so that all the changes are reverted (so the PR is "empty"). What are your plans for this PR?

PaulPyl:
I commit here the changes I wanted to do to align_SE, making it use get_alignment and get_MAPQ internally. The checks fail because the MAPQ values are not identical, but since there is no real ground truth this is a bit of an arbitrary hurdle, at least until we have defined how our MAPQ calculations should be done (as we discuss in #25). Fun fact, since I merged the main branch I can't compile it on my mac anymore, I wonder if anyone else has a mac to check if this is a general problem or if I just messed up my local environment somehow o.O

src/aln.cpp
```diff
@@ -826,7 +826,7 @@ static inline bool sort_highest_sw_scores_single(const std::tuple<int, alignment
 }
 
 
-inline void align_SE(const alignment_params &aln_params, Sam& sam, std::vector<nam> &all_nams, const KSeq& record, int k, const References& references, AlignmentStatistics &statistics, float dropoff, int max_tries) {
+inline void align_SE_old(const alignment_params &aln_params, Sam& sam, std::vector<nam> &all_nams, const KSeq& record, int k, const References& references, AlignmentStatistics &statistics, float dropoff, int max_tries) {
```
PaulPyl, line 829: This should be removed before merging, but I wanted to keep it around for reference.

marcelm: As soon as you actually delete this function, the diffs here on GitHub allow us to better see what has actually changed.

ksahlin:
@PaulPyl , I am also having problems installing on mac as I reported in https://github.com/ksahlin/StrobeAlign/issues/102. I haven't tried to compile strobealign since before summer though, so cannot say whether this is a recently introduced problem.

src/aln.cpp
PaulPyl: I removed it now to not clutter up the PR.

PaulPyl:
Ok, this PR now compiles again but the checks fail because the MAPQ values are not matching anymore. This relates to #25 but I am unsure how to proceed. We should rework aln.cpp so that the align_SE function calls get_alignment and get_MAPQ and then make sure those two work as expected. For get_MAPQ we maybe need to define what "as expected" would mean.

ksahlin:
I wish I had a good solution for this. 

I think the MAPQ question is more of an open-ended research question. Do the reads seem to have a more logical MAPQ value after your fixes in this pull request? If so I wouldn't mind a merge. 

It is on my plate to implement a smaller benchmark for this (in addition to rerunning the larger benchmarks I used for the paper).

What does Marcel think?

src/aln.cpp
```diff
@@ -625,284 +625,8 @@ static inline bool sort_highest_sw_scores_single(const std::tuple<int, alignment
...
-    float s1 = n_max.score;
-    int best_align_dist = ~0U >> 1;
-    int best_align_sw_score = -1000;
-
-    bool aln_did_not_fit;
-    alignment sam_aln;
-    int mapq = 60;
-    int min_mapq_diff = best_align_dist;
-
-    for (auto &n : all_nams) {
```
marcelm, line 655: The old `align_SE` iterates over `all_nams`, but the new one doesn’t. How does that work (why is that ok)?

PaulPyl: That is a good point, I somehow ended up believing that would happen in get_alignment but that is not the case. So probably this version of the function just grabs the first entry of nams and doesn't iterate.

PaulPyl: I think I looked at align_PE and that doesn't iterate either but just uses all_nams1[0] and all_nams2[0] as the input to get_alignment.

PaulPyl:
I closed this and deleted the branch, I think that issue needs a new run at it. Maybe after addressing the MAPQ score calculation.

marcelm:
Hm, sorry; it looked so nice to get rid of this amount of code ... maybe we should try to duduplicate/refactor this in smaller steps?

Please let me know when/if you start to work on this again. I wanted to do a couple of things that touch the `align_SE` function and don’t want to cause unnecessary merge conflicts.

PaulPyl:
You can go ahead and work on align_SE for now, I will let you know if I come back to it.

PR #80, opened 2022-09-16
Enable building with WFA2-lib

marcelm:
This adds support for linking against WFA2-lib in StrobeAlign. I’ve added the C++ demo program from WFA2-lib as `src/wfademo.cpp` and it works (this should of course be removed again).

I chose to go with the technique of downloading the dependency at build time. This is nicely supported by CMake. That way, we don’t have to copy the WFA2-lib sources into StrobeAlign. I have pinned to the most recent WFA2-lib commit so that we’ll always get the same code.

On the UPPMAX cluster rackham, the default CMake and GCC versions are too old, but more recent, working versions are available. I have been able to compile StrobeAlign like this:
```
module load cmake gcc
cmake -B build
make -C build
```
(I recently found out about the `-B` option to `cmake`. It replaces the `mkdir build && cd build && cmake ..` steps.)

I am sure that the way the `CMakeLists.txt` file is structured can be improved, but I’m still learning CMake best practices.

### DIFF
--- a/src/cmdline.cpp
+++ b/src/cmdline.cpp
@@ -30,7 +30,7 @@ std::pair<CommandLineOptions, mapping_params> parse_command_line_arguments(int a
 
     args::Group seeding(parser, "Seeding:");
     //args::ValueFlag<int> n(parser, "INT", "Number of strobes [2]", {'n'});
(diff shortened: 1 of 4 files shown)


ksahlin:
great, I will test this as soon as i have time!

PR #84, opened 2022-09-22
Refactor mapq computation

marcelm:
As part of the discussion in PR #79, I tried to understand what `get_MAPQ' does. This refactors that function so I can better understand it.

There’s this line:
```
min_matches  = (float)n_max.n_hits/10 > 1 ? (float)n_max.n_hits/10 : 1;
```
It uses the ternary operator to compute a maximum, which is why I replaced this with `std::max` in this PR. However, the given formula uses `min`, an the variable name also hints that a minimum is wanted. Our tests don’t cover this, so I don’t know what is actually correct.

### DIFF
--- a/src/aln.cpp
+++ b/src/aln.cpp
@@ -1810,13 +1810,12 @@ static inline int get_MAPQ(const std::vector<nam> &all_nams, const nam &n_max) {
     if (all_nams.size() <= 1) {
         return 60;
     }
-    // MAPQ = 40(1−s2/s1) ·min{1,|M|/10} · log s1
-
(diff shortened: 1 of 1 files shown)


marcelm:
I’m sure now the maximum should have been a min and have changed it now in an additional commit.

ksahlin:
I also had a look and I agree. Great! wonder what implications it has on the map score though. Although the formula need revision anyway..

marcelm:
Ah, yes, I checked the impact, but forgot to mention it, see the comment in #25 (went from 1780 to 378) that I added just now.

PR #85, opened 2022-09-23
Avoid bit-fiddling hacks for computing min/max

marcelm:
See https://graphics.stanford.edu/~seander/bithacks.html#IntegerMinOrMax for where these hacks come from. GCC actually recognizes these and replaces them with the same code that is produced by using std::min/max, which uses conditional move instructions (which are branchless).

See https://godbolt.org/z/KPWKdfjT9


ksahlin:
Indeed this is what I hade googled and used.. and great to know!

PR #89, opened 2022-10-11
Add some structs to replace tuples

marcelm:
Accessing tuple fields with `std::get<index>` is error prone and not so readable, so this replaces many tuples with proper structs that have named fields.
* `MersIndexEntry`: an element of `flat_vector` (with fields hash, position, packed)
* `KmerLookupEntry`: an element of `kmer_lookup`
, `QueryMer`: an element of `mers_vector_read`
* `Hit`
I’m not so happy with the names of the structs, yet. In particular, `Hit` collides a bit with the lowercase `hit` struct in `index.hpp`, but I think this is an improvement already and finding better names can be done in a separate step.

### DIFF
--- a/src/aln.cpp
+++ b/src/aln.cpp
@@ -91,15 +91,15 @@ static inline void find_nams_rescue(std::vector<std::tuple<unsigned int, unsigne
 //    std::vector<std::pair<int, int>> repetitive_fw, repetitive_rc;
     for (auto &q : query_mers)
     {
-        auto mer_hashv = std::get<0>(q);
+        auto mer_hashv = q.hash;
         if (mers_index.find(mer_hashv) != mers_index.end()){ //  In  index
             total_hits ++;
             auto ref_hit = mers_index[mer_hashv];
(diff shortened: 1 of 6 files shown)


ksahlin:
Regarding your comment about 'Hit'. Perhaps 'Match' will do? I typically use "Match" in the papers. Then the functions `find_NAM*`also make more sense since a NAM = "non-overlapping approximate _match_" which is formed from a set of matches (merged matches).

However, I had a quick look into what the old `hit` struct does but to me it seems it aims to do keep the same info as you new `Hit`. Perhaps it becomes redundant with your new Hit-struct?

ksahlin:
Also, all the other thinks looked good, really makes it cleaner. I realise now to make much more frequent use of structs and classes, for readability (which is nothing different from my go-to language Python..)

marcelm:
Structs are great! Apart from the readability, another advantage over tuples is that you can add methods to them; then they can evolve into proper classes.

I’ll leave the `Hit` name for now and look into that later.

PR #90, opened 2022-10-11
Move logging_variables to aln.hpp and rename to AlignmentStatistics

marcelm:
Only by inspecting the code became it clear to me that the values stored in `logging_vars` are actually only for the alignment stage. Renaming it to `AlignmentStatistics` makes this clearer. Also, move the struct to `aln.hpp` because that is the right place (it doesn’t have anything to do with index creation).

### DIFF
--- a/src/aln.cpp
+++ b/src/aln.cpp
@@ -820,7 +820,7 @@ static inline bool sort_highest_sw_scores_single(const std::tuple<int, alignment
 }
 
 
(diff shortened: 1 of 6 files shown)


ksahlin:
Ok, I thought some of the logging occurs when building the index, but probably this info is just printed to stdout, which is even worse i guess.. So makes sense.

marcelm:
Yes, the statistics from the index creation step are just printed directly to stderr. I think the reason the `logging_vars` struct exists is so that it can be passed into each alignment worker thread because directly printing to stderr wouldn’t work in threads.

We should possibly add an `IndexCreationStatistics` class.

ksahlin:
Yep, that was the idea with those `logging_vars`.  Sounds good with the separate `IndexCreationStatistics`!

PR #103, opened 2022-10-21
Refactor index

marcelm:
Hopefully no logic changes, only refactoring.

- Pass `StrobemerIndex` to functions instead of two separate `kmer_lookup` and `mers_vector` objects. This cleans up some functions signatures nicely.
- Added a `IndexParameters` struct for all parameters that influence index creation.
- Move some parameters from `mapping_params` into `IndexParameters` (not all, yet)
- Make more parameters `const`

Adding the `IndexParameters` struct increases again the number of parameters for some functions, but I actually want to move the indexing parameters directly into the StrobemerIndex, and then the extra parameters disappear once more. I’m not sure whether that is actually possible, but in any case, I find it helpful to have index parameters in a different place than mapping parameters because we could then more easily store them along with the index on disk (and because it’s cleaner).

### DIFF
--- a/src/index.hpp
+++ b/src/index.hpp
@@ -89,11 +89,12 @@ public:
 };
 
 /* Settings that influence index creation */
-struct IndexParameters {
-    int k;
-    int s;
-    int l;
-    int u;
+class IndexParameters {
+public:
+    int k { 20 };
+    int s { 16 };
+    int u { 7 };
+    int l { 0 };
 
     void verify() const {
         if (k <= 7 || k > 32) {
--- a/src/index.hpp
+++ b/src/index.hpp
@@ -139,7 +139,7 @@ struct StrobemerIndex {
 
     void write(const References& references, const std::string& filename) const;
     void read(References& references, const std::string& filename);
-    void populate(const References& references, mapping_params& map_param);
+    unsigned int populate(const References& references, const mapping_params& map_param);
 private:
     ind_mers_vector generate_seeds(const References& references, const mapping_params& map_param) const;
 
--- a/src/index.hpp
+++ b/src/index.hpp
@@ -88,20 +88,36 @@ public:
     void update(int dist);
 };
 
+/* Settings that influence index creation */
+struct IndexParameters {
+    int k;
+    int s;
+    int l;
+    int u;
+
+    void verify() const {
+        if (k <= 7 || k > 32) {
+            throw BadParameter("k not in [8,32]");
+        }
+        if (s > k) {
+            throw BadParameter("s is larger than k");
+        }
+        if ((k - s) % 2 != 0) {
+            throw BadParameter("(k - s) should be an even number to create canonical syncmers. Please set s to e.g. k-2, k-4, k-6, ...");
+        }
+    }
+};
+
 struct mapping_params {
     uint64_t q;
     int n;
-    int k;
-    int s;
+    int r;
     int t_syncmer;
     int w_min;
     int w_max;
     int max_secondary;
     float dropoff_threshold;
-    int r;
     int m;
-    int l;
-    int u;
     int c;
     float f;
     int S;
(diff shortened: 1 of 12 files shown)


ksahlin:
Noob question but are we sure that `>, <, etc` always has precedence over `&&, ||` ? 

Related; I got fooled by C++ trying python comparisons `if a < b < c` so now I am always extra careful using parenthesis in everything.

For example, I used `if (a < b) && (b < c)` with parenthesis. You have removed them to be  `if a < b && b < c`. I trust you, but couldn’t find such documentation on precedence/binding-order on google so wanted to ask you here.

src/index.hpp
```diff
@@ -106,14 +106,23 @@ struct IndexParameters {
             throw BadParameter("(k - s) should be an even number to create canonical syncmers. Please set s to e.g. k-2, k-4, k-6, ...");
         }
     }
+
+    int t_syncmer() const {
```
ksahlin, line 110: Why making these as methods as they are computed once given input parameters? Doesn’t making this a method increase runtime through repeated function calls when it is really a const INT?

marcelm: > Noob question but are we sure that `>, <, etc` always has precedence over `&&, ||` ?

Yes, see https://en.cppreference.com/w/cpp/language/operator_precedence. I don’t know all precedences by heart either, but I think the one fact that comparisons have higher precedence than logical and/or is one worth to memorize because it safes so many parentheses.

> Related; I got fooled by C++ trying python comparisons `if a < b < c` so now I am always extra careful using parenthesis in everything.

That is indeed something that is quite different. It’s the same the other way: People coming to Python from other languages are suprised by this behavior.

> For example, I used `if (a < b) && (b < c)` with parenthesis. You have removed them to be `if a < b && b < c`. I trust you, but couldn’t find such documentation on precedence/binding-order on google so wanted to ask you here.

I use parentheses for cases like `a && b || c` (even when technically not necessary) because I can never remember whether `&&` or `||` has higher precedence, and even if I did, I don’t want everyone else to have to look it up.

marcelm:
> Noob question but are we sure that `>, <, etc` always has precedence over `&&, ||` ?

Yes, see https://en.cppreference.com/w/cpp/language/operator_precedence. I don’t know all precedences by heart either, but I think the one fact that comparisons have higher precedence than logical and/or is one worth to memorize because it safes so many parentheses.

> Related; I got fooled by C++ trying python comparisons `if a < b < c` so now I am always extra careful using parenthesis in everything.

That is indeed something that is quite different. It’s the same the other way: People coming to Python from other languages are suprised by this behavior.

> For example, I used `if (a < b) && (b < c)` with parenthesis. You have removed them to be `if a < b && b < c`. I trust you, but couldn’t find such documentation on precedence/binding-order on google so wanted to ask you here.

I use parentheses for cases like `a && b || c` (even when technically not necessary) because I can never remember whether `&&` or `||` has higher precedence, and even if I did, I don’t want everyone else to have to look it up.

ksahlin:
About the precedence, great thanks! 

Not sure you saw my comment on the code above about "Turn w_min/w_max/t_syncmers into IndexParameters methods" ? I did do 'complete revision' after the comment.

src/index.hpp
marcelm: I’ll update this.

marcelm:
The reason I added a `t_syncmer()` method was that I wanted this attribute to be read-only, but when trying to make this work, I realized that it’s much easier to instead use `const` attributes.

So now `IndexParameters` has only `const` attributes that are set to their values just once at construction time. One tiny problem is that const attributes need to be set in the initializer list (the `... : member1(value1), member2(value2) { ... }` thing), which is hard to make work here because we would need to replicate the logic from `adjust_index_params_depending_on_read_length` somehow. So instead, there is now a static method that returns a new `IndexParameters` instance (this is the "named constructor idiom").

It’s really nice to have an immutable object for this. This makes it clear that these settings are never changed for a given index. (To make this even clearer, I may try to make the index settings part of `StrobemerIndex` itself.)

PR #114, opened 2022-10-26
Fix estimate_read_length and add a test for it

marcelm:
The `est_read_length` function divides by the number of requested reads (500) to compute the average, but it should divide by the number of actually returned reads. This makes a difference in the test dataset because it has fewer than 500 reads.

This changes the mapping results for the test dataset since previously, different presets for the index were selected. I have updated the expected single-end SAM file, but not the paired-end one because this makes one of the "rescuable.42" reads unmappable. I don’t know why at the moment.


marcelm:
I replaced the sequence of the "rescuable.42" read with one that can be rescued again, so this is now ready.

Interestingly, using the different index parameters (read length 300), we no longer have any mapping qualities equal to zero. But figuring out why MAPQs work the way they do and/or fixing them is a different issue.

Issue #115, opened 2022-10-26
Suggestions for the on-disk index


marcelm:
Now with a little bit of experience of using the on-disk index, I would like to suggest a couple of changes in the way the on-disk index is handled (considering the discussions in, for example, #111 and #109).

* [x] If `-i` is provided, do not compute alignments (reported in #108), only create the index and store it to disk. Possibly complain if read files are provided.
* [x] Let option `-i` not accept a path, and instead always write the index to a file named as the FASTA file plus `.sti` (example: the index for `phix.fasta` is written to `phix.fasta.sti`). This ensures that the correct extension is used.
* [x] Add an option that means "a pre-computed index exists, use it". The name of the index file is reconstructed from the FASTA file name by adding `.sti`.
* [x] Do not store the sequences in the index file. With the above, the name of the FASTA file is also available and reading it is quite cheap. (If we want to be fancy, we could store a checksum to ensure the contents of the FASTA file haven’t changed.)
* [x] Store the index parameters in the index file.
* [x] When mapping with a pre-computed index, load the index parameters from the index file (and complain if any are provided on the command line).

ksahlin:
I agree with all of the above. But how to make the most minimal user interface?

One idea: How about requiring `-i` to take a parameter for the read length then (i.e., dictating all indexing params). This will prevent the indexing to take two mandatory parameters complicating the interface further.

A proposition:

```
# Creates index ref.fa.stiX, -i needs a read length X to decide indexing parameters
strobealign -i X ref.fa

# Use previously created index refs.fa.stiX to align
# complain if inferred read params X' and indexed X (from ref.fa.stiX) differ                
strobealign -P ref.fa reads1.fq reads2.fq  

# Create index and align accepts any params and inferred read length
strobealign ref.fa reads1.fq reads2.fq     
```

marcelm:
I started work on this in #117 as you may have seen. I called the option `--use-index`, but we can add `-P` as an alias.

I’m not so sure about making `-i` require the read length as a parameter because it is a little bit inconsistent to use `-r` for the read length in one case and `-i` in another. Also, thinking about the above points, I realized it is more practical to allow providing a read file when creating the index. The reads are then only used to estimate the read length. The user doesn’t have to find out the read length themselves. So you would have these variants:
```
strobealign -i ref.fasta reads.1.fastq reads.2.fastq  # Use read length from the FASTQ files
strobealign -i -r 150 -r ref.fasta  # Use explicitly provided read length
```

[...]

PR #117, opened 2022-10-27
Turn -i into a flag and add --use-index option

marcelm:
- When `-i` is specified, write the index to reference path + `.sti`
- When --use-index is specified, read the index from reference path + `.sti`
- If `-i` is used and one or two reads files are provided, estimate read length from the reads
- Do not write the reference sequences to the index (read them from the FASTA instead)
- Write a "magic number" as the first four bytes to the index file so that we can be somewhat sure that it actually is a strobemer index when reading it in again.

## To Do 

- [x] Log message "Total time creating strobemers" is printed even with `--use-index`

See #115

### DIFF
--- a/src/cmdline.cpp
+++ b/src/cmdline.cpp
@@ -22,7 +22,8 @@ std::pair<CommandLineOptions, mapping_params> parse_command_line_arguments(int a
     args::Flag x(parser, "x", "Only map reads, no base level alignment (produces PAF file)", {'x'});
     args::ValueFlag<int> N(parser, "INT", "Retain at most INT secondary alignments (is upper bounded by -M and depends on -S) [0]", {'N'});
     args::ValueFlag<std::string> L(parser, "PATH", "Print statistics of indexing to PATH", {'L'});
-    args::ValueFlag<std::string> i(parser, "PATH", "Generate an index (.sti) file", { 'i' });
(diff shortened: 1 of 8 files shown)


ksahlin:
I agree also with the To-Do of Log message "Total time creating strobemers". If this message is consistent among settings it will be convenient in mapping runtime analysis because the message can be reg-exp'ed and subtracted in both settings from the total runtime.

marcelm:
> I agree also with the To-Do of Log message "Total time creating strobemers".

Now I’m not sure I understand. To elaborate: The message 'Total time creating strobemers' is printed even when the option `--use-index` is used (and I don’t know at the moment what it is measuring), but IMO it should not be printed. Instead, a message that says 'Total time creating strobemers' should be printed. But we can also change this to read "Total time creating/reading strobemers" or something like this so that it is the same in both cases.

ksahlin:
My mistake, I mixed it up with  "Total time indexing". 

Also, it is a super easy tweak to use whatever unique timing log message for the benchmarking. So ignore my comment in full. Currently I am looking for "Total time indexing:" (all-in-one) or "Total time reading index:" (pre-generated index) when subtracting index time from map time.

marcelm:
Alright, I think the messages look ok now from what I can tell. At some point, I want to go through those logging messages and make them a bit nicer somehow, and I’ll review them more at that time.

Issue #95, opened 2022-10-13
Possibly add an IndexCreationStatistics struct


marcelm:
To be used and returned from `index_vector`. Then all the logging would be done in `main.cpp`, which is how we would want it if StrobeAlign is to become a library. (Library functions shouldn’t output anything, only return the info needed so that the main program can decide whether to output it.)

PR #120, opened 2022-10-28
Addressing issue #95 by adding IndexCreationStatistics struct

PaulPyl:
This does what is described in issue #95, however regarding making StrobeAlign a library, there is more logging code in StrobemerIndex::populate that is used for timekeeping, possibly these times could be part of IndexCreationStatistics as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 build/
+.vscode/settings.json
+.DS_Store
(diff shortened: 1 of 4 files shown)


PaulPyl:
I actually just moved the timing code as well in this commit: de1f5a7

The checks still fail but in this branch I have not touched alignment or mapq code so I suspect that this is unrelated.

PaulPyl:
Seems like the code underlying the changes I did here has changes to I am throwing this one away as well.

PR #123, opened 2022-10-28
Introduce a Timer class and use it

marcelm:
This introduces a simple `Timer` class that simplifies some of the repetitive time measurement code. I only made the necessary changes in `aln.cpp` an not in `main.cpp` and `index.cpp` because I did not want to cause merge conflicts in the other open PRs.


marcelm:
I merged this into the wrong branch. I have now pushed this to main manually.

PR #132, opened 2022-10-31
Show the Git commit hash in the version number

### LINKED ISSUE #106: Subversions for master branch
I realised that now that we want to compare possibly multiple versions of strobealign. It would be good if we have a versioning number so that it is clear which version is installed. Not sure what the best solution is.

One suggestion: since we work with pull requests (well, I should start with it too..), how about labelling the version to be pushed to master with the current stable version-id plus the github 7-characters ID on the last commit before the merge?  So:

```

$ strobealign-0.7.1
  strobealign reference [reads1] [reads2] {OPTIONS}

    StrobeAlign 0.7.1

  OPTIONS: ...
```  

```

$ strobealign-master
  strobealign reference [reads1] [reads2] {OPTIONS}

    StrobeAlign 0.7.1-2039d7da

  OPTIONS: ...
```

marcelm:
This uses [cmake_git_version_tracking](https://github.com/andrew-hardin/cmake-git-version-tracking/), which uses `git --describe` to generate the version number. This gives us something like `v0.7.1.0-340-g18c7cb0`. This means that this version is Git commit 18c7cb0, which is 340 commits after the most recent annotated tag, which was `v0.7.1.0`:
```bash
$ build/strobealign --help|head
  strobealign reference [reads1] [reads2] {OPTIONS}

    StrobeAlign v0.7.1.0-340-g18c7cb0

  OPTIONS:
```
This works only if we start using *annotated* Git tags. That is, Git tags need to be made with `git tag -a`. I took the liberty of pushing an annotated tag named `v0.7.1.0` to the repo that refers to the same commit as the unannotated tag `v0.7.1`.

Closes #106


marcelm:
I also added a `--version` command-line argument. If provided, the version number is printed and the program exits.

ksahlin:
This is great!

Issue #133, opened 2022-10-31
The separate indexing and alignment interface. 


ksahlin:
Related to  #59 but wasen't sure if I should post it there or create separate issue, went with the latter.

I was just attempting to benchmark commit ddf90e5 but ran into a snag. 

First, it took me a while to figure out that `--use-index` takes the `ref.fasta` and infers if a `ref.fasta.sti` is present and use it. It complains if `--use-index ref.fasta.sti` is provided with  the error message `strobealign: FASTA file must begin with '>' character, not 'S'`. I guess the confusion came from the documentation
```
reference     A pregenerated strobemers index file (.sti) or a reference in FASTA
                    format
```

Second, and more pressing, one advantage with the old interface of naming the index file manually is if we have indexes for different read lengths we name and use them as `-i ref.fa.100.sti` `--use-index ref.fa.100.sti` and so on (which is e.g. how my previous benchmarks were set up for the old interface). The limitation with the current is that we could have at most one index present with the current interface. 

Third, a small issue is that the current version creates an index without complaining if `-i ref.fa` is specified without `-r` or read files. This generates a default 150bp index but it is not ideal. 

The more pressing of these three items is number 2, as I would prefer not to redesign the benchmark rules again. I think it is kind of neat to have the indexing and alignment in separate rules for easy measure of peak memory etc. Should we either:

1. Name the indexes by appending the inferred read length like  `-r 100 -i ref.fa` would automatically yield an index file `ref.fa.100.sti`. A slight problem with this is when read length is inferred from reads. It would then be better to append the read length in some of the boundary conditions of the if statements controlling the ranges and check if they are equal for the provided reads and pre-generated index.
2. Or revert/revise to previous interface where `-i` requires a manually specified `.sti` file name

marcelm:
The first problem (documentation) is on my to-do list (I have a separate list with small things that I didn’t want to post as issues here). I have submitted PR #134 as temporary workaround for the second issue. This is mentioned in #109, but we should possibly split it out. The third issue (index can be created without `-r`) is tracked in #109.

marcelm:
> 1. Name the indexes by appending the inferred read length like  `-r 100 -i ref.fa` would automatically yield an index file `ref.fa.100.sti`. A slight problem with this is when read length is inferred from reads. It would then be better to append the read length in some of the boundary conditions of the if statements controlling the ranges and check if they are equal for the provided reads and pre-generated index.

[...]

PR #134, opened 2022-10-31
Require a path for -i/--use-index

marcelm:
Temporary fix for #133. With this, options `-i` and `--use-index` require the path to the index file. Is this good enough? You do need to adjust the benchmark script, though.
```
strobealign -r 200 -i ref.200.sti ref.fasta
strobealign -r 200 --use-index ref.200.sti ref.fasta reads.1.fastq reads.2.fastq
```
This isn’t as nice as I would like it, but would at least allow you to run the benchmarks. We can then discuss how to improve this later.


marcelm:
I’ll close this because we have #195 that should take care of this in a better way.

PR #136, opened 2022-11-02
Fix std::max that should be std::min

marcelm:
This was decreasible for a drop in accuracy.

Partially reverts commit cd46f8f

This does not yet restore accuracy of v0.7.1, yet. There is a second commit that caused a decrease; I will fix that as soon as I have time.


ksahlin:
Great. Then I'll hold off with running the tests again until the second one also has been detected and merged into master.

PR #138, opened 2022-11-03
Fix conversion of reference to uppercase

marcelm:
I found that commit d9872741201396f2ce144e0b0ed0ce15d7ae5ac3, merged as part of PR #55, is responsible for a decrease in accuracy. The issue is that only the last reference in the FASTA file is converted to uppercase. This problem was not caught using the phiX reference because it only contains one sequence. This PR adds more extensive tests, fixes the bug and also fixes another issue where the `names` attribute could have too many entries.

I don’t know, yet, whether this restores accuracy to the level of v0.7.1. I will check that later.

### DIFF
--- a/src/refs.cpp
+++ b/src/refs.cpp
@@ -11,7 +11,6 @@ References References::from_fasta(const std::string& filename) {
     ref_lengths lengths;
 
     std::ifstream file(filename);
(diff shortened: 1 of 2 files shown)


ksahlin:
great!

PR #145, opened 2022-11-21
Run tests on a bigger dataset

### LINKED ISSUE #128: Run regular tests on a non-toy dataset
To avoid issues like the one fixed by #127, we need more extensive tests that cover more cases. It appears some issues are only found by using a larger reference, so that is what we should test. The smallest reference I have tried (*D. melanogaster*) has 130 Mbp, which takes to long to process be run as part of CI. So for the moment, I would just add a script that needs to be run manually. 

Perhaps we can also find a dataset with a reference size between phiX and *D. melanogaster* that is faster to process and still exhibits the problem from #127.

### LINKED ISSUE #140: Accuracy regressions since v0.7.1
This is to document which accuracy changes I found since v0.7.1.

## Test data

Only paired-end reads.
* simulated Drosophila
  - Queries: First 1 million 50 bp reads from /proj/snic2022-6-31/nobackup/strobemap_eval/reads_PE/drosophila/
  - Reference: /proj/snic2022-6-31/nobackup/strobemap_eval/genomes/drosophila.fa
* real Drosophila
  - Queries: First 1 million reads of SRR6055476
  - Reference: [GCF_000001215.4](https://ftp.ncbi.nlm.nih.gov/genomes/all/GCF/000/001/215/GCF_000001215.4_Release_6_plus_ISO1_MT/GCF_000001215.4_Release_6_plus_ISO1_MT_genomic.fna.gz)
* CHM13
  - Queries: First 100'000 reads of /proj/snic2022-6-31/nobackup/strobemap_eval/reads_PE/CHM13/200_?.fq
  - Reference: /proj/snic2022-6-31/nobackup/strobemap_eval/genomes/CHM13.fa

Command used: `strobealign-${commithash} -t 1 ${reference} ${r1} ${r2} | samtools view --no-PG -o out.bam`

## Commits with accuracy or other output changes

|Commit|Accuracy in CHM13|Description|Fix/discussion|Trigger in real Drosophila|
|-|-|-|-|-|
|v0.7.1|94.3725||baseline||
|cd46f8f|94.363|Use std::{max,min,abs} instead of ternary operators|#136|SRR6055476.524|
|14d512f|94.3650|Fix a probable logic error due to a typo|||
|d987274|94.169|Do not convert reference to uppercase in seq_to_randstrobes2|#138|?|
|13bdc5d|94.169|Factor out rescue_read|#139|SRR6055476.15|
|6973009|92.875|Invert conditions to allow early break|#127|?|
|d82ce31|93.997|Bug in only this one commit: `filter_cutoff` was set incorrectly|||
|720887d|92.876|Fix incorrect refactor, part 1|||
|8d512a9|94.169|Fix incorrect refactor, part 2|||
|e758839|94.1765|Merge pull request #136 from ksahlin/fix-max||
|464a852|94.3745|Merge pull request https://github.com/ksahlin/StrobeAlign/pull/138 from ksahlin/uppercase||

marcelm:
This adds another GitHub Actions job that tests StrobeAlign on a somewhat bigger dataset. The idea is to compare the SAM/BAM output against a "baseline" commit for which we assume StrobeAlign produces correct results.

The file `tests/baseline_commit.txt` sets this baseline commit. It is currently ef86bce15896b5cca437c54ff95dfad40da2145a, which is the same as the current main. We have established that this version gives us results that are sufficienty close in accuracy to v0.7.1, so this should be a good starting point. Whenever mapping results change on purpose, we need to update the commit hash in that file. I am not super happy with this procedure and would prefer to not have this manual update, but have to think a bit how to improve upon that.

The steps of the GitHub Action are as follows:
1. A *D. melanogaster* reference and 100'000 real reads from the Sequence Read Archive are downloaded. This step takes a bit long, so I tried to enable caching for it. This means that the dataset is only re-downloaded from SRA if the `download.sh` script changes, and otherwise the cached files directly from GitHub servers are used.
2. A StrobeAlign binary is built for the baseline commit.
3. A StrobeAlign binary is built for HEAD.
4. The test data is mapped using the baseline binary.
5. The test data is mapped using the HEAD binary.
6. The BAM outputs are compared using a new `samdiff` script. This is based on the script that I wrote for the detective work in #140.

I will temporarily add a commit with a deliberate bug so you can see what the `samdiff` script produces if there are differences. Note that the script only compares reference name and position for each SAM record. It ignores changes in MAPQ, CIGAR, TLEN etc. This is on purpose because this test is about noticing when alignment accuracy decreases. Those other changes should already be caught by the existing tests.

@ksahlin Regarding the choice of dataset, I feel a bit bad that I’m not using one of the simulated datasets that you generated for this. One reason is that these files need to be put somewhere for download and I didn’t quite know where. The main reason, however, was that the real dataset was much better able to flag the bad commits. I evaluated all three datastes (your simulated dataset 1 and 2 and the real Drosophila dataset I use now), and looked at twelve different commits between v0.7.1 and main that introduced some relevant change in alignment results. Dataset 1 flagged 2/12, dataset 2 flagged 6/12 and the real Drosophila dataset flagged 11/12 commits.

The downside with using a real dataset is that we don’t have expected mapping locations. We can start using a simulated dataset when that becomes important.

Closes #128
Closes #140

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,3 +32,39 @@ jobs:
           make install
       - name: Run
         run: tests/run.sh
+
+  compare:
+    if: >-
+      github.event_name != 'pull_request' ||
+      github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - run: cat tests/baseline-commit.txt >> $GITHUB_ENV
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ env.baseline_commit }}
+          path: baseline
+      - name: Build baseline version
+        run: |
+          cd baseline
+          cmake -B build
+          make -j3 -C build
+      - name: Install Linux dependencies
+        if: runner.os == 'Linux'
+        run: sudo apt-get install samtools python3-pysam
+      - name: Install macOS dependencies
+        if: runner.os == 'macOS'
+        run: brew install samtools pysam
+      - name: Download test dataset
+        run: tests/download.sh
+      - name: Generate baseline BAM
+        run: |
+          baseline/build/strobealign tests/drosophila/ref.fasta tests/drosophila/reads.1.fastq.gz tests/drosophila/reads.2.fastq.gz | samtools view --no-PG -o baseline.bam
+      - name: Build HEAD version
+        run: |
+          cmake -B build
+          make -j3 -C build
+      - name: Generate HEAD BAM
+        run: |
+          build/strobealign tests/drosophila/ref.fasta tests/drosophila/reads.1.fastq.gz tests/drosophila/reads.2.fastq.gz | samtools view --no-PG -o head.bam
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,6 +56,11 @@ jobs:
       - name: Install macOS dependencies
         if: runner.os == 'macOS'
         run: brew install samtools pysam
+      - name: Cache test dataset
+        uses: actions/cache@v3
+        with:
+          key: test-data-${{ hashFiles('tests/download.sh') }}
+          path: tests/drosophila/
       - name: Download test dataset
         run: tests/download.sh
       - name: Generate baseline BAM
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,11 +60,12 @@ jobs:
         run: tests/download.sh
       - name: Generate baseline BAM
         run: |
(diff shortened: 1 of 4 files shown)


ksahlin:
This is super! And great that the drosophila real data flags even more commits.

Does this mean the there are more commits producing differences besides the ones detected in #140? I count 10 commits besides baseline in that issue.

marcelm:
> Does this mean the there are more commits producing differences besides the ones detected in #140? I count 10 commits besides baseline in that issue.

The two additional commits are ba8e901 and d8a1341. Not sure why left them out of the table, but the first commit added a bug that was fixed in the second commit one day later.

I noticed that there’s a 13th commit 5b462d6f594a28f1e9a23bd133e485b233c1f3e2 mentioned in [one of the comments](https://github.com/ksahlin/StrobeAlign/issues/140#issuecomment-1302126477). That one is also flagged by the real Drosophila dataset.

ksahlin:
I see!

marcelm:
Merging this because I finally want to be able to get back to refactoring :smile:

ksahlin:
sounds good :)

PR #146, opened 2022-11-21
Deliberate bug

marcelm:
This is a temporary PR related to #145. It shows how the baseline comparison test fails when a bug is introduced that changes mapping results.

Note how the normal tests on phiX do not catch this.


marcelm:
Direct link to the test output: https://github.com/ksahlin/StrobeAlign/actions/runs/3515333783/jobs/5890460827#step:13:119

ksahlin:
very cool!

Issue #148, opened 2022-11-21
Picard ValidateSamFile: INVALID_ALIGNMENT_START


marcelm:
Picard ValidateSamFile has many complaints like this:
```
ERROR::INVALID_ALIGNMENT_START:Record 23701, Read name SRR6055476.11851, Mate Alignment start (3496047) must be <= reference sequence length (44411) on reference NW_007931074.1
```

ksahlin:
Well, the error message looks scary. Are we reporting alignments outside reference sequences? Or, more likely from the error message, is this inferred by Picard from an erroneous 'mate location' field?

marcelm:
This seems to be about an unmapped read:
```
SRR6055476.11851  105  NT_033778.4     3496047 0   98=1X43=1X8=  NW_007931074.1  3496047  0  ... ...  NM:i:2  AS:i:282
SRR6055476.11851  149  NW_007931074.1  3496047 0   *             NT_033778.4     3496047  0  ... ...
```
(BTW, did you know you can run `samtools flags 149` to see that 149 means UNMAP,REVERSE,READ2?)

This indeed looks a bit weird in some ways.
- The unmapped read has a reference name associated with it. Is it on purpose, that is, does this carry any information that you would like to keep or might it just as well be set to `*`?
- The unmapped read has its position set to the same as the mapped one although they are on different contigs
- The REVERSE flag is set for the unmapped read. I guess one can do that, but my feeling is it shouldn’t be set and unmapped read should always be stored in their original orientation.

PR #151, opened 2022-11-22
Fix invalid alignment starts

marcelm:
This fixes the SAM validation issue #148

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,4 +76,4 @@ jobs:
         run: python3 tests/samdiff.py baseline.bam head.bam
       - name: Validate with Picard
         run: |
-          PicardCommandLine ValidateSamFile IGNORE=RECORD_MISSING_READ_GROUP IGNORE=MISSING_READ_GROUP IGNORE=INVALID_ALIGNMENT_START I=head.bam
+          PicardCommandLine ValidateSamFile IGNORE=RECORD_MISSING_READ_GROUP IGNORE=MISSING_READ_GROUP I=head.bam
--- a/src/refs.cpp
+++ b/src/refs.cpp
@@ -46,3 +46,9 @@ References References::from_fasta(const std::string& filename) {
 
     return References(std::move(sequences), std::move(names), std::move(lengths));
 }
+
+void References::add(std::string&& name, std::string&& sequence) {
+    names.push_back(name);
+    sequences.push_back(sequence);
+    lengths.push_back(sequence.size());
+}
--- a/src/refs.cpp
+++ b/src/refs.cpp
@@ -8,7 +8,6 @@
 References References::from_fasta(const std::string& filename) {
     std::vector<std::string> sequences;
     ref_names names;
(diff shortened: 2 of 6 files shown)


marcelm:
Ahh, nice and now the new test is triggered. I need to look into this later.

ksahlin:
Cool indeed! I looked at the output of the validation script. It reports 10 reads for which score got better (less softclipps in beginning). For these alignments, 9 have MAPQ 60 and 1 have mapq 0.

Tthe reporting script also says `both multimappers, score got better:         4`.

Are some of the multimappers given a score 60 or how do these numbers relate to each other?

marcelm:
The part about fixing #148 was ok, so I reduced the scope of this PR to only fix that. I’ll open a separate PR for the TLEN issue.



> Tthe reporting script also says `both multimappers, score got better: 4`.
> 
> Are some of the multimappers given a score 60 or how do these numbers relate to each other?

That refers to the alignment score (the `AS` tag). That is, a read was a multimapper in the "before" dataset and also in the "after" dataset, but it was mapped to a different position and has a better alignment score at that new position.

ksahlin:
I am probably still misunderstanding something, but let me see if I got this right:

1. A multimapper: A read which maps equally well to more than one location, and this is inferred by a read having MAPQ of 0.
2. `both multimappers,`: means that both the reads in a _read-pair_ were multimappers which means they both had MAPQ 0.
3. `both multimappers, score got better: 4` informs us that four read-pairs aligns better to a location other than their alignment location in strobealign v0.7.1. That is, both the reads within a pair has a new location. 
4. In point 3 the "score" is inferred by the sum of their AS score (?) 

Given the points above would it mean that:
-  strobealign v0.7.1 misclassified (at least) 4 read-pairs as aligning equally well to several locations, but, given the new higher sum of AS scores, it is possible that they have single a unique best location? 
- As for the 10 cases reported separately, these refer to improved AS score for _one of the reads_ in a read pair.  These are separate (not overlapping with) the report of `both multimappers`. 

Are these statements correct?

marcelm:
> I am probably still misunderstanding something, but let me see if I got this right:
> 
>     1. A multimapper: A read which maps equally well to more than one location, and this is inferred by a read having MAPQ of 0.

Correct.

>     2. `both multimappers,`: means that both the reads in a _read-pair_ were multimappers which means they both had MAPQ 0.

No, this refers to single reads. "both" should be read as "both before and after".

I had started to make the `samdiff` script pair-aware but removed that when it got messy and when I realized it wasn’t really necessary for my purposes at the time.

Your points 3 and 4 also assume that "both" means "both reads in a pair", so they are also not correct. Sorry, I will need to re-phrase this in the script to make the meaning of the numbers clearer. The logic computing those numbers is not very complicated at all (see https://github.com/ksahlin/StrobeAlign/blob/main/tests/samdiff.py#L53), I just need to convey this better in the output.

ksahlin:
ah I see, the script is clear. I now understand the reporting limit (10) etc.

PR #155, opened 2022-11-23
Try to make samdiff script output a bit better to interpret

marcelm:
Example output:
```
Before/after comparisons

      19269 times: read was unmapped before and after
*         2 times: read became mapped
*         1 times: read became unmapped
     180484 times: read was mapped to same locus before and after
*        89 times: read was multimapper before and after, same alignment score (AS)
*         9 times: read was multimapper before and after, better alignment score (AS)
*       146 times: read changed in another way (shown above if any)
     200000 total reads
```

The `*` in the beginning means that the row should be zero, but is not. If there are any `*`, then the script exits with exit code 1 (we want this to make the tests fail).

### DIFF
--- a/tests/samdiff.py
+++ b/tests/samdiff.py
@@ -121,23 +121,23 @@ def main():
 
     def stat(description, value, should_be_zero: bool = True):
         ok = not should_be_zero or value == 0
-        print(f"{'*' if not ok else ' '} {value:>9} times: read {description}")
+        print(f"{'*' if not ok else ' '} {value:>9} reads {description}")
 
     print("Before/after comparisons")
     print()
-    stat("was unmapped before and after", unmapped_same, False)
+    stat("were unmapped before and after", unmapped_same, False)
     stat("became mapped", became_mapped)
     stat("became unmapped", became_unmapped)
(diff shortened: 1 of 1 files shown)


ksahlin:
I had a look at the script and output again. Looks good. However, why is there no block checking if `before` result (v0.7.1) is better than `after` (current master)? 

That is,
```
                if b_score > a_score:
                    multimapper_worse += 1
                    continue
```

marcelm:
I added the `multimapper_worse` statistics. New example output:
```
Before/after comparisons

      18270 reads were unmapped before and after
*         1 reads became mapped
*      1001 reads became unmapped
     177434 reads were mapped to same locus before and after
*       363 reads were multimapper before and after, same alignment score (AS)
*        10 reads were multimapper before and after, better alignment score (AS)
*        91 reads were multimapper before and after, worse alignment score (AS)
*      2830 reads changed in another way
     200000 total reads
```

ksahlin:
Alright, nice! Although I hope those are not real numbers, then the stats look quite grim

marcelm:
Nah, this just shows a comparison between two random old BAM files I found.

PR #158, opened 2022-11-25
Add link to RECOMB-Seq video to README

marcelm:
Not sure if you want this, but I just found that video and thought it gave a really good overview in a short amount of time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,8 +62,15 @@ Changelog
 
 See [Changelog](CHANGES.md).
 
+
+Video
+-----
+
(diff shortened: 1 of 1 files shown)


marcelm:
Ah, and there’s also https://www.youtube.com/watch?v=DS4tURz1Wio ... I haven’t watched it, yet, because it’s much longer, what do you think, should it be added as well?

ksahlin:
Yes, perhaps good. Could be added as: 

"For a more detailed presentation of the underlying seeding mechanism in strobealign (strobemers) see [this talk](https://www.youtube.com/watch?v=DS4tURz1Wio).

marcelm:
Alright, added as well!

PR #159, opened 2022-11-25
Consistently use the spelling "strobealign"

marcelm:
The README is a bit inconsistent in how it spells "StrobeAlign". It often uses all-lowercase "strobealign". This makes the name consistently be "StrobeAlign" in the README.

Lowercase `strobealign` remains when the binary itself is meant.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,13 @@
 strobealign
-==============
+===========
 
-Strobealign is a fast **short-read aligner**. It achieves the speedup by using a dynamic seed size obtained from syncmer-thinned strobemers. 
+strobealign is a fast **short-read aligner**. It achieves the speedup by using a dynamic seed size obtained from syncmer-thinned strobemers.
 
-Strobealign is multithreaded, aligns single-end and paired-end reads, and outputs mapped reads either in SAM format (default) or PAF format (extension-free mapping).
+strobealign is multithreaded, aligns single-end and paired-end reads, and outputs mapped reads either in SAM format (default) or PAF format (extension-free mapping).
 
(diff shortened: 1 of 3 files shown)


marcelm:
Ok. Maybe as a data point, I can say that I am quite happy that I switched from the spelling "cutadapt" to "Cutadapt" a while ago. With the lowercase spelling, it was annoying to decide how to write the beginning of sentences because either I used a lowercase "c" and then it interrupted the reading flow because it looked weird; or I used an uppercase "C" and then it was inconsistent again. You essentially have to come up with your own rules that are different from how the English grammar usually works. When I started to treat the program name as a proper noun (which it is), this became a lot simpler, and I was also able to distinguish it from the name of the binary, which was also nice.

That said, I have also published software with all-lowercase names and it wasn’t an issue at all.

I’ll let this go and update the PR as you requested.

ksahlin:
I am usually flexible with these types of decisions, but now that the manuscript is more or less done (with 'strobelign' in the title and throughout) I think it weighs a lot. But I see your point and would be happy to change if this was not the case.

marcelm:
I thought as much – of course being consistent with the manuscript is more important.

I was just to update the README accordingly, but I realized that now you *do* need to decide how to spell strobealign at the beginning of a sentence: Strobealign or strobealign?

marcelm:
When the word strobealign occurs at the beginning of a sentence in the README, it’s also the beginning of a paragraph and then it’s not ambiguous to use the lowercase spelling, so I did that.

marcelm:
Can this be merged?

ksahlin:
Yes

PR #164, opened 2022-11-28
Slightly faster syncmer computation

marcelm:
This PR contains mostly refactoring commits, but the last one 33964c1c9de157a3fd7338712efd4ff117d031b9 is more interesting because it speeds up computation of snycmers a little bit.

I noticed that `make_string_to_hashvalues_open_syncmers_canonical()` keeps track of the s-mer hashes in `qs` and their start positions in `qs_pos`, but the positions can actually be computed when needed. This is possible because `qs_pos` only ever contains consecutive indices (5, 6, 7, etc.).

For mapping 1 million reads from the real Drosophila dataset, overall runtime decreases from ~304 to ~301 seconds (when run with `-t 1`). It appears that the compiler is already quite good at reducing the overhead from using the second `std::deque`, so the benefit isn’t huge. In any case, 1% is better than nothing an readability also improves IMO.

### DIFF
--- a/src/cmdline.cpp
+++ b/src/cmdline.cpp
@@ -30,7 +30,7 @@ std::pair<CommandLineOptions, mapping_params> parse_command_line_arguments(int a
 
     args::Group seeding(parser, "Seeding:");
     //args::ValueFlag<int> n(parser, "INT", "Number of strobes [2]", {'n'});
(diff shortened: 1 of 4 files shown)


ksahlin:
final function looks much simpler!

PR #165, opened 2022-11-30
Log index creation times in a nicer way

marcelm:
This revamps the way index creation times are kept track of and logged to make them more consistent and easier to interpret.

## Before
```
This is StrobeAlign v0.7.1.0-460-ge79fa3f
Estimated read length: 151 bp
Time reading references: 0.499928 s
Time generating seeds: 2.05037 s
Time sorting seeds: 1.86406 s
Time copying flat vector: 0.125275 s
Total time generating flat vector: 4.06984 s
Total time generating hash table index: 1.30964 s
Total time indexing: 5.38767 s
Running in paired-end mode
```

## After
```
This is StrobeAlign v0.7.1.0-470-g4d250c8
Estimated read length: 151 bp
Time reading references: 0.46 s
Indexing ...
  Time generating seeds: 1.99 s
  Time sorting seeds: 1.85 s
  Time generating flat vector: 0.12 s
  Time generating hash table index: 1.31 s
Total time indexing: 5.31 s
Running in paired-end mode
```

## Details

* Ensured that index-generation runtimes are disjoint. Before, adding up the various individual times would give a duration longer than the "Total time indexing", which makes sense when you know in detail what happens under the hood, but is confusing otherwise.
* Indent all indexing times except the total by two spaces to make it clear that they represent numbers for substeps.
* Removed the "copying flat vector time" as it was included in the other numbers (counted twice)
* "Total time generating flat vector" now reports the time "copying flat vector" reported.
* Precision changed so that two figures after the decimal point are reported. The measurements aren’t more accurate than that anyway.
* Some refactoring along the way: Use Timer everywhere, keep track of statistics within `StrobemerIndex` etc.

### DIFF
--- a/src/aln.hpp
+++ b/src/aln.hpp
@@ -56,7 +56,6 @@ struct AlignmentStatistics {
     std::chrono::duration<double> tot_find_nams_alt;
     std::chrono::duration<double> tot_sort_nams;
     std::chrono::duration<double> tot_extend;
-    std::chrono::duration<double> tot_rc;
     std::chrono::duration<double> tot_write_file;
 
     unsigned int tot_ksw_aligned = 0;
@@ -73,7 +72,6 @@ struct AlignmentStatistics {
         this->tot_find_nams_alt += other.tot_find_nams_alt;
         this->tot_sort_nams += other.tot_sort_nams;
         this->tot_extend += other.tot_extend;
-        this->tot_rc += other.tot_rc;
         this->tot_write_file += other.tot_write_file;
         this->tot_ksw_aligned += other.tot_ksw_aligned;
         this->tot_rescued += other.tot_rescued;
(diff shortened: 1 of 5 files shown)


ksahlin:
Looks good. Not sure why the first point would be a downside though. Are you referring to that it is more convenient having the statistics available as soon as generated for benchmarking/testing?

marcelm:
> Are you referring to that it is more convenient having the statistics available as soon as generated for benchmarking/testing?

No, more mundane: In interactive use, it’s a bit nicer to get some progress indication while waiting for the program. This is only relevant for somewhat larger genomes, and even then it’s not really important because indexing the human genome takes three minutes on my laptop, so it’s not that one would have no output for half an hour (wondering whether anything is broken).

Some progress report would be nice when mapping the reads, though, because that can take a much longer time depending on the input files. But that is a matter for a different PR.

ksahlin:
I see. Yes, not so important.

PR #168, opened 2022-12-02
Output mapped reads in the order that they had in the input file and decrease default chunk size

### LINKED ISSUE #121: SAM records are output out of order
It seems that the alignment records are not necessarily output in the order in which the reads appear in the input FASTQ. Judging from the output, it appears that a chunk that has been aligned by one of the worker threads is immediately sent to stdout.

This is a problem because two runs of StrobeAlign can lead to different outputs.

marcelm:
Ensuring correct order is done as discussed in #121: Read chunks are numbered consecutively, and processed chunks are buffered until it is their turn. This could theoretically affect performance a little bit because the lock that protects the output buffer is held a bit longer sometimes, but I wasn’t able to measure a significant difference (it must be less than ~1% if there is a difference).

This PR also reduces the default chunk size from 100'000 to 10'000. Together with ensuring reproducible read order, this finally allows us to run the tests on the somewhat larger dataset with multiple threads.

I also added a command-line option `--chunk-size`, which is hidden from help output. I use it in the tests to set a very low chunk size, which enables multithreading even on the tiny phiX dataset, and allows to test whether the output is indeed in the correct order.

Closes #121

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,8 @@
 
 ### Bug fixes
 
+* Issue #121: When more than one thread is used, changed behavior so that
+  alignments are output in the order that the reads had in the input file.
 * PR #114: Fix read-length estimate on input files with few reads.
 * PR #78: Fixed incorrect template length (TLEN) sign.
 * Issue #56: Unmapped reads did not have quality values in the SAM output.
--- a/src/aln.cpp
+++ b/src/aln.cpp
@@ -1664,7 +1664,7 @@ static inline void rescue_mate(
     sam_aln.ref_start =  ref_start + info.ref_offset +1; // +1 because SAM is 1-based!
     sam_aln.is_rc = a_is_rc;
     sam_aln.ref_id = n.ref_id;
-    sam_aln.is_unaligned = false;
+    sam_aln.is_unaligned = info.cigar == "*";
     sam_aln.aln_length = info.length;
     tot_ksw_aligned ++;
     tot_rescued ++;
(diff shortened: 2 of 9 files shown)


marcelm:
CI is failing because the chunk size changed. This is expected because the insert size estimates are produced on smaller chunk sizes. I will therefore change the commit that we use as baseline to the one that introduced the chunk size change.

ksahlin:
This sounds great!

marcelm:
Merging this as I want this off the table before starting the weekend :smile:.

ksahlin:
what is too long alignment though? :) I remember I had some heuristic to ignore extension alignment when the extracted reference segment was too long (like 2000bp). I was still unsure why this happened in some cases.

marcelm:
Yes, exactly that: https://github.com/ksahlin/strobealign/blob/8d3ef7e113abe9159a38343d970e864fc5afaab4/src/aln.cpp#L35-L43

With the changed chunk size, this got triggered in the Drosophila test dataset for read `@SRR6055476.50009`. The chunk needs to start a couple of reads earlier. I used reads `@SRR6055476.50004` to `@SRR6055476.50009` (paired end) to reproduce it reliably. Should I open an issue for this?

ksahlin:
Yes this is definitely an issue I have not fully pinned down. So separate issue sounds good.

Potential cause: I remember that the 'long alignment' was triggered more often in the beginning of batches where insert size estimation had not been stabilised. Is insert size re-estimated for every batch size? 

Also, will we do a lot of unnecessary computations ber batch if batch size is too small? (although you verified that time for alignment does not change that much)

ksahlin:
I see, makes sense. Deterministic results is a strong priority. Also, it is good if performance is not hurt much by re-estimation. 

However, now that batch size is 10x smaller and issues arise when insert size has not stabilised, it could be that more of these bugs appear. I think solving this bug is related to having a heuristic that only uses the isize estimation if it has 'converged' (whatever that implies). Something like that we will not use it unless it has at least 50 observations or similar.

PR #169, opened 2022-12-02
Ensure that reads with too long alignment get reported as unmapped

marcelm:
Without this, the read is marked as mapped, but has "*" as a CIGAR. We then get "[W::sam_parse1] mapped query must have a CIGAR; treated as unmapped" from `samtools view` when converting to BAM.


marcelm:
Hm, I need this for #168. I think it’s easier if I just add this commit to #168.

PR #176, opened 2022-12-13
Enable assertions during tests

marcelm:
This is achieved by manually setting the compiler options for the "RelWithDebInfo" build type (and omitting `-DNDEBUG`).

Also, we need to ensure that we use the bash option `-o pipefail`. Without it, the exit code from `strobealign ... | samtools view ...` is 0 as long as samtools exits successfully even when strobealign crashed.

This also disables some assertions that I added in 97939f3e. I never noticed that they were triggered because I did not have assertions enabled. The assertions were a replacement for some previous debugging code that checks conditions which no longer appear to be true.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
       - uses: actions/checkout@v3
       - name: Build
         run: |
-          cmake -B build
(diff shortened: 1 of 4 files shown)


marcelm:
I made a mistake when updating the logic in the `samdiff.py` script, but that’s fixed now.

PR #177, opened 2022-12-13
Refactor StrobemerIndex

marcelm:
I’m working on reducing memory usage and need to make some refactorings along the way in order to encapsulate the index and some of the other data structures better, which will then make it easer to change the way the index is structured.

For example, I made `mers_index` private to ensure that it is accessed only through methods provided by the `StrobemerIndex` class. `KmerLookupEntry::count` also becomes a method. These "getter" methods are inlined by the compiler, so there’s no cost for this.

Lastly, commit bb4613176a46114832b785e4f935bfc1fc9b0a9f changes results: When sorting multiple `Hit`s, the current behavior is to sort them by count first and then by offset. I want to change this in a later commit so that a Hit doesn’t know where in the flat vector a strobemer is stored (that is, there’s no offset anymore), so I’ve removed the offset from the comparison. Since this changes some alignments, I will push another commit to change the baseline commit. (I have not pushed it yet so that the test failure is visible.)

### DIFF
--- a/src/aln.cpp
+++ b/src/aln.cpp
@@ -160,7 +160,7 @@ static inline void find_nams_rescue(
 
     for (auto &it : hits_per_ref) {
         auto ref_id = it.first;
-        std::vector<hit> hits = it.second;
+        std::vector<hit>& hits = it.second;
         std::sort(hits.begin(), hits.end(), [](const hit& a, const hit& b) -> bool {
                 // first sort on query starts, then on reference starts
                 return (a.query_s < b.query_s) || ( (a.query_s == b.query_s) && (a.ref_s < b.ref_s) );
@@ -286,7 +286,7 @@ static inline std::pair<float,int> find_nams(
 
     for (auto &it : hits_per_ref) {
         auto ref_id = it.first;
-        std::vector<hit> hits = it.second;
+        const std::vector<hit>& hits = it.second;
         open_nams = std::vector<nam> (); // Initialize vector
         unsigned int prev_q_start = 0;
         for (auto &h : hits){
--- a/src/aln.cpp
+++ b/src/aln.cpp
@@ -132,8 +132,7 @@ static inline void find_nams_rescue(
     for (auto &q : query_mers) {
         auto ref_hit = index.find(q.hash);
         if (ref_hit != index.end()) {
-            auto query_e = q.position + q.strobe_offset + k;
-            Hit s{ref_hit->second.count, ref_hit->second.offset, q.position, query_e, q.is_reverse};
+            Hit s{ref_hit->second.count, ref_hit->second.offset, q.start, q.end, q.is_reverse};
             if (q.is_reverse){
                 hits_rc.push_back(s);
             } else {
@@ -275,8 +274,7 @@ static inline std::pair<float,int> find_nams(
                 continue;
             }
             nr_good_hits++;
-            int query_e = q.position + q.strobe_offset + k; // h.query_s + read_length/2;
-            add_to_hits_per_ref(hits_per_ref, q.position, query_e, q.is_reverse, index, k, ref_hit->second.offset, count, 100000);
+            add_to_hits_per_ref(hits_per_ref, q.start, q.end, q.is_reverse, index, k, ref_hit->second.offset, count, 100000);
         }
     }
 
--- a/src/aln.cpp
+++ b/src/aln.cpp
@@ -18,8 +18,8 @@ struct Hit {
     bool is_rc;
 
     bool operator< (const Hit& rhs) const {
-        return std::tie(count, offset, query_s, query_e, is_rc)
-            < std::tie(rhs.count, rhs.offset, rhs.query_s, rhs.query_e, rhs.is_rc);
+        return std::tie(count, query_s, query_e, is_rc)
+            < std::tie(rhs.count, rhs.query_s, rhs.query_e, rhs.is_rc);
     }
 };
 
--- a/src/aln.cpp
+++ b/src/aln.cpp
@@ -91,6 +91,32 @@ inline aln_info ssw_align(const std::string &ref, const std::string &query, int 
     return aln;
 }
 
+void add_to_hits_per_ref(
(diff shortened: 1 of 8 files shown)


src/aln.cpp
```diff
@@ -91,6 +91,32 @@ inline aln_info ssw_align(const std::string &ref, const std::string &query, int
     return aln;
 }
 
+void add_to_hits_per_ref(
```
ksahlin, line 94: will this function be inline too? Or does it not matter?

ksahlin:
It seems you finally could get rid of the small `hit` struct in commit [855ec8f](https://github.com/ksahlin/strobealign/pull/177/commits/855ec8fd36c1481ab0978b5fdeb41232fe517f96). Nice!

marcelm:
>     1. The `Hit` that you modified (capital H) is only used in `find_nams_rescue`. The other hit struct is not modified. This means that your change only affects the rescue part of the code.

If you are referring only to the commit that changes the results, then yes, that only affects the rescue part.

>     2. Sorting on offset: I don't remember if there was any benefit of sorting them also on offset. I looked at the commit but I also do not recall implementing the revised code, probably because it has been revised earlier.

The previous revision is in 8aa076c397e0843d1e7d06bd43642135645789f5. Before that, the `Hit` was just a `std::tuple<...>`. Tuples can be compared, so sorting them works out of the box. When I converted the tuple to `Hit`, I had to add `operator<` so that hits can also be sorted. I implemented it so that it would give the same result as comparing a tuple would. So I think the fact that the offset was used for sorting at all was more of an accident.

>  I know that the main idea is to sort them on count to take the least abundant hits first. So I cant imagine it having any direct affect on the algorithm. How big differences do you see?

The [samdiff output](https://github.com/ksahlin/strobealign/actions/runs/3690294137/jobs/6247144834#step:13:110)  shows that there are very minor differences only. The ten individual alignments that change seem to be just as good as before. I checked also accuracy for a couple of your datasets and it didn’t change at all.

> Also, all changes you see on drosophila will be exaggerated on larger more repetitive genomes as there are more highly abundant seeds and therefore more rescue missions.

The accuracy on your simulated, highly repetitive "dataset 1" and "dataset 2" does not change.

> It seems you finally could get rid of the small `hit` struct in commit [855ec8f](https://github.com/ksahlin/strobealign/pull/177/commits/855ec8fd36c1481ab0978b5fdeb41232fe517f96). Nice!

It’s still used in `hits_per_ref`, so a little bit more work is needed. But it’s on my radar :-).

src/aln.cpp
marcelm: To find out whether it’s inlined, I actually checked the assembly output (which is *long*) ... It appears it is not inlined. I have not noticed a difference in runtime.

I would let the compiler make the decision. At optimization level `-O3`, GCC treats all functions as if they had been marked with `inline` (`-finline-functions`) anyway. Inlining can have a negative impact because the code gets larger and then things may not fit in the cache anymore, so trusting the compiler’s heuristics is probably a good idea.

Just for the fun of it, I enabled `-Winline` as a test, which prints a warning when a function marked as inline cannot be inlined. This is the case for many functions, including `find_nams_rescue`.

(That said, I realize I should have marked this function `static`. I think there are more functions that should be static and will do that in a separate PR.)

src/aln.cpp
```diff
@@ -132,8 +132,7 @@ static inline void find_nams_rescue(
     for (auto &q : query_mers) {
         auto ref_hit = index.find(q.hash);
         if (ref_hit != index.end()) {
-            auto query_e = q.position + q.strobe_offset + k;
-            Hit s{ref_hit->second.count, ref_hit->second.offset, q.position, query_e, q.is_reverse};
+            Hit s{ref_hit->second.count, ref_hit->second.offset, q.start, q.end, q.is_reverse};
```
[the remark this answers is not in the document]
marcelm: Yes, this works the same way as before. I’ll add a comment below in a minute to explain.

src/randstrobes.cpp
```diff
@@ -306,9 +305,8 @@ mers_vector_read randstrobes_query(
     RandstrobeIterator randstrobe_rc_iter { string_hashes, pos_to_seq_coordinate, w_min, w_max, q, max_dist };
     while (randstrobe_rc_iter.has_next()) {
         auto randstrobe = randstrobe_rc_iter.next();
-        unsigned int offset_strobe = randstrobe.strobe2_pos - randstrobe.strobe1_pos;
-        QueryMer s {randstrobe.hash, randstrobe.strobe1_pos, offset_strobe, true};
-        randstrobes2.push_back(s);
+        QueryMer query_mer{randstrobe.hash, randstrobe.strobe1_pos, randstrobe.strobe2_pos + k, true};
```
marcelm, line 308: In commit f7c7a3ff6c686453b1c2eb1f7a020501350527de, I changed the `QueryMer` so that it does no longer store strobe offsets, but directly the end of the second strobe. Here you can see that previously the QueryMer was initalized with `offset_strobe`, but now it’s initialized with `randstrobe.strobe1_pos, randstrobe.strobe2_pos + k` (and the name of the field has changed from `strobe_offset` to `end`)

src/aln.cpp
ksahlin: > At optimization level -O3, GCC treats all functions as if they had been marked with inline (-finline-functions) anyway.

oh didn't know this. thanks!

src/randstrobes.cpp
ksahlin: Okay! My second concern was that we only use 8 bits for storing the offset in _reference seeds_, which leads to a distance constraints of 256nt between strobe start positions. If the very end was stored, it would lead to constraint of 256-k. But I now see that your change only affected _query seeds_ and space (using only 8 bits is not a problem for _query seeds_ as we don't to this bit fiddling with them.

ksahlin:
So I have checked this PR and agree with, and understand, everything now.

marcelm:
Great, then I’ll merge this.

Issue #173, opened 2022-12-07
Memory reduction


marcelm:
I’ve been trying to understand the index data structures. I’ll summarize here for reference, but will also write down a suggestion for how to perhaps reduce memory usage.

1. A vector of randstrobes is created ("flat vector"). The elements are structs ("MersIndexEntry") with the following members:
     * hash (8 bytes)
     * position (4 bytes)
     * reference index
     * second strobe offset (packed together with the reference index into a 4-byte field)

      The vector is sorted by hash.
2. The randstrobes vector is split up into two vectors:
   * One vector (`h_vector`) contains only the hashes
   * One vector (`mers_index`) contains the other fields (position and packed reference index/second strobe offset)

   Both vectors need the same amount of memory (8 bytes per element).
3. The flat vector is deallocated.
4. A hash table (`kmer_lookup`) is created from the `h_vector` that maps a hash value to a (mers_index position, count) pair. The key needs 8 bytes and the value needs 4+4=8 bytes as well. The hash table has low overhead, so memory usage per randstrobe is only a little bit higher than 16 bytes.

   To determine which randstrobes exist for a given *k*-mer, it is looked up in `kmer_lookup`, resulting in an index and a count, and then its info is found at `mers_index[index:index+count]`.
7. The `h_vector` is deallocated.

An issue with this procedure is that the hash values exist twice at the end of step 4, increasing peak memory usage. One way to solve this would be to write the hash values to disk when splitting up the flat vector, and then read them back in when creating the hash table. This may not even be slower if enough RAM is available because the file would be in the filesystem cache.

The duplicate `h_vector` does not exist when using a pre-generated index. It contains only the `kmer_lookup` and `mers_index` tables.

Here is another suggestion that should have greater impact. I noticed that the `count` field takes up 4 bytes but contains the value 1 most of the time (most strobemers are unique), which seems a bit of a waste. Reducing the count to a single byte was my first idea, but that is difficult to achieve because the struct size would be padded to 8 bytes anyway (to ensure aligned memory access).

Instead, I think we can get rid of `count` entirely if we use one bit in the `mers_index` entry to indicate whether there is another entry after it that has the same hash. Instead of iterating over `mers_index[index:index+count]`, one would start at `mers_index[index]` and iterate until that bit was unset.

[...]

Issue #27, opened 2022-06-01
The index: 32bit hashes instead of 64bit


ksahlin:
Note to developer:

Using 32-bit hashes may reduce both vector and hash table size. For human, it is expected to reduce peak memory from 32Gb (64bit) down to around 20-24Gb

At the final hash value computation h1/2+h2/2 when the seed has been decided, simply hash this 64 bit down to 32 bit with wyhash or XXH32.

This will result in more collisions on large genomes though as there are only 4.2billion unique slots. For human, we store about 450M unique values, but for larger genomes this may be a problem (can this be solved with 32/64 bit template?). 

I see quite a reduction in both space and alignment time if we change to 32bit hash value (on a small dataset!).

marcelm:
Turning the hash type into a template parameter would indeed be the way to go.

One question: What happens at the moment when there’s a hash collision? At which time is that resolved?

ksahlin:
As for has collisions in the hash table they are resolved by the robin_hood hash table. As for true randstrobe hash collisions where the underlying strobemers are different, they will be noticed at the NAM stage. Specifically, when parsing out the leftmost and rightmost position of a strobe in the NAM. We do a sanity check that these strobes match the reference exactly. If they don', it is either of the following:


1. _false symmetrical matches_. The paper discusses this as a feature. It is basically that (s1, s2) will get the same hash value as (s2,s1). This is neccessary when masking seeds, and also turns out to be help a tad with sensitivity. See section [Modifications to strobemers in paper](https://genomebiology.biomedcentral.com/articles/10.1186/s13059-022-02831-7). 
2. A 'true' hash collision from completely different strobes, i.e. (s1+s2)=(s3+s4). 
3. If not 1 or 2, it is some other bug..

In all the above cases, the procedure is to check for case 1 by reversing the NAM and check if it fits the reference, if it does, it was because of _false symmetrical matches_. This is taken care of in function `reverse_nam_if_needed`. If this is not the case, give up.

marcelm:
Thanks, right, there was something about slightly better sensitivity, I remember now. I did mean the randstrobe collisions.

[...]

PR #179, opened 2022-12-16
Reduced memory usage with a pre-generated index

marcelm:
Here’s a draft proposal for reducing memory requirements.

As discussed in #173, each entry in the index hash table (`mers_index`) maps a randstrobe to an (offset, count) pair that point to a slice of the flat vector, where the actual information about the randstrobe is stored. For CHM13, ~500 million unique randstrobes are generated, so the hash table has 500 million entries. However, 97.4% of the randstrobes occur only once, so to store one of those, we need an (offset, count) pair (8 bytes) and the `ReferenceMer` entry in the flat vector itself (another 8 bytes). The idea implemented here is to store those randstrobes directly in the hash table:
- The most significant bit of the "count" value is used as a marker
- If it is set, the (offset, count) pair is re-interpreted as a `ReferenceMer`, where offset is interpreted as position and count is interpreted as the "packed" reference index and strobemer offset.
- If the bit is not set, (offset, count) are interpreted as before.

On constructing the index, we only add randstrobes with count greater than one to the flat vector, which gives us the memory savings.

The reason why I’ve marked this as "draft" is that peak memory on index construction is a bit larger because I had to undo the optimization of throwing away the hash values and filling the mers_index afterwards because I now need both to be available. One option would be to fall back to the previous way of doing things if `--use-index` is not provided.

I also have been working on a way to reduce the memory usage even during index creation time, but that is in a different branch

Here are some numbers on CHM13 using the 10 M 200 bp paired-end reads from your evaluation dataset.

| | main | this branch |
|-|-|-|
| Size of `.sti` file | 11.7 GiB | 8.2 GiB |
| Peak mem when creating index |28.6 GiB | 29.6 GiB |
| Time to create index | ~5:00 | ~5:00 |
| Peak mem with `-t 8 --use-index` |24.4 GiB | 20.7 GiB |
| (User) runtime mapping 10 M paired-end reads | ~2500s | ~2500s |

~~I think there is something fishy going on with the huge peak memory usage while mapping reads. I only discovered this just now when I wanted to see whether runtime changes when mapping a large number of reads. That is a separate issue however.~~ Fixed by #181. (*Edit:* Changed table above to no longer show peak memory for mapping 10 million reads. It’s the same as when mapping any number of reads. *main* includes the fix in #181.)

Note also how the index itself only contains 8.2 GiB of data. When it’s loaded, that becomes 17 GiB because of the maximum load factor of the hashtable and because its size can only be a power of two. If we could reduce this to the absolute minimum somehow, strobealign would only need ~12 GiB of RAM (3.2 GiB for the reference plus 8.2 GiB for the index plus a little extra).

### DIFF
--- a/src/aln.cpp
+++ b/src/aln.cpp
@@ -101,15 +101,29 @@ void add_to_hits_per_ref(
     KmerLookupEntry lookup_entry,
     int min_diff
 ) {
-    for (size_t j = lookup_entry.offset(); j < lookup_entry.offset() + lookup_entry.count(); ++j) {
-        auto r = index.flat_vector[j];
+    // Determine whether the hash table’s value directly represents a
+    // ReferenceMer (this is the case if count==1) or an offset/count
+    // pair that refers to entries in the flat_vector.
+    if (lookup_entry.is_reference_mer()) {
+        auto r = lookup_entry.as_reference_mer();
         int ref_s = r.position;
         int ref_e = r.position + r.strobe2_offset() + k;
-
         int diff = std::abs((query_e - query_s) - (ref_e - ref_s));
         if (diff <= min_diff) {
             hits_per_ref[r.reference_index()].push_back(hit{query_s, query_e, ref_s, ref_e, is_rc});
             min_diff = diff;
+        }
+    } else {
+        for (size_t j = lookup_entry.offset(); j < lookup_entry.offset() + lookup_entry.count(); ++j) {
+            auto r = index.flat_vector[j];
+            int ref_s = r.position;
+            int ref_e = r.position + r.strobe2_offset() + k;
+
+            int diff = std::abs((query_e - query_s) - (ref_e - ref_s));
+            if (diff <= min_diff) {
+                hits_per_ref[r.reference_index()].push_back(hit{query_s, query_e, ref_s, ref_e, is_rc});
+                min_diff = diff;
+            }
         }
     }
 }
--- a/src/aln.cpp
+++ b/src/aln.cpp
@@ -12,7 +12,7 @@ using namespace klibpp;
 
 struct Hit {
     unsigned int count;
-    unsigned int offset;
+    KmerLookupEntry lookup_entry;
     unsigned int query_s;
     unsigned int query_e;
     bool is_rc;
@@ -98,11 +98,10 @@ void add_to_hits_per_ref(
     bool is_rc,
     const StrobemerIndex& index,
     int k,
-    int offset,
-    int count,
+    KmerLookupEntry lookup_entry,
     int min_diff
 ) {
-    for (int j = offset; j < offset + count; ++j) {
+    for (size_t j = lookup_entry.offset(); j < lookup_entry.offset() + lookup_entry.count(); ++j) {
         auto r = index.flat_vector[j];
         int ref_s = r.position;
         int ref_e = r.position + r.strobe2_offset() + k;
@@ -132,7 +131,7 @@ static inline void find_nams_rescue(
     for (auto &q : query_mers) {
         auto ref_hit = index.find(q.hash);
         if (ref_hit != index.end()) {
-            Hit s{ref_hit->second.count(), ref_hit->second.offset(), q.start, q.end, q.is_reverse};
+            Hit s{ref_hit->second.count(), ref_hit->second, q.start, q.end, q.is_reverse};
             if (q.is_reverse){
                 hits_rc.push_back(s);
             } else {
@@ -150,7 +149,7 @@ static inline void find_nams_rescue(
             if ((count > filter_cutoff && cnt >= 5) || count > 1000) {
                 break;
             }
-            add_to_hits_per_ref(hits_per_ref, q.query_s, q.query_e, q.is_rc, index, k, q.offset, count, 1000);
+            add_to_hits_per_ref(hits_per_ref, q.query_s, q.query_e, q.is_rc, index, k, q.lookup_entry, 1000);
             cnt++;
         }
     }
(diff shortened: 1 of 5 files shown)


ksahlin:
Also, FWIW, ~12 GiB looks suspiciously close to minimap2's peak memory usage. Perhaps Heng has though exactly about this in his hash table implementation.

marcelm:
>     1. Wouldn't this also give us a tad speedup if reads mostly have unique seeds? (since we do not have to access the flat_vector as much?

If there is a speedup, it’s not much. The runtime was a bit lower (maybe 1-2%), but I didn’t run it often enough so that I could say anything with statistical certainty.

>     2. Wouldn't the following reduce peak memory during indexing?
>        A. construct and sort `flat_vector`
>        B. save `flat_vector` to disk
>        C. Read `flat_vector`as a stream and construct the hash table (using your new approach of storing randstorbes directly in hash table). Keeping offset incrementing only for non-unique entries. (here things over hard cutoff of 1000 can also be ignored)
>        D. Read in the `flat_vector` again (only randstrobes occurring more than once).
> 
> It would only increase indexing runtime slightly I think. While keeping peak memory as low as for alignment.

That’s a good suggestion!

Maybe I should mention the other thing I have actually already implemented (in branch `index`). It doesn’t quite work (yet) as I want it, but I want to mention the idea: The reason the flat vector is created in full in memory at all is that this is needed for sorting it. But we no longer need the hashes with count==1 to be sorted. (And technically, we don’t even need the hashes with count > 1 to be sorted, the ones with the same hash value just need to be adjacent to each other so that they can be looked up together.)

So we can just iterate over all randstrobes and directly insert them into the hash table and the flat vector. More precisely:
- Check if the randstrobe is already in the table. If not, add it as a "direct" entry.
- If it is, convert the entry to an "indirect" one if count=1 (extract the existing entry, add it to the flat vector), then add the new randstrobe.
- At the end, sort the flat vector, and go over its entries to fix up the "offset" fields in the hash table.
(This requires that the flat vector still contains the hashes.)
The issue I have at the moment is that I need to iterate over the all randstrobes once in the beginning to count the unique hashes so that I can reserve the correct size. If I don’t do that, the hash table gets resized when adding entries, which blows up memory usage and runtime. 

>     3. I am also interested in what you wrote about the hash table. Potentially a big gain to be made! Did you try if the other implementation in robin_hood is better? It should be more memory sparse, but not near the minimum I predict.

I tried it just now. Peak memory usage with `unordered_node_map` is 24 GB.

ksahlin:
> That’s a good suggestion! Maybe I should mention the other thing I have actually already implemented (in branch index). 

When I suggest something, you typically have something better going on ^^. Do you have any estimate on how much the peak memory for this approach would be? Would be great have a peak memory that is not much more than during alignment. Maybe wishful thinking. 

Also, to challenge your approach with 'extreme scenarios', the approach would not have as much gain over our previous approach for very repetitive genomes withour many unique seeds. However, CHM13 is a good standard. Just pointing it out for extreme cases such as repetitive plant genomes. Good that we have those datasets to benchmark on.

> The issue I have at the moment is that I need to iterate over the all randstrobes once in the beginning to count the unique hashes so that I can reserve the correct size. If I don’t do that, the hash table gets resized when adding entries, which blows up memory usage and runtime.

Indexing time is not crucial imo. Better to have a stable peak memory. Also, since hash table look ups are not a major part of the runtime, maybe it is worth looking at other header-only memory sparse implementations, for example [sparsepp](https://github.com/greg7mdp/sparsepp), or find any other alternative from [this benchmark](https://martin.ankerl.com/2022/08/27/hashmap-bench-01/#memory-usage)? [This](https://github.com/lh3/minimap2/blob/master/khash.h) seems to be minimap2's hash table implementation. Something about quadratic probing. Haven't looked up what it means yet. 

Another direction is to use 32bit hashes that I played around with a bit before and suggested in https://github.com/ksahlin/strobealign/issues/27. My main concern with the 32bit int hash table are collisions for very large genomes as we are capped by 4.2billion entries.

Just brainstorming a bit. Of course, the trade-off with how much slower it gets should IMO always be consulted. 


> I tried it just now. Peak memory usage with unordered_node_map is 24 GB.

Okay, that wasn't a viable direction.

ksahlin:
I was curious so I checked the results for `rye` (7.3G) and `maize` (2.4G) genomes with seeds from 2*150nt reads.
For rye, 83% of the seeds occur once out of 668M total distinct hash values. Maize has about 86% of the seeds occurring only once from a total of 208M distinct hash values. So still big gain to be made with your approach.

marcelm:
Nice results for rye and maize! 

> Do you have any estimate on how much the peak memory for this approach would be? Would be great have a peak memory that is not much more than during alignment. Maybe wishful thinking.

It would indeed be nearly the same as during alignment. I suggest that I open a separate PR for this. I think it is still valuable to first get this PR to a stage where it can be merged because that other method is a bit slower, so maybe it’s worth to have both and then be able to switch between them (high memory/fast vs low memory/slow).

> Indexing time is not crucial imo. Better to have a stable peak memory. Also, since hash table look ups are not a major part of the runtime, maybe it is worth looking at other header-only memory sparse implementations, for example [sparsepp](https://github.com/greg7mdp/sparsepp), or find any other alternative from [this benchmark](https://martin.ankerl.com/2022/08/27/hashmap-bench-01/#memory-usage)?

I tested gtl::flat_hash_map a couple of days ago. It used more memory and was a lot slower when creating the index. I’ll need to test sparsepp, but it claims to also have one byte overhead per entry, so this would be the same as what we have now. I haven’t systematically tried the other hash table implementations.

> [This](https://github.com/lh3/minimap2/blob/master/khash.h) seems to be minimap2's hash table implementation. Something about quadratic probing. Haven't looked up what it means yet.

To try this, I think we would need to find a C++ wrapper for it (haven’t searched, this probably exists somewhere).

> Another direction is to use 32bit hashes that I played around with a bit before and suggested in #27. My main concern with the 32bit int hash table are collisions for very large genomes as we are capped by 4.2billion entries.

I’ll comment over at #27.

marcelm:
... and this can be closed because it has been superseded by #191.

PR #183, opened 2022-12-19
Warn when binary was compiled without optimizations

marcelm:
When switching between branches, it often happens that CMake changes the CMAKE_BUILD_TYPE to Debug, which makes the resulting binary very slow. I have not been able to figure out, yet, why this is the case. I guess that somehow the CMakeCache.txt file is invalidated by changing the branches, and that this leads CMake to ignore the configured build type.

I would like to avoid that anyone benchmarks a strobealign binary compiled without optimizations, so until we’ve solved this, I’d like to print a fat warning when the build type is Debug.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -49,6 +49,12 @@ std::string sam_header(const References& references, const std::string& read_gro
     return out.str();
 }
 
+void warn_if_no_optimizations() {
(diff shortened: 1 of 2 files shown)


marcelm:
Ok to merge this?

ksahlin:
yes! sorry forgot to thumbs up.

PR #187, opened 2022-12-22
Fix crash in rescue_mate for short reads

marcelm:
This factors out a `has_shared_substring` function, adds a (failing) test for it and then fixes the issue.

The problem is that `read_seq.size() - k` becomes negative when the read is shorter than `k`, and
then the comparison is always true (because we compare to a `size_t`).


ksahlin:
excellent!

PR #188, opened 2023-01-03
Switch to C++ 17

marcelm:
I don’t know whether C++ 14 was chosen for a particular reason. If not, can we switch to C++ 17?

It has a couple of nice additions that I have refrained from using because we are still at C++ 14:

- Structured binding: https://en.cppreference.com/w/cpp/language/structured_binding
- string_view: https://en.cppreference.com/w/cpp/string/basic_string_view
- std::filesystem: https://en.cppreference.com/w/cpp/filesystem
- std::clamp https://en.cppreference.com/w/cpp/algorithm/clamp

Summary of C++ 17 changes: https://en.cppreference.com/w/cpp/17 That page also lists compiler support.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,7 @@ find_package(ZLIB)
 find_package(Threads)
 find_package(OpenMP)
 
-set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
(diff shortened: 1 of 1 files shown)


ksahlin:
I am okay with switching to C++17. The only reason keeping to C++14 was, from what I remember, that I had an easier time getting strobealign to compile on my mac (given the versions I had of gcc and clang at the time on my mac). With the new Cmake installation it is not an issue anymore.

marcelm:
Great, thanks! I think that will make a couple of things a bit nicer.

PR #189, opened 2023-01-03
Use pattern-defeating quicksort (pdqsort) to sort randstrobe hashes

marcelm:
This switches from `std::sort` to [pdqsort](https://github.com/orlp/pdqsort), which reduces the time needed for sorting the `ind_flat_vector` by roughly half.

I’m not sure whether this should be merged, but I wanted to at least document this experiment. Note that if I can finish my memory-saving ideas, then sorting won’t be as time-intensive anymore anyway, so the savings would be much lower.

To get a speedup at all, this requires that the `ind_flat_vector` entries (which are `MersIndexEntry` structs) are only compared by hash, not as it is currently done by hash, then reference position, then the "packed" attribute. Is there an assumption in the mapping/alignment code that requires that entries are not only sorted by hash, but also by position and the "packed" attribute?

Just simplifying the comparison in this way already speeds up sorting by ~10% (this is done in commit 26b1992c0899b367099684a326cf927d8e0f52f2).

This changes mapping results and I still need to check whether also accuracy changes.

The second commit switches from std::sort to pdqsort. Pdqsort is a lot faster than std::sort for many types of data, see
benchmarks at <https://github.com/orlp/pdqsort#benchmark>.

In our case, pdqsort is about twice as fast as std::sort when sorting randstrobe hashes, but only when using the `pdqsort_branchless` variant, and when the simpler comparison function is used.

On a 820 Mbp reference (herring), sorting time changes as follows:
- ~14 s original
- ~12 s with simplified comparison
- ~7 s with simplified comparison and using `pdqsort`

### DIFF
--- a/ext/pdqsort/license.txt
+++ b/ext/pdqsort/license.txt
@@ -1,0 +1,16 @@
+Copyright (c) 2021 Orson Peters <orsonpeters@gmail.com>
+
+This software is provided 'as-is', without any express or implied warranty. In no event will the
+authors be held liable for any damages arising from the use of this software.
+
+Permission is granted to anyone to use this software for any purpose, including commercial
+applications, and to alter it and redistribute it freely, subject to the following restrictions:
+
+1. The origin of this software must not be misrepresented; you must not claim that you wrote the
+   original software. If you use this software in a product, an acknowledgment in the product
+   documentation would be appreciated but is not required.
+
+2. Altered source versions must be plainly marked as such, and must not be misrepresented as
+   being the original software.
+
+3. This notice may not be removed or altered from any source distribution.
--- a/ext/pdqsort/pdqsort.h
+++ b/ext/pdqsort/pdqsort.h
@@ -1,0 +1,532 @@
+/*
+    pdqsort.h - Pattern-defeating quicksort.
+
+    Copyright (c) 2021 Orson Peters
+
+    This software is provided 'as-is', without any express or implied warranty. In no event will the
+    authors be held liable for any damages arising from the use of this software.
+
+    Permission is granted to anyone to use this software for any purpose, including commercial
+    applications, and to alter it and redistribute it freely, subject to the following restrictions:
+
+    1. The origin of this software must not be misrepresented; you must not claim that you wrote the
+       original software. If you use this software in a product, an acknowledgment in the product
+       documentation would be appreciated but is not required.
+
+    2. Altered source versions must be plainly marked as such, and must not be misrepresented as
+       being the original software.
+
+    3. This notice may not be removed or altered from any source distribution.
+*/
+
+
+#ifndef PDQSORT_H
+#define PDQSORT_H
+
+#include <algorithm>
+#include <cstddef>
+#include <functional>
+#include <utility>
+#include <iterator>
+
+#if __cplusplus >= 201103L
+    #include <cstdint>
+    #include <type_traits>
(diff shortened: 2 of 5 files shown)


ksahlin:
> To get a speedup at all, this requires that the ind_flat_vector entries (which are MersIndexEntry structs) are only compared by hash, not as it is currently done by hash, then reference position, then the "packed" attribute. Is there an assumption in the mapping/alignment code that requires that entries are not only sorted by hash, but also by position and the "packed" attribute?

I think we had a similar discussion in another pull request or issue. From what I remember there is no idea behind why I decided to sort on all values in the tuple (now struct). Sorting only on hash should suffice for the flat vector. However, it is good to monitor how results change. If it results in non-negligible accuracy decrease we would have to take a further look. I cannot find a reason why it would lead to worse or better results at the moment.

Also: could this sorting method increase speed when sorting smaller vectors on other places, e.g. in `aln.cpp`?

marcelm:
> I think we had a similar discussion in another pull request or issue.

Ah, you’re right, there was something. I found it now, it was in #177, but there the discussion was about the `Hit` struct.

> From what I remember there is no idea behind why I decided to sort on all values in the tuple (now struct). Sorting only on hash should suffice for the flat vector. However, it is good to monitor how results change. If it results in non-negligible accuracy decrease we would have to take a further look. I cannot find a reason why it would lead to worse or better results at the moment.

Good, then I’ll check accuracy on some datasets and then we can decide.

> Also: could this sorting method increase speed when sorting smaller vectors on other places, e.g. in `aln.cpp`?

Good point, that would be great!

ksahlin:
Sounds good!

ksahlin:
By the way, regarding sorting only by hash (not reference position). I think this may even be better than sorting also on reference positions. This is because of coverage distribution for perfect repeats. In case of perfect repeats, if we sort by reference position, it _may be_ that reads will always be mapped to the same repeat copy in that reference ("may be" because I am unsure if there is anything downstream that alleviates this - but I don't think so). 

By not sorting on reference position, reads aligning within the repeat with different start seeds could get distributed among the copies 'a bit more randomly'. There is to my knowledge currently no mechanism that distributes read alignments 'evenly' in case of several identical alignments due to perfect repeats.

marcelm:
Here are some measured accuracies for a couple of datasets, comparing main and the two commits in this PR.

Dataset | Main (43423e0) |Only sort by hash (26b1992) | Use pdqsort (6e451fe) |
-|-|-|
Drosophila (2x50bp, 100'000 reads) | 89.97 | 89.979 | 89.9575 |
CHM13 (2x100bp, 1M reads) | 93.21375 | 93.2107 | 93.21825 |
CHM13 (2x150bp, 1M reads) | 94.1531 | 94.14345 | 94.1543 |
CHM13 (2x200bp, 1M reads) | 94.4124 | 94.41345| 94.4156 |
Rye (2x150bp, 1M reads) | 90.20815 | 90.23165 | 90.2243 |

The differences are very small, but from those numbers, your intuition about sorting only by hash being better could be correct.

ksahlin:
Yes, perhaps. At minimum, it looks like a sufficient amount of experiments to claim that it at least does not seem consistently worse. Together with the speedup it definitely justifies switching to sorting where we only compare by hash. Neat!

marcelm:
Alright! I’ll push a commit that updates the baseline commit and then merge the PR if the tests pass.

PR #191, opened 2023-01-11
Memory reduction II

marcelm:
This is an extension of PR #179 (the first commits are the same). PR #179 reduced memory usage significantly when mapping using a pre-generated index, but increases memory usage at index generation time. The aim of this PR is to reduce memory usage even at index generation time.

It would have been possible to just add these commits to PR #179, but because this goes in a somewhat different direction and has different tradeoffs than what we discussed there, I’m opening a new PR.

The main downside of the method used here is that indexing time increases significantly (definitely more than 25%). I have thought about this a bit over the last weeks and am now at a point where we just need to discuss whether this is acceptable.

Memory usage is reduced to nearly the same level as when using the new index layout introduced in #179 a mapping using a pre-generated index.

Since UPPMAX servers are down for maintenance, I will need to add runtime and memory usage numbers later.

The main idea (on top of those in #179) is to [build the index directly while iterating over randstrobes](https://github.com/ksahlin/strobealign/commit/20db8150254e5adeb3e9bc5fc814aa1c8a662541). Previously, the `ind_flat_vector` was filled and then the `mers_index` and `flat_vector` were created from it. Here, we generate the  randstrobes on-the-fly and directly fill the `mers_index` and the flat vector. This is like using a generator in a Python. The previous `RandstrobeIterator` generates a vector of all randstrobes for an entire reference and uses lots of memory for large contigs. The new `RandstrobeIterator2` returns the randstrobes one by one and only keeps a small window of the relevant syncmers in memory. (In a `deque`.)

The biggest problem is that we can no longer easily estimate how much memory to pre-allocate for the hash table (`mers_index`). Previously, we could just count the distinct hash values in the sorted flat vector, which is fast and easy. Since we no longer have a full flat vector available, this is no longer possible.

If we let the hash table resize automatically without pre-allocation, memory usage grows a lot because at resize time, the old and new hash table are kept in memory so that one can be copied to the other.

For the moment, I solved this by adding a pre-processing step that generates all randstrobes without storing them and just counts how many unique hashes are created. This takes of course a lot of extra time. We also save some time because we no longer need to sort the full flat vector, only the part of it that contains randstrobes that occur more than once. Sorting time becomes nearly negligible while it was a big chunk of index generation time previously, but with all randstrobes being generated twice, runtime still increases by a bit. (I will add numbers later.)

To count unique hashes, I initially used a `robin_hood::unordered_flat_set` to keep track of all hashes, but of course also that needs a lot of memory, so I switched to using the [HyperLogLog algorithm by Flajolet et al (2007)](https://en.wikipedia.org/wiki/HyperLogLog), which uses only $O(1)$ memory.

## To Do

- [x] Add runtime and memory usage numbers
- [x] `StrobemerIndex::populate()` became quite large because I inlined `index_vector()`, `generate_and_sort_seeds()` and `randstrobes_reference()` into it. Clean it up (probably by factoring out some functions).
- [X] ~~HyperLogLog cannot deal with more than $2^{32}$ unique values, which will give problems for large references. Use HyperLogLog++ instead, which is also more accurate. (Need to find a C++ implementation somewhere.) ~~ Moved to issue #193.
- [x] `RandstrobeIterator2` gives slightly different results than `RandstrobeIterator`. Figure out why.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,9 @@
 
 ### Features
 
+* PR #191 and PR #192: RAM usage has been decreased significantly. For the
+  human genome, for example, it went from from 28 GiB to 21 GiB. (Mapping
+  runtime is unaffected.)
 * There is now a Conda package for StrobeAlign on Bioconda.
 * The index can now be pre-generated and safed to disk.
 * Invalid or missing input files no longer lead to a crash.
--- a/ext/hyperloglog/hyperloglog.hpp
+++ b/ext/hyperloglog/hyperloglog.hpp
@@ -1,0 +1,374 @@
+#if !defined(HYPERLOGLOG_HPP)
+#define HYPERLOGLOG_HPP
+
+/**
+ * @file hyperloglog.hpp
+ * @brief HyperLogLog cardinality estimator
+ * @date Created 2013/3/20
+ * @author Hideaki Ohno
+ */
+
+#include <vector>
+#include <cmath>
+#include <sstream>
+#include <stdexcept>
+#include <algorithm>
+#include "murmur3.h"
+
+#define HLL_HASH_SEED 313
+
+#if defined(__has_builtin) && (defined(__GNUC__) || defined(__clang__))
+
+#define _GET_CLZ(x, b) (uint8_t)std::min(b, ::__builtin_clz(x)) + 1
+
+#else
+
+inline uint8_t _get_leading_zero_count(uint32_t x, uint8_t b) {
+
+#if defined (_MSC_VER)
+    uint32_t leading_zero_len = 32;
+    ::_BitScanReverse(&leading_zero_len, x);
+    --leading_zero_len;
+    return std::min(b, (uint8_t)leading_zero_len);
+#else
+    uint8_t v = 1;
+    while (v <= b && !(x & 0x80000000)) {
+        v++;
+        x <<= 1;
+    }
+    return v;
+#endif
+
+}
+#define _GET_CLZ(x, b) _get_leading_zero_count(x, b)
+#endif /* defined(__GNUC__) */
+
+namespace hll {
+
+static const double pow_2_32 = 4294967296.0; ///< 2^32
+static const double neg_pow_2_32 = -4294967296.0; ///< -(2^32)
+
+/** @class HyperLogLog
+ *  @brief Implement of 'HyperLogLog' estimate cardinality algorithm
+ */
+class HyperLogLog {
+public:
+
+    /**
+     * Constructor
+     *
+     * @param[in] b bit width (register size will be 2 to the b power).
+     *            This value must be in the range[4,30].Default value is 4.
+     *
+     * @exception std::invalid_argument the argument is out of range.
+     */
+    HyperLogLog(uint8_t b = 4) throw (std::invalid_argument) :
+            b_(b), m_(1 << b), M_(m_, 0) {
+
+        if (b < 4 || 30 < b) {
+            throw std::invalid_argument("bit width must be in the range [4,30]");
+        }
+
+        double alpha;
+        switch (m_) {
+            case 16:
+                alpha = 0.673;
+                break;
+            case 32:
+                alpha = 0.697;
+                break;
+            case 64:
+                alpha = 0.709;
+                break;
+            default:
+                alpha = 0.7213 / (1.0 + 1.079 / m_);
+                break;
+        }
+        alphaMM_ = alpha * m_ * m_;
+    }
+
+    /**
+     * Adds element to the estimator
+     *
+     * @param[in] str string to add
+     * @param[in] len length of string
+     */
+    void add(const char* str, uint32_t len) {
+        uint32_t hash;
+        MurmurHash3_x86_32(str, len, HLL_HASH_SEED, (void*) &hash);
+        uint32_t index = hash >> (32 - b_);
+        uint8_t rank = _GET_CLZ((hash << b_), 32 - b_);
+        if (rank > M_[index]) {
+            M_[index] = rank;
+        }
+    }
+
+    /**
+     * Estimates cardinality value.
+     *
+     * @return Estimated cardinality value.
+     */
+    double estimate() const {
+        double estimate;
+        double sum = 0.0;
+        for (uint32_t i = 0; i < m_; i++) {
+            sum += 1.0 / (1 << M_[i]);
+        }
+        estimate = alphaMM_ / sum; // E in the original paper
+        if (estimate <= 2.5 * m_) {
+            uint32_t zeros = 0;
+            for (uint32_t i = 0; i < m_; i++) {
+                if (M_[i] == 0) {
+                    zeros++;
+                }
+            }
+            if (zeros != 0) {
+                estimate = m_ * std::log(static_cast<double>(m_)/ zeros);
+            }
+        } else if (estimate > (1.0 / 30.0) * pow_2_32) {
+            estimate = neg_pow_2_32 * log(1.0 - (estimate / pow_2_32));
+        }
+        return estimate;
+    }
+
+    /**
+     * Merges the estimate from 'other' into this object, returning the estimate of their union.
+     * The number of registers in each must be the same.
+     *
+     * @param[in] other HyperLogLog instance to be merged
+     * 
+     * @exception std::invalid_argument number of registers doesn't match.
+     */
+    void merge(const HyperLogLog& other) throw (std::invalid_argument) {
+        if (m_ != other.m_) {
+            std::stringstream ss;
+            ss << "number of registers doesn't match: " << m_ << " != " << other.m_;
+            throw std::invalid_argument(ss.str().c_str());
+        }
+        for (uint32_t r = 0; r < m_; ++r) {
+            if (M_[r] < other.M_[r]) {
+                M_[r] |= other.M_[r];
+            }
+        }
+    }
+
+    /**
+     * Clears all internal registers.
+     */
+    void clear() {
+        std::fill(M_.begin(), M_.end(), 0);
+    }
+
+    /**
+     * Returns size of register.
+     *
+     * @return Register size
+     */
+    uint32_t registerSize() const {
+        return m_;
+    }
+
+    /**
+     * Exchanges the content of the instance
+     *
+     * @param[in,out] rhs Another HyperLogLog instance
+     */
+    void swap(HyperLogLog& rhs) {
+        std::swap(b_, rhs.b_);
+        std::swap(m_, rhs.m_);
+        std::swap(alphaMM_, rhs.alphaMM_);
+        M_.swap(rhs.M_);       
+    }
+
+    /**
+     * Dump the current status to a stream
+     *
+     * @param[out] os The output stream where the data is saved
+     *
+     * @exception std::runtime_error When failed to dump.
+     */
+    void dump(std::ostream& os) const throw(std::runtime_error){
+        os.write((char*)&b_, sizeof(b_));
+        os.write((char*)&M_[0], sizeof(M_[0]) * M_.size());
+        if(os.fail()){
+            throw std::runtime_error("Failed to dump");
+        }
+    }
+
+    /**
+     * Restore the status from a stream
+     * 
+     * @param[in] is The input stream where the status is saved
+     *
+     * @exception std::runtime_error When failed to restore.
+     */
+    void restore(std::istream& is) throw(std::runtime_error){
+        uint8_t b = 0;
+        is.read((char*)&b, sizeof(b));
+        HyperLogLog tempHLL(b);
+        is.read((char*)&(tempHLL.M_[0]), sizeof(M_[0]) * tempHLL.m_);
+        if(is.fail()){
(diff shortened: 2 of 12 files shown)


ksahlin:
Nice! I don't have a big issue with indexing time being 25% slower, as it is already pretty fast. If the peak memory of the new implementation is kept at ~2/3 of peak memory for v0.7.1, I consider that a relatively big win.

ksahlin:
Another benefit of reading the randstrobes an extra pass before indexing is for what we discussed about [64 vs 32bit hashes](https://github.com/ksahlin/strobealign/issues/27#issuecomment-1357491555). With your new approach, one could determine if it is suitable to use 32bit hashes, depending on the expected collision rate. 

For example, from memory I think indexing hg38 has about 500-600M (~2^29) unique seeds (depending on read length), perhaps this is on the border line for what is tolerated given 4.3B (2^32) unique hash values (550/4300=0.13). But for smaller genomes, it would be suitable. 

I implemented 32bit hashes (a long time ago) and saw a sizeable reduction in memory and I also think I observed a small but significant speedup (say ~10%). The speedup wouldn't be that much in your new implementation though, since you already gain some speed with storing positions directly for unique hashes. (Also, I believe any speedup in this step will not compare to changing the extension alignment to WFA).

Regardless it could be interesting to try if your new approach 'sets up' nicely for dynamically deciding the size of hash values. Not a big priority if not easily implemented I would say. What do you think?

marcelm:
Here are some numbers for benchmarks run on rackham. Runtimes are extremely noisy, so take these with a grain of salt (250 could actually be 200 to 300 it appears). I will do some benchmarks on my own machine on (smaller references) overnight. That should have less noise.

But in a preliminary summary:
- Runtime goes up by a lot in some cases. It’s not as bad when we compare to the pre-pdqsort version.
- Peak memory drops a lot, but is still a little bit higher than when using a pre-generated index.

I’m still unsure whether the runtime increase isn’t too high. But dynamically deciding the size of the hash table would be a very interesting benefit of this indeed.

## Indexing time

In seconds

dataset     | before pdqsort | main (w/pdqsort) | this PR |
-|-|-|-|
rye 2x150bp | 510 | 390 | 800 |
CHM13 2x100bp | 250 | 190 | 280 |
CHM13 2x150bp | 230 | 190 | 340 |
CHM13 2x200bp | 250 | 210 | 300 |

## Peak memory when creating the index

In GiB

dataset     | main | this PR |
-|-|-|
rye 2x150bp | 47.9 | 44.7 |
CHM13 2x100bp | 28.6 | 22.6 |
CHM13 2x150bp | 28.6 | 22.4 |
CHM13 2x200bp | 28.6 | 22.3 |

ksahlin:
The indexing times with v0.7.1 for CHM13 and rye were ~270 and ~600 respectively. So the numbers you display are not a drastic increase if we compare to those values (although I will wait for more stable runtimes as you mention). Peak memory for v0.7.1 was 31.5Gb and 50Gb, respectively. So a decent cut. 

> I’m still unsure whether the runtime increase isn’t too high. But dynamically deciding the size of the hash table would be a very interesting benefit of this indeed.

Well I would say that I am still positive :) Alignment time will likely be the bottleneck in many scenarios. Loosing ~30secs in indexing time for human (compared to v0.7.1) seems worth it for (i) slightly faster alignment, (ii) reduced peak memory, and (iii) setting up for 32bit hash value indexing.

(The 190 seconds for CHM with **main (w/pdqsort)** is as fast as minimap2 indexing though (191sec reported in my paper), but I don't want to become greedy.)

marcelm:
Here are some more measurements.

## Indexing time

For the runtime measurements, I used [hyperfine](https://github.com/sharkdp/hyperfine), which is a dedicated benchmarking tool. I used the default settings, which means that it runs the command at least ten times (that’s where the standard deviation comes from).

dataset | 603124d (before pdqsort) | 449a2fe2f128bfa3bf196b39b590f9641f72371c (w/pdqsort) | this PR |
-|-|-|-|
*Oryza sativa* (rice, 381 Mbp) | 21.229 s ±  0.038 s |17.513 s ±  0.047 s|
*Clupea harengus* (herring, 818 Mbp) | 45.123 s ±  0.893 s|33.356 s ±  0.240 s|
*D. melanogaster* (146 Mbp) | 7.869 s ±  0.050 s | 6.518 s ±  0.027 s|
*E. coli* (4.7 Mbp) |225.2 ms ± 0.9 ms|193.4 ms ±   1.1 ms|

## Peak memory when creating the index

dataset | main (449a2fe2f128bfa3bf196b39b590f9641f72371c | this PR |
-|-|-|-|
*Oryza sativa* (rice, 381 Mbp) | 3.6 GiB | 2.8 GiB |
*Clupea harengus* (herring, 818 Mbp) | 7.0 GiB | 5.6 GiB |
*D. melanogaster* (146 Mbp) | 1.1 GiB | 0.8 GiB |
*E. coli* (4.7 Mbp) | 56.4 MiB | 64.9 MiB |

ksahlin:
I guess we wait for the numbers in the last column "**this PR**" of the indexing time table. But my overall take from this is that indexing is fast and hopefully remains relatively fast with the reduced memory footprint(?).

marcelm:
> I guess we wait for the numbers in the last column "this PR" of the indexing time table.

Oops, I pasted an old version of the table ... updated now.

ksahlin:
It looks like counting unique strobemers is responsible for near 2x slowdown compared to w/pdqsort (see herring). Only generating randstrobes should be a quick parallel task across chromosomes. Would it be possible to parallelise over chromosomes for the two randstrobe passes? This assumes, for the first pass, that it is possible to intersect the seeds for each chromosome into a total set of unique seeds.

marcelm:
The good news is that making the estimation of unique randstrobe counts multithreaded was a lot easier than I thought. The bad news is that the resulting estimates are no longer correct. They vary depending on the order in which the threads work on the contigs, and they are too large by orders of magnitude.

I’m not sure what is going on. I hope I’m just doing something wrong, but it could also be that the `HyperLogLog::merge()` method is buggy.

marcelm:
I think I fixed it! I added another column to the above runtime table.

marcelm:
> Would be interesting to see what happens for CHM13 and rye in parallell mode. Just to make sure there are no unexpected effects, such as the merging overhead overtaking sequential indexing time.

For CHM13, I tested running only the estimation part on my machine (which is possible because it only needs RAM for the reference). Runtime with 1 thread was 84 seconds and 23 seconds with 4 threads. 84/23=3.7, which is quite ok considering that other things were running on my machine at the same time.

Merging is quite fast (it’s essentially a pairwise `std::max` on two relatively short vectors), and it is only done once for each contig (plus once for each thread). CHM13 only 25 contigs, so there is virtually no overhead. Herring and fruit fly on the other hand, are good test cases because they have 6914 and 1870 contigs, respectively.

> Overall I am very positive about this approach and pull request. What is your opinion @marcelm ?

I’ve been working on this and thinking about this for way too long and would like to be able to finally work on something else. Until now, I wasn’t sure whether this approach was viable, but with this most recent change, I’m ok with the slowdown.

I’m still not fully sure where the slowdown comes from. Here is an example for herring:

Before (main):
* Generating randstrobes: 14 s
* Sorting: 6 s
* Generating hash table: 11 s
* Other stuff: 1 s
* Total: 32 s

After:
* Time estimating number of unique hashes: 17 s (or faster with more than one thread)
* Generating randstrobes and filling the hash table at the same time: 40 s
* Sorting: 1 s

I don’t get where the 40 seconds come from – it should only take a little bit longer than 25 seconds.

A couple of seconds are lost because the `RandstrobeIterator2` that generates the strobemers on-the-fly is a little bit slower than `RandstrobeIterator`. It needs ~16 seconds instead of 14. (I have tried to find out why but cannot say. But I don’t suggest using it for the reads for that reason.)

When filling the hash table on-the-fly (20db815), an additional lookup is made for each randstrobe in order to determine whether it already exists in the table (and if it does, then some extra work is necessary, but that happens only for non-unique randstrobe and therefore not often), but that extra lookup seems to be cost only about 2 seconds – I don’t know where the rest of the time goes.

> are you waiting to merge until the TODO's you added are completed or is there anything else to consider?

I would like to ignore the To Do about switching to HyperLogLog++ for the moment. Regarding different results for the RandstrobeIterator2, I think there was somehting about how randstrobes were generated at the end of a sequence. Accuracy changes a little bit; I should probably investigate:
* rye 2x150 90.22765 → 90.21695
* CHM13 2x100 93.21825 → 93.2193
* CHM13 2x200 94.4156 → 94.4089
* CHM13 2x150 94.1543 → 94.145

marcelm:
I just ran `perf` to get a better picture of where the time is spent. It appears that actually 80% of the time in `StrobemerIndex::populate` is spent in `mers_index.find()`, so that would explain it.

Perhaps it would be better to focus on optimizing `SyncmerIterator::next()`, which is responsible for 40% overall indexing time, and improving it would also benefit mapping runtime.

ksahlin:
> I don’t get where the 40 seconds come from – it should only take a little bit longer than 25 seconds.

Did [your comment](https://github.com/ksahlin/strobealign/pull/191#issuecomment-1381994339) give an answer to the discrepancy? Otherwise, wouldn't an approach such as the below one work with minimal memory overhead if several chromosomes? 

for each chromosome:
(i) generate randstrobes
(ii) fill up hash table

ksahlin:
Ah I see, you allocate space for the full hash table up front, which makes my approach still use an extra amount of memory proportional to the largest chromosome.

ksahlin:
Something with cache-inefficiency? What about chunking the iterator to spit out X randstrobes where X is something between 1 and chromosome size?

ksahlin:
Right, I see.. we dont need `mers_index.find()` in the old approach because we sort first so everything added is guaranteed to be unique. Now we need to check if the seed is already in the hash table, and that takes time. Got it.

marcelm:
> Right, I see.. we don't need `mers_index.find()` in the old approach because we sort first so everything added is guaranteed to be unique. In the new approach, we need to check if the seed is already in the hash table, and that takes time. Got it.

Actually, I noticed now that we don’t actually need the `.find()` and can instead use `.insert()` directly. Its return value tells us whether the insertion was sucessful. If the key did not exist, we’re done (that should be the case most of the time), and if it did exist, we can use the iterator that it returns to update the value. In theory, this saves us one lookup: Previously, we had a `find()` plus an `operator[]` for the insertion. Both need to look up the key. If we do an `insert()`, only one lookup should be necessary.

I just made that change and runtime remains identical because of course the compiler is smart enough to avoid the double lookup in the first place ...

marcelm:
> Something with cache-inefficiency? What about chunking the iterator to spit out X randstrobes where X is something between 1 and chromosome size?

Now we’re talking! This makes this PR as fast as the non-pdqsort version!

### Indexing time when processing in chunks

dataset | 603124d (w/o pdqsort) | 449a2fe2f128bfa3bf196b39b590f9641f72371c (w/pdqsort) | dbb30ebed9f7269f0f7cfd4d79812993cf68488 (chunked) |
-|-|-|-|
*Oryza sativa* (rice, 381 Mbp) | 21.23 s ±  0.04 s |17.51 s ± 0.05 s|21.050 s ±  0.835 s|
*Clupea harengus* (herring, 818 Mbp) | 45.12 s ±  0.90 s|33.36 s ± 0.24 s|42.208 s ±  1.350 s |
*D. melanogaster* (146 Mbp) | 7.87 s ±  0.05 s| 6.52 s ±  0.03 s|8.036 s ±  0.385 s|

The weird thing: I tried chunk size $2^{20}$ first and then lowered this more and more, but the timings *did not change*. Even with a "chunk size" of 1, the runtime is as above! Perhaps it’s something about whether code is inlined or not. We can probably get rid of the chunking when we figure out what triggers this change in performance.

ksahlin:
Whoa, nice to see! Getting closer to a merge then? Only the accuracy-investigation left I guess.

ksahlin:
I am primarily interested to understand whether there is a flaw in one of the methods. Experimentally, it looks like the differences are about 1 or 2 in 10,000(?) if you investigate 100k reads that will be on the order 10 or 20 reads. Since the difference is not consistent (diff methods are best for diff datasets), it looks like its just random variation coming from e.g. perfect repeat copies. But good to hear that you also want to understand the underlying cause.

ksahlin:
> The only difference is that RandstrobeIterator2 generates one more randstrobe at the end of each sequence. I think it is correct to do this because the mapping rate goes up slightly.

Really neat that you figured that one out. Also, maybe one more seed will have bigger impact for shorter reads (75-100nt)? But my evaluation can benchmark such things when I run it. 

I was wondering if you saw some effect in alignment runtime? Referring specifically to that seeds are not sorted by position now. While I have no reason to suspect it, I am wondering if it has any effect on merging matches into 'nams' or any other effect that I haven't thought about.
 
But yes, I am happy to merge this. Great work :)

marcelm:
I checked alignment runtime more systematically just now on fruit fly and saw no difference. Also, while working on this, I sporadically checked it also on other datasets to make sure there were no unexpected effects and it was always the same (module noise).

> But yes, I am happy to merge this. Great work :)

Thanks! I will merge this now.

PR #198, opened 2023-01-20
Add a "dumprandstrobes" command

marcelm:
This is meant for debugging. For a given reference FASTA, the command outputs the regions covered by randstrobes in BED format.

I factored out the command-line handling for the seeding parameters into a separate class so that it can be used from both binaries.


marcelm:
Maybe not so helpful in IGV, but I think it looks aesthetically pleasing ...
randstrobes

ksahlin:
It does! Makes me want to start a project to compute probabilities that at least one seed matches in a region given some mutations and parameter settings. With Benjamin we did this, sort of, for regular strobemers. But it would be interesting based on the strobealign seeds. Perhaps can lead us to 'optimal' parameter settings. Will note it down.

marcelm:
I added a commandline option `--syncmers` that can be used to dump syncmers instead of randstrobes, and I renamed the tool to `dumpstrobes`.

PR #202, opened 2023-01-25
"Speed up" reverse-complementing

### LINKED ISSUE #174: Speed up reverse complementing?
As suggested in https://github.com/ksahlin/strobealign/pull/172#issuecomment-1341184338: Can reverse-complementing be sped up with some low-level tricks?

marcelm:
I wrote a small test program to measure the differences in runtime. Reading in and reverse-complementing 1 million 150 bp reads takes 4.5 µs per read with the previous code and 2.2 µs per read with this version using a lookup table. Just parsing the FASTQ takes 1.2 µs, so reverse complementing time goes from 3.3 µs to 1.0 µs per read.

I tried to measure what the effect is on mapping time, but it makes nearly no difference (less than 1%). For the fruit fly dataset, mapping time per read went from something like 194.7±0.5µs before to 194.0±0.5 µs after the change.

It appears to me that reverse complementing is not responsible for any significant part of the mapping runtime. We currently have a caching mechanism for reverse complements, but I’m tempted to remove that and just *always* compute each read’s reverse complement because the caching itself complicates the code and also makes it slower.

Closes #174

### DIFF
--- a/src/aln.cpp
+++ b/src/aln.cpp
@@ -576,11 +576,11 @@ bool reverse_nam_if_needed(nam& n, const Read& read, const References& reference
 
     std::string seq, seq_rc;
     if (n.is_rc) {
-        seq = read.rc();
+        seq = read.rc;
         seq_rc = read.seq;
     } else {
         seq = read.seq;
-        seq_rc = read.rc();
+        seq_rc = read.rc;
     }
     std::string read_start_kmer = seq.substr(n.query_s, k);
     std::string read_end_kmer = seq.substr(n.query_e-k, k);
@@ -671,7 +671,7 @@ inline void align_SE(
         std::string r_tmp;
         bool is_rc;
         if (n.is_rc){
-            r_tmp = read.rc();
+            r_tmp = read.rc;
             is_rc = true;
         }else{
             r_tmp = read.seq;
@@ -776,7 +776,7 @@ inline void align_SE(
 
     if (!all_nams.empty()) {
         sam_aln.mapq = std::min(min_mapq_diff, 60);
-        sam.add(sam_aln, record, read.rc());
+        sam.add(sam_aln, record, read.rc);
     }
 }
 
@@ -853,7 +853,7 @@ static inline void align_SE_secondary_hits(
         std::string r_tmp;
         bool is_rc;
         if (n.is_rc){
-            r_tmp = read.rc();
+            r_tmp = read.rc;
             is_rc = true;
         }else{
             r_tmp = read.seq;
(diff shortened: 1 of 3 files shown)


ksahlin:
Hmm okay. Just to verify: 

We have a default maximum of M locations to try for each read (default 20). In worst case for a read, all M locations are tried and all of them are reverse complements. This would mean with your new code it would be Mµs for a read instead of the old (3.3+ (M-1)X)µs per read (X is the hash table lookup time for the stored rev comp read). 

Since the large majority of read will not hit this worst case I guess it is fine. There is definitely value in making the code simple.

You have my approval to do as you wish :)

marcelm:
No, we would just always keep the read and its reverse complement in memory. Then when the match is to the reverse strand, the reverse version is used and otherwise the normal one. Or do I misunderstand?

ksahlin:
Sorry, I misunderstood. Long day.. Attempt 2:

Ok I see, it is in some way the opposite of what I wrote. Each read will have both directions computed and stored. We know that at least half of the reads need to be reversed regardless in PE mode. So, compared computing RC only on a need case basis, it will in 'worst case'  be an extra X/2µs for computing RC for X reads if there are X read-pairs (this happens if all reads have a unique location).

Then, isen't what I wrote above both simpler (or as simple) and faster (assuming that the large majority of reads will have unique locations)? That is, to obtain the FW or RC string whenever needed with potential of repetitive calls?

Anyways, since this does not seem to affect runtime in anyway, I am happy with what you proposed, to avoid starting another investigation of these two approaches. You have the executive decision.

marcelm:
> Then, isen't what I wrote above both simpler (or as simple) and faster (assuming that the large majority of reads will have unique locations)? That is, to obtain the FW or RC string whenever needed with potential of repetitive calls?

We already have a [`Read` class](https://github.com/ksahlin/strobealign/blob/70b3f51ace8a478e14a4aad102e7ef24a098dd8a/src/revcomp.hpp#L24) that is used consistently for getting either the forward or reverse sequence of a read. It has a `.rc()` method that computes the reverse complement on first access. Version 0.7.1 did the same thing, but the logic was not encapsulated in a separate class. I refactored this in 4b99f3945e7673843d5319cdaa36d3545bfa8356. The benefit with this is that we don’t need to compute the reverse complement if it turns out that we never need it, but the downside is that there is the overhead of checking whether we have reverse-complemented or not.

I’ll push a commit shortly that just always computes the reverse complement when the `Read` is constructed. This actually saves another half microsecond per read.

> Anyways, since this does not seem to affect runtime in anyway, I am happy with what you proposed, to avoid starting another investigation of these two approaches. You have the executive decision.

Ok, I’ll push the mentioned commit and then merge this!

ksahlin:
perfect

PR #204, opened 2023-01-26
Suggestion to rename some key types

marcelm:
I’ve been struggling a bit with the names of some of the types and variables. Whenever I haven’t looked at the code in `index.cpp/.h` for a week, I always need to re-read what `mers_vector`, `ind_mers_vector`, `ind_flat_vector`, `ReferenceMer` actually are.

I have a couple of preliminary suggestions for which I’ve added commits in this PR. I want to get some feedback before continuing, and also want to look at the code a bit to see how the new names would "feel". I’ll update this list if I make changes.

## Changed type names

| now | suggested | comment |
|-|-|-|
| `ReferenceMer` | `RefRandstrobe` | location of a randstrobe on the reference (ref. index, position, strobe2_offset)|
| `MersIndexEntry` | `RefRandstrobeWithHash` | same as `RandstrobeOccurrence`, but stores also the hash|
| `kmer_lookup` | `RandstrobeMap` | hash table for looking up reference randstrobes by hash|
| `KmerLookupEntry` | `RandstrobeMapEntry` | result of looking up a randstrobe in `RandstrobeMap`|
| `mers_vector` | `RefRandstrobeVector` | vector of reference randstrobes|
| `QueryMer` | `QueryRandstrobe` | a randstrobe on a query sequence (hash, start, end, is_reverse)|
| `mers_vector_read` | `QueryRandstrobeVector`|

## Other changes

* Added `randstrobe_hash_t` typedef
* Removed the `ind_mers_vector` typedef and wrote `std::vector<RefRandstrobeWithHash>` directly instead (longer but clearer) where necessary.

## Possible further changes

* Some variable names should be changed to be in line with the above renames. I like the convention of having the variable name be the lowercase version of the class name (or at least a variant thereof). So `kmer_index` (which is of type `RandstrobeMap`) would become `randstrobe_map` for example. 
* The two other `...Vector` typedefs could perhaps also be removed, just like `ind_mers_vector`, because there isn’t much benefit of abbreviating e.g. `std::vector<QueryRandstrobe>` as `QueryRandstrobeVector`.

**Updated:** `RandstrobeOccurrence` → `RefRandstrobe`, `RandstrobeOccurrenceWithHash` → `RefRandstrobeWithHash`,  `RandstrobeVector` → `RefRandstrobeVector`

### DIFF
--- a/src/aln.cpp
+++ b/src/aln.cpp
@@ -130,7 +130,7 @@ void add_to_hits_per_ref(
 
 static inline void find_nams_rescue(
     std::vector<nam> &final_nams,
-    const mers_vector_read &query_mers,
+    const QueryRandstrobeVector &query_randstrobes,
     const StrobemerIndex& index,
     int k,
     unsigned int filter_cutoff
@@ -142,7 +142,7 @@ static inline void find_nams_rescue(
     hits_fw.reserve(5000);
     hits_rc.reserve(5000);
 
-    for (auto &q : query_mers) {
+    for (auto &q : query_randstrobes) {
         auto ref_hit = index.find(q.hash);
         if (ref_hit != index.end()) {
             Hit s{ref_hit->second.count(), ref_hit->second, q.start, q.end, q.is_reverse};
@@ -270,7 +270,7 @@ static inline void find_nams_rescue(
 
 static inline std::pair<float,int> find_nams(
     std::vector<nam> &final_nams,
-    const mers_vector_read &query_mers,
+    const QueryRandstrobeVector &query_randstrobes,
     const StrobemerIndex& index,
     int k
 ) {
@@ -278,7 +278,7 @@ static inline std::pair<float,int> find_nams(
     hits_per_ref.reserve(100);
 
     int nr_good_hits = 0, total_hits = 0;
-    for (auto &q : query_mers) {
+    for (auto &q : query_randstrobes) {
         auto ref_hit = index.find(q.hash);
         if (ref_hit != index.end()) {
             total_hits++;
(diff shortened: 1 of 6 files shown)


ksahlin:
I fully agree that the variable names are outdated. Furthermore, I think most of you suggestions look good. Is it intentional that you don't want to mirror `QueryRandstrobeVector` with `RefRandstrobeVector`?

marcelm:
> Is it intentional that you don't want to mirror `QueryRandstrobeVector` with `RefRandstrobeVector`?

You mean instead of `RandstrobeVector`, right?

It didn’t come to mind, maybe because I was focused so much on the index recently that I felt the reference randstrobe was the "normal" case and the query randstrobe the "unnormal" one (and should therefore get the qualifier).

Admittedly, `RandstrobeVector` is a bit inconsistent and actually using the rule that a type `std::vector<Name>` should be aliased to `NameVector`, it should be `RandstrobeOccurrenceVector`. If we want it to be `RefRandstrobeVector`, then `RandstrobeOccurrence` should also be `RefRandstrobe`. Together with the existing `Randstrobe` class, we would have this:

* `Randstrobe`: (hash, pos1, pos2)
* `RefRandstrobe`: (ref_index, pos1, strobe2_offset)
* `RefRandstrobeWithHash`: (ref_index, pos1, strobe2_offset, hash)
* `QueryRandstrobe`: (hash, start, end, is_reverse)

I think I like this. I’ll leave this for today, but make the changes later and see how it looks.

ksahlin:
> You mean instead of RandstrobeVector, right?

Yes I did 

> I think I like this. I’ll leave this as-is for today, but make the changes later and see how it looks.

I like your revised names too.

marcelm:
I have made the changes. I think it looks good now. We can change things later again if we find somethin we don’t like, but I think it’s a really good start.

ksahlin:
Looks good

PR #205, opened 2023-01-26
Use fast uppercase function also for reads

marcelm:
Gives ~0.5% faster mapping (or less)


ksahlin:
["många bäckar små"](https://www.google.com/search?q=m%C3%A5nga+b%C3%A4ckar+sm%C3%A5+engelska&oq=m%C3%A5nga+b%C3%A4ckar+sm%C3%A5+&aqs=chrome.1.69i57j0i512l2j0i22i30l2j0i15i22i30j0i22i30l4.7029j0j7&sourceid=chrome&ie=UTF-8)

PR #208, opened 2023-01-27
Strip the /1 and /2 suffixes from read names

### LINKED ISSUE #141: The /1 and /2 suffixes should be stripped from query names
A FASTQ record with header `@readname/1` or `@readname/2` should be named `readname` in the SAM output.

The SAM spec says:
> QNAME: Query template NAME. Reads/segments having identical QNAME are regarded to come from
the same template.
That is, if we want paired-end reads to be considered coming from the same template, then they need to get the same QNAME.

BWA-MEM does this.

marcelm:
Closes #141


marcelm:
@ksahlin Are you fine with merging this?

ksahlin:
yes!

Issue #206, opened 2023-01-27
Suppress output of unaligned reads?


sjaenick:
Hi @ksahlin and @marcelm,

would you consider adding a command line option to suppress output for unaligned
reads, i.e. similar to bowtie2's `--no-unal`?

marcelm:
That shouldn’t be too hard to add if @ksahlin thinks it’s worth to have this (you can of course always pipe into `samtools view -F 4`).

ksahlin:
Sure, why not. If it can be implemented without introducing overhead (which I assume it can with pretty much an if-statement).

sjaenick:
For some PE test data with many unmapped reads, this can speed up things to a certain extent:

strobealign -t 20 -N 5
real    1m47,766s
user    17m27,093s
sys     1m28,432s
~29G SAM output

strobealign -t 20 -N 5 -U
real    1m25,848s
user    16m54,221s
sys     0m39,950s
~5.9G SAM output

marcelm:
Interesting that formatting these SAM records takes so much time. This reminds me that issue #63 is still open.

PR #210, opened 2023-01-28
Add commandline flag to suppress output of unmapped reads

sjaenick:
This PR introduces a -U parameter to suppress the output of unmapped reads.

See https://github.com/ksahlin/strobealign/issues/206

### DIFF
--- a/src/aln.cpp
+++ b/src/aln.cpp
@@ -614,7 +614,8 @@ inline void align_SE(
     const References& references,
     AlignmentStatistics &statistics,
     float dropoff,
-    int max_tries
+    int max_tries,
+    bool output_unmapped
 ) {
     auto query_acc = record.name;
     Read read(record.seq);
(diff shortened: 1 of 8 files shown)


marcelm:
Thanks! I think I would prefer to make the `Sam` class responsible for knowing whether unmapped reads should be output. Then the `if (output_unmapped)` check would only have to be in `Sam::add_unmapped` and the `align_SE/_SE_secondary_hits/_PE` functions don’t need to change at all.

sjaenick:
Good idea, I updated the PR.

marcelm:
Great, looks really good now.

@ksahlin are you fine with this taking one of the possible single-letter command line options? (There are only 52 of those ...)

ksahlin:
Sounds good to me. Thanks @sjaenick!

marcelm:
This also needs a changelog entry, but because adding it to this PR would likely conflict with #211, I’ll just add that to #211. Thanks again!

PR #211, opened 2023-01-28
Documentation and changelog

marcelm:


marcelm:
Note also the extra paragraph in the "Usage" section where I recommend to pipe the output directly into `samtools view` or `samtools sort`.

ksahlin:
I really like the changes you have made to the README!

ksahlin:
I was also debating whether to remove the long "Benchmark" section because it clutters the readme. On the other hand it is nice showing some convincing stats. What do you think?

marcelm:
I didn’t dare suggest it ... I also think it’s a bit long. Maybe move it to a separate file? Perhaps a nice graph can remain in the main README?

ksahlin:
I thrive on 'harsh' feedback, that is how i level up :)

Ok, so then let us move it to a separate file.

PR #213, opened 2023-01-28
Add support for (1) interleaved reads & (2) reading from pipes

luispedro:
I was exploring adding support for strobealign in [NGLess](http://ngless.embl.de/) and it would enable some optimizations if (1) it supports the interleaved reads format, including mixing paired and single-end reads and (2) reads can be piped in.

Adding support for the former is actually quite easy and I feel the result is slightly cleaner (less repeated code)

Adding support for piping in reads is a bit more complex because of the read length estimation. But the complexity is all hidden inside the InputBuffer/RewindableFile classes.


marcelm:
Cool, this would be nice to have! I’ll review the code as soon as I can. I’m happy you were able to get rid of some code duplication as well.

Just a note: I usually prefer smaller, self-contained commits (max. one refactoring, bug fix, added feature per commit), but it’s ok now. (No need to change anything now.)

src/pc.cpp
```diff
@@ -31,16 +31,64 @@ void strip_suffix(std::string& name) {
     }
 }
 
-size_t InputBuffer::read_records_PE(std::vector<klibpp::KSeq> &records1, std::vector<klibpp::KSeq> &records2, AlignmentStatistics &statistics) {
+void distribute_interleaved(
+    std::vector<klibpp::KSeq>& records,
+    std::vector<klibpp::KSeq>& records1,
+    std::vector<klibpp::KSeq>& records2,
+    std::vector<klibpp::KSeq>& records3,
+    std::optional<klibpp::KSeq>& lookahead1) {
```
marcelm, line 39: Can you please put the `) {` on a separate line? It helps me see where the function parameters end and the function body begins.

src/pc.cpp
```diff
@@ -31,16 +31,64 @@ void strip_suffix(std::string& name) {
     }
 }
 
-size_t InputBuffer::read_records_PE(std::vector<klibpp::KSeq> &records1, std::vector<klibpp::KSeq> &records2, AlignmentStatistics &statistics) {
+void distribute_interleaved(
```
marcelm, line 34: Can you please add a brief comment describing what this does?

src/pc.hpp
```diff
@@ -23,19 +24,34 @@ class InputBuffer {
 
 public:
 
-    InputBuffer(input_stream_t& ks1, input_stream_t& ks2, int chunk_size)
-    : ks1(ks1), ks2(ks2), chunk_size(chunk_size) { }
+    InputBuffer(std::string fname1, std::string fname2, int chunk_size, bool is_interleaved)
+    : ks1(open_fastq(fname1)),
+    ks2(fname2 == "" ? std::nullopt : std::optional{open_fastq(fname2)}),
```
marcelm, line 29: `std::optional` is nice. We need to use it more often in the code base (we only recently switched to C++).

tests/run.sh
```diff
@@ -25,6 +25,11 @@ strobealign -x tests/phix.fasta tests/phix.1.fastq | tail > phix.se.paf
 diff -u tests/phix.se.paf phix.se.paf
 rm phix.se.paf
 
+# Single-end PAF (stdin input)
+cat tests//phix.1.fastq | strobealign -x tests/phix.fasta - | tail > phix.se.paf
```
marcelm, line 29: `//` → `/`

src/fastq.cpp
```diff
@@ -1,10 +1,66 @@
...
 input_stream_t open_fastq(std::string& filename) {
-    gzFile fp = gzopen(filename.c_str(), "r");
-    if (fp == nullptr) {
-        throw InvalidFile("Could not open FASTQ file");
+    if (filename == "-") {
+        filename = "/dev/stdin";
     }
-    // 16384 is the default buffer size. We need to pass it to be able to pass gzclose.
-    return klibpp::make_ikstream(fp, gzread, 16384, gzclose);
+    RewindableFile* rf = new RewindableFile(filename, true);
```
marcelm, line 61: Can you avoid the "naked" `new` here? Can you just return a ` RewindableFile` instance instead of a pointer to it (or use a `unique_ptr` if you must)? Then you wouldn’t need a `RewindableFile` destructor either.

luispedro: The previous code also used a naked pointer and a naked `malloc()`, it just hid it behind a typedef (for `gzFile`) and the `gzopen` function

A `unique_ptr` may do the trick, but this is fundamentally a `RAII`-style instance and not a value type. The cleanest may be, however, to move everything into the `InputBuffer` constructor because then it's more clearly symmetrical: `new` in the constructor, `delete` in the destructor.

luispedro:
Okay, I resolved the conflicts and cleaned up the history now. The design is cleaner too: there is a single `RewindableFile` and the `KStream` is embedded within it so that the two cannot get out of sync.

src/pc.cpp
luispedro: done

src/pc.cpp
luispedro: Added

src/main.cpp
```diff
@@ -82,13 +95,17 @@ int run_strobealign(int argc, char **argv) {
...
     if (opt.c >= 64 || opt.c <= 0) {
         throw BadParameter("c must be greater than 0 and less than 64");
     }
+
+    InputBuffer input_buffer = get_input_buffer(opt);
+    if (!opt.r_set && !opt.reads_filename1.empty()) {
+        opt.r = estimate_read_length(input_buffer);
+        logger.info() << "Estimated read length: " << opt.r << " bp\n";
+    }
+    input_buffer.rewind_reset();
```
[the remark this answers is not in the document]
luispedro: The API is not very generic, to be clear, but it serves this purpose: you can (and you must, unless you are happy to just keep a copy of all the input in memory) rewind the input exactly once

CHANGES.md
```diff
@@ -35,6 +35,8 @@
 * Suppress some logging output by default.
 * Added option `-v` for showing full logging output.
 * Issue #206: Added option `-U` for suppressing output of unaligned reads. Thanks @sjaenick.
+* Add support for interleaved reads
```
marcelm, line 38: Feel free to add a "Thanks @luispedro" on these two bullet points.

marcelm (commented):
Thanks, I like this version!
@ksahlin I would like to merge this if it’s ok with you.

ksahlin:
Thank you for this PR @luispedro ! Yes @marcelm, I am happy to merge this.

> Feel free to add a "Thanks @luispedro" on these two bullet points.

Yes, I think its time to properly add credits to contributors in this repo (and elsewhere).

CHANGES.md
```diff
@@ -35,6 +35,8 @@
...
+* Add support for reading from a pipe (including stdin)
```
marcelm, line 39:
```suggestion
* PR #213: Add support for interleaved reads. Thanks @luispedro.
* PR #213: Add support for reading from a pipe (including stdin). Thanks @luispedro.
```

marcelm:
Thanks again!

PR #220, opened 2023-01-31
Changelog for v0.8.0

marcelm:
The last remaining to do before the release.


marcelm:
I added a sentence about NBIS and hope that is ok.

ksahlin:
Yes, great.

PR #225, opened 2023-02-02
Show a progress indicator

marcelm:
I don’t know if this is your style, but I have added something similar to other programs, and I found it made the output just a tad friendlier.

During mapping, this progress indicator displays the number of mapped reads and the average (wall-clock) time per spent per read. 

Because the runtime per read fluctuates a bit in the beginning, the progress indicator is shown only after the first second to ensure we don’t have misleading numbers on the screen in case only a small number of reads were mapped.

Also, the progress indicator is only shown if the program is run interactively (when stderr is an actual terminal) to avoid cluttering log files when the user redirects stderr.

Totally fine with me if you don’t want this.

Animated preview:
[asciicast](https://asciinema.org/a/JRtBxNjRvTxph3IJpBe4P6mAB)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,10 @@
 # Strobealign Changelog
 
+## development version
+
+* Add progress report (only shown if output is not a terminal and can be
+  disabled with `--no-progress`)
+
 ## v0.8.0 (2023-02-01)
 
 ### Summary
--- a/src/aln.hpp
+++ b/src/aln.hpp
@@ -59,6 +59,7 @@ struct AlignmentStatistics {
     std::chrono::duration<double> tot_extend;
     std::chrono::duration<double> tot_write_file;
 
+    unsigned int n_reads = 0;
     unsigned int tot_ksw_aligned = 0;
     unsigned int tot_rescued = 0;
     unsigned int tot_all_tried = 0;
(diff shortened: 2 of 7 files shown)


ksahlin:
It looks cool!

How much does it slow down the mapping process? If < 1% it's totally fine. I guess one could alwaus increase the step size from 50k to 200k or 500k similar to BWA/minimap2. 

(Also, note that the extension alignment is 6.44 out of 10s for that data :)

marcelm:
I don’t expect there to be a slowdown at all. This is implemented in a loop in the main process that spends most of its time sleeping. Every second, it wakes up, collects stats from the threads and prints it. I’m sure there is no measurable overhead.

> Also, note that the extension alignment is 6.44s out of 10s for that data :)

Yes, yes, I’ve started to work on this in the wfa2 branch ... I wanted to get the release out first.

ksahlin:
> Yes, yes, I’ve started to work on this in the wfa2 branch ... I wanted to get the release out first.
 
Hehe sorry, eager-kid-at-christmas mode 😃

marcelm:
So ... too flashy or can I merge this?

ksahlin:
Yes, merge is good. And I like the process indicator.

PR #226, opened 2023-02-03
Fix has_shared_substring omitting some l-mers

marcelm:
The iteration stopped a bit too early and therefore did not generate all possible l-mers.

`has_shared_subtring` is a heuristic used in `rescue_mate`. If query and reference do not share any substrings of length $l=\frac23k$, the read is marked as unmapped and no rescue is attempted.

This increases accuracy slightly for very short reads (50 bp).
* CHM13: 90.46475 → 90.48315
* fruit fly: 90.0643 → 90.0838

This changes how often rescue and alignment is done. For CHM13 (2x50):

* Total calls to ssw: 1972617 → 2003581
* Calls to ksw (rescue mode): 1257900 → 1288864

So probably runtime increases (the increase in ssw calls is much lower for the higher read lengths).

It appears to me that the behavior is a bug and should be fixed. If there’s a reason to not look at the last l-mers in the read, then this should be documented. If we want to avoid the runtime penalty, we should think about other ways of improve the heuristic.

But maybe this discussion becomes moot when WFA2 is integrated ... We can just leave this PR upon until then.

### DIFF
--- a/src/aln.cpp
+++ b/src/aln.cpp
@@ -695,7 +695,7 @@ bool has_shared_substring(const std::string& read_seq, const std::string& ref_se
     int sub_size = 2 * k / 3;
     int step_size = k / 3;
     std::string submer;
-    for (size_t i = 0; i + k <= read_seq.size(); i += step_size) {
+    for (size_t i = 0; i + sub_size < read_seq.size(); i += step_size) {
         submer = read_seq.substr(i, sub_size);
         if (ref_seq.find(submer) != std::string::npos) {
             return true;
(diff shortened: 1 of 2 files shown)


ksahlin:
Yes, it seems like a bug that should be fixed. But I agree, let us keep it open until wfa2. Also, I wonder why I chose 2k/3. Seems for this targeted search we could afford lower l, like k/2.

marcelm:
As discussed on Tuesday, it’s time to merge this.

I have measured the runtime on a couple of read lengths on the fruity fly and human genomes and have not found a significant difference in speed.

Issue #26, opened 2022-06-01
Base level alignment


ksahlin:
Note to developer:

The extension step (nucleotide level alignment) is the bottleneck in strobealign. There are different three ways to reduce this:
1. Direction 1 (change the alignment module): 
	1. Change to base level alignment with [WFA](https://github.com/smarco/WFA2-lib) ([WFA publ](https://academic.oup.com/bioinformatics/article/37/4/456/5904262#246118479)) as is done in [Accelalign](https://github.com/raja-appuswamy/accel-align-release)
3. Direction 2 Speedup the current module used (SSW):
	1. By using 8bit slots in alignment matrix?
	2. By not computing alignment twice - is ssw does this?
4. Direction 3 (partitioned SSW)
	1. Finish implementing partitioned SW (split alignment into several small hamming or SW alignments) if seeds are in middle.

ksahlin:
I think this is worth exploring WFA as we are currently relying on SSW which has its issues being a local aligner as mentioned in issue #54. The method seems to have great performance, see table2 in [WFA paper](https://academic.oup.com/bioinformatics/article/37/4/456/5904262?login=true#284254896) for timings to e.g. SeqAn and ksw2 in Table 2. Furthermore, the maturity of the implementation is here with [WFA2](https://github.com/smarco/WFA2-lib) in terms of providing different penalty models (including dual gap cost penalties!), traceback cigar etc,

Also in strobealign the extension step is over 50% of the runtime for many biological datasets, see 'aln' field for BIO150 and BIO250 in attached figure for extension with SSW.

<img width="551" alt="image" src="https://user-images.githubusercontent.com/1714667/189323313-fe55725d-5b82-4a16-b07b-ca0bc8ca9c2e.png">

ekg:
This will also make it easy to scale to longer sequences, provided your seed chaining can do it. BiWFA uses order divergence space and is consequently very cache coherent and actually fast even for ~500bp sequences.

jianshu93:
Hi all, just noticed this. I do not think WFA is more efficient when it is the only option. As a general purpose mapper, like bwa, mini-bwa, bowtie2, based level extension is unavoidable because sequences can be less than 95% identity to each other, below which WFA/WFA2 lose its advantage completely. I would suggest using metagenomic data for the test. Not only human genomes as tests. Remember human genomes is only less than 0.1% of all genomes we have. I think SIMD SSW is a good for now.

ksahlin:
[...]

PR #228, opened 2023-02-06
Various refactorings

marcelm:
Here are some refactorings of the alignment code. I’m doing this as part of #26 "cleaning up as I go" while trying to understand how the alignment/mapping is done.

* Change `HammingToCigarOQX2` so that it returns an `aln_info` struct to make it more compatible with `ssw_align` (also rename it to `hamming_align`)
* Move all NAM-related stuff to `nam.cpp/.h` (this was very helpful as it made it clear that `hit` and `Hit` are only used privately by the nam-finding functions)
* Move all PAF-output-related stuff to `paf.cpp/.h`
* Convert to 1-based coordinates at the last possible moment when writing the SAM output. This way, the details of the file format do not leak into the alignment code. This makes it simpler, for example, to switch from SAM to BAM output
if we wanted to because BAM uses 0-based coordinates.

With the above, `aln.cpp` went from 2340 to 1900 lines.


src/aln.cpp
```diff
@@ -27,7 +27,7 @@ static inline bool score(const nam &a, const nam &b) {
     return a.score > b.score;
 }
 
-inline aln_info ssw_align(const std::string &ref, const std::string &query, int match_score, int mismatch_penalty, int gap_opening_penalty, int gap_extending_penalty) {
+inline aln_info ssw_align(const std::string &ref, const std::string &query, alignment_params parameters) {
```
ksahlin, line 19: Is it nicer/faster to send alignment_params by reference?

ksahlin:
This looks great!

src/aln.cpp
marcelm: I thought a bit about this when making that change before I settled on this variant. `alignment_params` consists of four 32-bit fields, so they can be passed using two 64-bit registers. When they are passed by reference, only one 64-bit register is needed for the memory address. On the other hand, then a memory access is needed within the function, which is slower than accessing a register. In the end, I am sure this does not make a difference at all and both variants are fine.

ksahlin: thanks for the explanation!

PR #229, opened 2023-02-06
Alignment with WFA2

marcelm:
This is unfinished and currently *slower* than what we have in main: 33µs vs 21µs per read.

I wanted to get it to work first and did not spend so much time on making this either nice or fast, so there’s a chance we’ll be faster in the end, see the to do items below.

At the moment, WFA2lib does not fit so well into how strobealign computes alignments. Strobealign cuts out the region of the reference to which the query maps plus 50 bp added upstream and downstream and then asks the aligner to do a local alignment. The normal case is that most of bases of the query are aligned to the reference, but if one or both ends of the query don’t match (due to adapter contamination, the read being chimeric or containing a structural variant etc.) we can get soft clipping because of the local alignment.

WFA2 does not support local alignment, only ends-free alignment. I am using this here, allowing both ends of the reference sequence to be deleted at no cost, but the read itself is always aligned end-to-end, without soft clipping. There’s a mode called "extension alignment" that’s intended for seed-and-extend algorithms. That may allow soft clipping, but I am not sure. This would also need changing the alignment so that it is computed in parts (before the seed, seed itself, after the seed), as I think you started to implement it in some commented-out code in `ssw_align`. For a fair comparison, we would need to implement this for StripedSmithWaterman first.

* [ ] Do not create a new `WFAlignerGapAffine` each time `ssw_align` is called. This is explicitly advised against in the WFA2-lib documentation as being inefficient.
* [ ] Implement piecewise alignment
* [ ] Play around with the heuristics
* [ ] Make `fixwfa2cigar` more efficient (and less ugly)


ksahlin:
Nice investigation! 

> (Implement piecewise alignment) as I think you started to implement it in some commented-out code

Yes, I started the piecewise alignment and it sort of works (and is also faster than current approach). However, what made me drop this at the time was that I couldn't get it to produce the same result as semi-global for larger indels.

To make the piecewise approach provide identical results to current approach, (I think) we need a way to pass (initialise) the first row/column with a "current SW-score". Initialising to 0 does not seem to work with the current dropp-off heuristic in SSW.
 
Example: Imagine we have a 150nt query with a 50nt deletion w.r.t the reference, where the query read has cigar 100=50D50=. We would align through the delation with a full SSW, but if we split the alignment at, say, position 75 in the read,  we would only get the left extension. The right side with fail to align through the deletion, because of a SSW heuristic that drops the alignment if too low score. This Drop-off heuristic is nice to speed things up, but result in suboptimal alignment if piecewise aligned. 

Therefore, I had the idea to use the score from other pieces, but I am unsure that even works (e.g., which score to compute and pass first..?) 

>  There’s a mode called "extension alignment" that’s intended for seed-and-extend algorithms.

That sounds interesting (although perhaps the same drawback as above?)

ksahlin:
In case it helps, WFA is implemented in AccelAlign here: https://github.com/raja-appuswamy/accel-align-release/blob/905cbf1c80685494ffd2fbc6e9c92b445a832434/accalign.cpp#L2044

They seemed to see a quite significant speedup compared to KSW2 at least, see [this table](https://bmcbioinformatics.biomedcentral.com/articles/10.1186/s12859-021-04162-z/tables/6). (compare e.g. NO-KSW2 and NO-WFA, (NO stands for no embedding, embedding is their unique approach)

ksahlin:
As for the actual piecewise alignment, I think my implementation fell short of only breaking it into two parts.

If three parts, the two extension alignments on 'each side' could benefit from being aligned with an initial score obtained from the middle segment (the NAM seed itself). In such case, I think the piecewise alignment should have the potential of being similar to a glocal (semi-global) alignment with a dropoff score (details to be hashed out). Perhaps you have better ideas or understanding than me.

One benefit that comes for free with the piecewise alignment is that we could test hamming distance on each piece. It is possible that only a small segment has an indel, so if we can get away with hamming distance comparison in two of the three pieces we are already much better off than our current approach.

marcelm:
I’m at home now and measured again. The original version of this PR is 7 times slower than main on the machine I have here. I measured on my work laptop yesterday.

I have pushed a commit that avoids re-creating the aligner for every alignment. With it, runtime is 4.5 times slower than main.

I’ll need to look into what AccelAlign does a bit closer. They directly use the C interface while I have been using the C++ bindings. This should not make a difference performance wise. It appears they don’t support soft clipping in that function.

I have a couple of smaller things to try, but I am currently not convinced this is going to be faster than what we have. Although I am kind of hoping I made a simple, easy to fix mistake.

sjaenick:
https://academic.oup.com/bioinformatics/advance-article/doi/10.1093/bioinformatics/btad074/7030690?login=false

Just gonna leave this here, as it seems relevant

ksahlin:
Thanks @sjaenick

Yes it's a great paper, read it in detail :) 

I don't think we need the space optimised version presented in that paper though, the [original WFA](https://academic.oup.com/bioinformatics/article/37/4/456/5904262) seems fine for short reads.

marcelm:
WFA2 has an optimization that is only possible when the match score is zero and, as far as I understand it, the evaluations in the paper were made with that choice. Our match score is 2. When I change that to zero, runtime is reduced further to 2.0 times slower than main.

PR #231, opened 2023-02-07
Add support for gzipped references

luispedro:
Since `zlib.h` was already used, this does not require any new dependencies, but it was easier to change the FASTA parser to work _char by char_ as it matches the `zlib.h` structure better.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@
 
 * Add progress report (only shown if output is not a terminal and can be
   disabled with `--no-progress`)
+* Support gzip-compressed references. Thanks @luispedro
 
 ## v0.8.0 (2023-02-01)
 
--- a/src/refs.cpp
+++ b/src/refs.cpp
@@ -3,52 +3,115 @@
 #include <fstream>
 #include <sstream>
 #include <algorithm>
+#include <zlib.h>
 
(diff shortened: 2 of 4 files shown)


marcelm:
Nice, this was actually on my to-do list!

I benchmarked this and unfortunately, this version is three times slower than the current one (for uncompressed input), presumably because it works character by character. It’s only a couple of seconds at start-up time, but I still don’t think this would be a regression we would be ready to accept. Do you think you can improve that? We could also use a C++ wrapper for zlib such as [zstr](https://github.com/mateidavid/zstr).

luispedro:
The new commit makes it as fast (maybe even faster) than `main` at the cost of more complex code

marcelm:
Thanks. Have you benchmarked this? For me, this new version is still a tad slower than the one in main. Reading CHM13 (3.1 Gbp) now takes 3.0 seconds vs 2.5 seconds before. I played around with the code a bit, but don’t know why that is. BTW, your buffer size can be much smaller; at least for me, there’s no difference in runtime when I reduce the size to 256 bytes.

I want to say that it doesn’t matter that it’s a bit slower, but then that it’s a lot more complex at the same time makes me hesitant. Manually managing a buffer like this shouldn’t be necessary in C++ for just parsing a text file.

PR #234, opened 2023-02-10
More refactoring and a minor bug fix

marcelm:
This is a large diff, but most of this is restructuring, with one exception: `align_segment` used to receive a `ref_segm` and a `ref_segm_len` parameter. I removed the length parameter because it just `ref_segm.length()`. This changes the output in a couple of cases because it turns out that the passed `ref_segm_len` was sometimes incorrect (when the segment is close to the end of the contig).

Other than that, the biggest change is that I added an `Aligner` class. Instead of `ssw_align`, one now calls `aligner.align()`. This is nice in several ways:
- The aligner can be more easily replaced
- Creating a single aligner and reusing it whenever possible speeds up WFA2 (it does not make a difference for SSW)
- The aligner itself can track how often it is called, which removes quite a few lines manipulating `tot_ksw_aligned`.


ksahlin:
Looks great!

PR #238, opened 2023-02-14
Fix occasional incorrect soft clipping

marcelm:
I noticed that some reads have CIGARs like `186=115S` although they start with an `N` character that should definitely be a mismatch. It turns out that the loop in the `align_hamming` function that creates the CIGARs was slightly buggy.

The problem was less obvious because it was obscured by the soft-clip algorithm being a bit pessimistic. In many cases, it returned an empty alignment although a better solution was possible. In that case, the regular alignment algorithm was triggered, wich then did correct soft clipping.

To fix this, this commit changes the soft clip detection algorithm. Instead of detecting where the maximum of the partial sums of the scores is (from either direction), it tries to find a segment in the alignment that has maximal sum of scores. This works better for more complicated situations where, e.g., both ends need to be soft clipped. I’m sure the algorithm is described in the literature somewhere as "maximum xyz segment problem", but I couldn’t find it. The idea that helped me come up with this is to consider the problem a special case of the Smith-Waterman algorithm, but where we only look at one diagonal.


ksahlin:
Nice catch!

PR #239, opened 2023-02-14
Use get_alignment() in align_SE_secondary_hits

marcelm:
This removes duplicated code, giving us one function that is used from both single-end and paired-end mode responsible for extending seeds.

Results are not 100% identical. I’ll need to find out how accuracy changes to see whether that means it has gotten better or worse.


marcelm:
When I tried to measure accuracy, I noticed that the results on rackham and my own computer were different for the exact same commit and the same dataset. It turns out that the `global_ed` member of the `alignment` struct is not initialized, which means that it essentially has a random value at runtime. Since `align_SE_secondary_hits` makes decisions based on `global_ed`, this leads to nondeterministic behavior (I think depending on the compiler version in this case).

I pushed a commit to fix that. I also pushed [86086b6](https://github.com/ksahlin/strobealign/pull/239/commits/86086b6d8cf1114117734ad5c8861d758500aba3), which fixes another small issue with the way the hamming distance is computed (and simplifies the code a bit further).

Here’s how the accuracy changes for single-end reads on some datasets.

dataset | before | after
-|-|-|
fruit fly 1x50 | **82.0218** | 82.0217 |
fruit fly 1x100 | 89.1783 | **89.1824** |
fruit fly 1x150 | 90.8951 | **90.904** |
fruit fly 1x300 | **93.2264** | 93.2247 |
CHM13 1x50 | 81.127 | **81.1337** |
CHM13 1x100 | 90.5423 | **90.5432** |
CHM13 1x150 | 92.369 | **92.3735** |
CHM13 1x200 | 93.2037 | **93.2063** | 
CHM13 1x300 | **94.1506** | 94.1502 |
rye 2x150 | **80.3848** | 80.3844 |

ksahlin:
Great!

PR #243, opened 2023-02-21
Clean up NAMs a little bit

marcelm:
The main change in this PR is to factor out a `merge_hits_into_nams` function that removes a nice chunk of duplicated code from `find_nams` and `find_nams_rescue`. Also:
* Rename `nam` to `Nam`
* Move `Hit` struct declaration into `find_nams_rescue` as it is used only there (and rename to `RescueHit`)
* Rename `hit` to `Hit`


ksahlin:
Yes, that `merge_hits_into_nams()` was a nice factorisation!

PR #246, opened 2023-02-24
Add a "Contributing" section to the README

marcelm:
@luispedro Here’s a description of how to run one of the testing scripts.


ksahlin:
I think the subsections should probably be with '###'. Now they have the same font size as main sections which is a bit confusing in the contributing section.

(Didn't catch this before because I did not read the markdown rendered version.)

PR #249, opened 2023-02-28
Some harmless refactoring

marcelm:
These are the refactoring commits extracted from PR #242. I think they are helpful no matter whether #242 gets merged or not.

In particular:
* Everything that produces an `aln_info` object is moved to `src/aligner.hpp/`.cpp`
* Add a `Cigar` class (this uses the same representation of CIGAR operations as used in BAM files, so this is also a step towards supporting BAM output)


ksahlin:
great!

Issue #222, opened 2023-02-01
Installation works only from the Git repository, not from the release archive


marcelm:
Since we now compute the version number by inspecting the Git history at build time (with cmake_git_version_tracking), building from the release archives (tarball/zip file) no longer works.

Perhaps this works: We could include a pre-generated `version.hpp` file in the release archives and then somehow not run the Git history CMake module if no `.git` directory was found.

PR #255, opened 2023-03-15
Support reading gzip-compressed references

marcelm:
This PR uses the zstr external library for this. It wraps zlib in a C++-compatible way so that we can use the FASTA parsing function unchanged.

While zstr can also read uncompressed input, it is a bit slower doing so than when using a regular std::ifstream. To avoid a performance regression, we use zstr::ifstream only if the input file name ends with `.gz` but fall back to std::ifstream otherwise.

Supersedes and closes #231

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,10 +11,10 @@
   without soft clipping are compared more fairly. Use `-L` to change the end
   bonus. (This emulates a feature found in BWA-MEM.)
 * Issue #238: Fix occasionally incorrect soft clipping.
-* PR #239: Align single- and paired-end reads in the same way.
 * PR #239: Fix an uninitialized variable that could lead to nondeterministic
   results.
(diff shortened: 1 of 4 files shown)


marcelm:
@ksahlin Would it be ok to make a new release? A couple of improvements have been made since 0.8.0 and it would be nice to show to the outside world that we’re still working on this.

ksahlin:
Yes, great let's do it. I have been/am swamped with 3 revision deadlines in the coming week, so I have lost track a bit of where we are. I assume WFA+ksw2 wont make it into this release? But after March 23rd I wanted to rerun the indel evaluation with `bcftools norm`.

marcelm:
Alright, I’ll take care of a new release. I noticed I need to fix #222 first, so it’ll take some time. WFA+ksw2 won’t make it, but that’s ok. (And no worries, I’ve also been busy with other things.)

ksahlin:
Sounds good!

PR #256, opened 2023-03-15
Make release tarball installable

### LINKED ISSUE #222: Installation works only from the Git repository, not from the release archive

marcelm:
This replaces cmake-git-version-tracking with with a custom solution. The behavior is as follows:
- If building from a cloned Git repository, `git describe` is used to set the version number based on the most recent Git tag. This is the case during development and gives us version numbers like `v0.8.0-112-gc7aa34f-dirty` ("dirty" means that there are uncommitted changes)
- If building from a release tarball/zip file, `git describe` wil fail, and the version set in `CMakeLists.txt` is used instead.

This means that releasing a new strobealign version requires tagging the version number in Git *and* updating it in `CMakeLists.txt`.

Based on:
- https://github.com/nocnokneo/cmake-git-versioning-example
- https://github.com/andrew-hardin/cmake-git-version-tracking/
- https://jonathanhamberg.com/post/cmake-embedding-git-hash/

Closes #222


ksahlin:
That sounds great!

PR #258, opened 2023-03-20
Fix compilation with MinGW

teepean:
Small fix to fix an error with MinGW. Without the fix aln.cpp has the following error:

```
C:/msys64/home/User/strobealign/src/aln.cpp:260:39: error: no matching function for call to 'min(long unsigned int, std::__cxx11::basic_string<char>::size_type)'
  260 |         const int ext_right = std::min(50ul, ref.size() - n.ref_e);

```
Fix compilation with MinGW. Credits Biswa96

### DIFF
--- a/src/aln.cpp
+++ b/src/aln.cpp
@@ -254,7 +254,11 @@ static inline alignment get_alignment(
         const int diff = std::abs(n.ref_span() - n.query_span());
         const int ext_left = std::min(50, projected_ref_start);
         const int ref_start = projected_ref_start - ext_left;
+#if defined (__MINGW32__)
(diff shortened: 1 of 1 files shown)


src/aln.cpp
```diff
@@ -254,7 +254,11 @@ static inline alignment get_alignment(
         const int diff = std::abs(n.ref_span() - n.query_span());
         const int ext_left = std::min(50, projected_ref_start);
         const int ref_start = projected_ref_start - ext_left;
+#if defined (__MINGW32__)
```
marcelm, line 257: Thanks! There’s no need for this to be conditional, just replace the code with the version that works for MinGW.

marcelm:
Thanks!

PR #260, opened 2023-04-03
mod to display cmdline options to sam header

telmin:
Hi.

I am in the process of evaluating strobealign. I think it is very nice software.

By the way, the @PG tag in the SAM header, the spec says that the CL tag should show the command line options when run the analysis.
So I thought it was correct to modify it this way.

I would be happy if you could merge them.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -29,7 +29,7 @@ static Logger& logger = Logger::get();
 /*
  * Return formatted SAM header as a string
  */
-std::string sam_header(const References& references, const std::string& read_group_id, const std::vector<std::string>& read_group_fields) {
+std::string sam_header(const References& references, const std::string& read_group_id, const std::vector<std::string>& read_group_fields, const std::string& cmd_line) {
     std::stringstream out;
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -131,6 +131,11 @@ void show_progress_until_done(std::vector<int>& worker_done, std::vector<Alignme
 int run_strobealign(int argc, char **argv) {
     auto opt = parse_command_line_arguments(argc, argv);
 
+    std::stringstream cmd_line;
```
marcelm, line 134: Can you move this into the `if (map_param.is_sam_out) {` block below? Then it’s closer to where it’s used.

marcelm (commented):
Thanks! Just one comment.

src/main.cpp
telmin: Thanks for your comment. You are indeed correct. I will fix it.

telmin: I fixed it. Would you check it please.

marcelm:
Thanks again! We may need to add an option to disable adding the `@PG` header because this could lead to non-reproducible output, but that is something for later.

PR #261, opened 2023-04-11
Emit M instead of =/X CIGAR operations by default

marcelm:
I needed to do some refactoring so that CIGARs are internally always represented as `Cigar` objects and only converted to their string representation at the very last moment when the alignments are serialized to SAM format. This makes it easy to optionally convert them to use `M` instead of `=`/`X` operations since the logic for that needs to be in only one place.

I chose `--m-op` as the name for the command-line option for now, but I’m not super happy with it. Bowtie2 and minimap2 have an `--eqx` switch and the opposite behavior: `M` by default and `=`/`X` is optional. Perhaps we should change the default as well for better compatibility? Also, that would avoid the spurious variant calling issue as long as htslib is not fixed.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,8 @@
 
 * #258: Fix compilation on MinGW. Thanks @teepean.
 * #260: Include full command line in the SAM PG header.
+* #20: Add command-line option `--m-op`. If used, emit `M` CIGAR operations
+  instead of `=` and `X`.
 
 ## v0.9.0 (2023-03-16)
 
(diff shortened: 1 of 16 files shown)


ksahlin:
Great! And I agree that we should change defaut to be M and have the flag `--eqx` to write `=/X`cigar strings.

I will restart the SNV/indel calling benchmark once this is merged.

marcelm:
Alright, updated to use `M` by default, and the previous behavior can be restored with `--eqx`.

ksahlin:
The evaluation of this commit was posted here: https://github.com/ksahlin/strobealign/issues/245

PR #262, opened 2023-04-14
Refactor

marcelm:
These are two refactorings I did while working on something else.

* Simplify calling `query_randstrobes` by passing an `IndexParameters` instance instead of each individual parameter
* Add a `RandstrobeMapEntry::for_each` function that factors out the "iterate over all RefRandstrobes" functionality. This encapsulates the logic for deciding whether we have a direct entry or multiple entries that need to be looked up in the flat vector. This allows us to make a couple of functions in `RandstrobeMapEntry` private.

### DIFF
--- a/src/index.hpp
+++ b/src/index.hpp
@@ -55,6 +55,8 @@ private:
 
 using RefRandstrobeVector = std::vector<RefRandstrobe>;
 
+struct StrobemerIndex;
+
 /*
  * An entry in the randstrobe map that allows retrieval of randstrobe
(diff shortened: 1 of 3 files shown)


ksahlin:
Looks good. Will this complicate or simplify merging the new index layout which does not require the `as_direct()` check?

marcelm:
Theoretically, it could simplify it because it abstracts away the index layout, but it is probably the same in practice. I don’t think it’ll be an issue in any case.

PR #264, opened 2023-04-19
Clean up and comment randstrobes_query

marcelm:
I just understood why `randstrobes_query` re-uses the syncmers that it computed for forward sequence for the reverse complement, and why it does not do the same thing for the randstrobes, so I would like to document it. This cleans up the function a little bit and adds a couple of comments.


ksahlin:
Yes, great. I should have documented it better in the code.

PR #269, opened 2023-04-27
Use ISA-L to speed up gzip decompression

telmin:
Hi.

I speeded up the loading part of Strobealign in this PR.
In this PR, I replaced the use of zlib's gzread for gzip decompression with a class that performs gzip decompression and double buffering using ISA-L.
As a side effect, double buffering is used to read raw fastq, which also speeds up.

The benchmark for gzip decompression with gzopen is shown below.

```
Total time mapping: 2610.28 s.
Total time reading read-file(s): 2284.26 s.
Total time creating strobemers: 25.25 s.
Total time finding NAMs (non-rescue mode): 42.51 s.
Total time finding NAMs (rescue mode): 72.65 s.
Total time sorting NAMs (candidate sites): 121.26 s.
Total time base level alignment (ssw): 167.53 s.
Total time writing alignment to files: 0.00 s.
```

On the other hand, this PR will speed up

```
Total time mapping: 727.12 s.
Total time reading read-file(s): 110.86 s.
Total time creating strobemers: 47.43 s.
Total time finding NAMs (non-rescue mode): 70.84 s.
Total time finding NAMs (rescue mode): 151.78 s.
Total time sorting NAMs (candidate sites): 232.37 s.
Total time base level alignment (ssw): 324.62 s.
Total time writing alignment to files: 0.00 s.
```

This PR improves scalability in systems with a very large number of cores.

Could you please review it?

### DIFF
--- a/src/iowrap.cpp
+++ b/src/iowrap.cpp
@@ -22,6 +22,28 @@ int64_t GeneralIO::read(void* buffer, size_t length)
     return gzread(file, buffer, length);
 }
 
+void RawIO::open(const std::string& filename)
+{
+    fd = ::open(filename.c_str(), 0);
+    if(fd < 0) {
+        throw InvalidFile("Could not open FASTQ file: " + filename);
+    }
+}
+
+void RawIO::close()
+{
+    if(fd != -1) {
+        ::close(fd);
+    }
+    fd = -1;
+}
+
+int64_t RawIO::read(void* buffer, size_t length)
+{
+    ssize_t ret = ::read(fd, buffer, length);
+    return ret;
+}
+
 void IsalIO::initialize()
 {
     isal_inflate_init(&state);
--- a/src/iowrap.cpp
+++ b/src/iowrap.cpp
@@ -1,5 +1,11 @@
 #include "iowrap.hpp"
 #include "exceptions.hpp"
+#include <sys/stat.h>
+#include <sys/mman.h>
+#include <sys/types.h>
+#include <fcntl.h>
+#include <system_error>
+#include <cstring>
 
 void GeneralIO::open(const std::string& filename)
 {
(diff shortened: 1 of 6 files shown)


telmin:
I am trying to fix CI, but I can't figure out why it SEGVs on MacOS and why the build fails with pythonbindings. Can you please give me some advice?
I tried to create a similar environment locally for pythonbindings and the build passes. it may be a problem with the SOURCE_DIR variable, but I am not sure what the cause is...

marcelm (commented):
Hi, thanks for this! This is a really good start.

I am not finished with my review, but I thought it would be good to give early feedback, so I’m submitting these comments already now.

Sorry about asking for changes regarding code style. There was no way for you to know my preferences as it is currently not documented. I’m using this opportunity to write some initial coding guidelines so we have something we can point future contributors to.

I think we need to figure out some build issues:
1. Usage of ISA-L needs to be optional since it is not available everywhere (Apple Silicon for example)
2. We can tackle this later, but should keep it in mind: We need to make it possible to dynamically link to `libisal` so that an already installed one can be used. So when building the Conda package, for example, we would just add `isa-l` as a dependency.
3. The Python bindings don’t build at the moment. This one is on me, feel free to ignore that problem for now.

marcelm:
> I tried to create a similar environment locally for pythonbindings and the build passes. it may be a problem with the SOURCE_DIR variable, but I am not sure what the cause is...

For testing the Python bindings, you should compile via `pip install`. So something like this:
```
python3 -m venv .venv
.venv/bin/pip install .
```
I was able to reproduce the error when I did the above. But as I said, I will try to help with figuring out that issue.

marcelm:
>     1. Usage of ISA-L needs to be optional since it is not available everywhere (Apple Silicon for example)

Hm, it seems to be available for aarch64, so I take that back.

src/iowrap.cpp
```diff
@@ -0,0 +1,206 @@
...
+        }
+    }
+    return actual_count;
+}
+
+void RawIO::preload(size_t size)
+{
+    read_buffer_work.resize(preload_size);
+    auto actual_size = ::read(fd, read_buffer_work.data(), size);
+    if(actual_size != size) {
```
marcelm, line 75: I get a warning here:
```
warning: comparison of integer expressions of different signedness: ‘long int’ and ‘size_t’ {aka ‘long unsigned int’} [-Wsign-compare]
```

src/iowrap.cpp
```diff
@@ -0,0 +1,206 @@
...
+    }
+    return actual_count;
+}
+
+void IsalIO::decompress(size_t count)
+{
+    uncompressed_data_work.resize(std::max(count, previous_member_size));
+    uint8_t* ptr = uncompressed_data_work.data();
+
+    while(ptr - uncompressed_data_work.data() < count) {
```
marcelm, line 179: I get a warning here as well:
```
warning: comparison of integer expressions of different signedness: ‘long int’ and ‘size_t’ {aka ‘long unsigned int’} [-Wsign-compare]
```

marcelm:
Once again regarding this:

> 2\. We can tackle this later, but should keep it in mind: We need to make it possible to dynamically link to `libisal` so that an already installed one can be used. So when building the Conda package, for example, we would just add `isa-l` as a dependency.

Maybe it is actually better to always assume that libisal is available on the system. Looking at this [isal-l wiki page](https://github.com/intel/isa-l/wiki/Ports--Repos), it’s clear that libisal is available on many Linux distributions and on macOS via brew. So this is probably simpler and more reliable than using `ExternalProject`.

marcelm:
It got this working using the system isa-l by replacing `ExternalProject_*` with this:
```
find_package(PkgConfig REQUIRED)
pkg_check_modules(ISAL REQUIRED IMPORTED_TARGET GLOBAL libisal>=2.30.0)
target_include_directories(salib PUBLIC src/ ext/ ${PROJECT_BINARY_DIR} ${ISAL_INCLUDE_DIRS})
target_link_libraries(salib PUBLIC ZLIB::ZLIB Threads::Threads zstr::zstr PkgConfig::ISAL)
```
And in `iowrap.hpp`, I had to change `#include "igzip_lib.h"` to `#include "isa-l/igzip_lib.h"`. I then installed libisal with `sudo apt install libisal-dev` and strobealign compiled fine, and even `pip install .` worked.

telmin:
Thanks for the review. I think I was able to fix all the points you made.
I should have used isa-l from the package manager, as you pointed out.

I am not sure what is causing it to SEGV under MacOS...

src/iowrap.hpp
```diff
@@ -0,0 +1,139 @@
...
+#include <thread>
+
+class AbstractIO {
+   public:
+    AbstractIO(const std::string&) { }
+
+    virtual ~AbstractIO() { }
+
+    virtual int64_t read(void* buffer, size_t length) = 0;
+    virtual std::string ReaderName() const = 0;
```
marcelm, line 20: As far as I can see, this method is never used, so you could just remove it.

src/iowrap.hpp
```diff
@@ -0,0 +1,139 @@
...
+#define STROBEALIGN_IOWRAP_HPP
+
+#include <cstdint>
+#include <string>
+
+#include <zlib.h>
+#include <vector>
+#include "isa-l/igzip_lib.h"
+
+#include <thread>
```
marcelm, line 11:
```suggestion
#include <thread>
#include <vector>
#include <zlib.h>
#include <isa-l/igzip_lib.h>
```
(Yes, I realize I suggested to write `#include "isa-l/..."` originally, but this is probably more consistent.)

src/iowrap.cpp
```diff
@@ -0,0 +1,206 @@
+#include "iowrap.hpp"
```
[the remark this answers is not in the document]
marcelm: Thanks for reformatting `iowrap.hpp`. Can you please run `clang-format` also on `iowrap.cpp`?

src/iowrap.hpp
telmin: I think it would be useful to have a method to display the name for debugging purposes to see which Reader is being used. How do you think?
How about use [[maybe_unused]] ?

telmin:
Thank you for your review. I have fix the points you mentioned.

marcelm (commented):
I’m just testing this locally and noticed that the tests (`tests/run.sh`) now run much slower and take more memory. Before, they ran in ~60ms and needed 5 MB of memory, and now they need >3 seconds and need 1.5 GB of memory. I assume this is the 256 MiB buffer that is initialized for each input FASTQ file and for each thread.

What is the benefit of the buffering? Have you tried different buffer sizes? I wonder whether much smaller buffer of perhaps 256 KiB wouldn’t be enough.

Generally, I recommend doing one thing per PR. As it is, it is now a bit difficult to tease apart what the effects of each of the changes is.

marcelm:
Regarding the macOS failure, this is the error message:
```
Total time writing index: 0.02 s
tests/run.sh: line 56:  2953 Segmentation fault: 11  strobealign -r 150 -i tests/phix.fasta
```
The `strobealign -i` command does not even read from any input FASTQ files, and it crashes right at the end shortly before it would exit anyway. Maybe some cleanup that doesn’t work as expected?

telmin:
The reason I am double-buffering the gzip decompression is to speed up the process; trying to do gzip decompression without double-buffering would be very slow, so I am buffering to avoid this.

But you are right, I think we should properly consider if we really need GB of Buffer.
So I took some benchmarks.
In the following benchmark, decompression is done n bytes at a time and the decompressed buffer is allocated to 4*n bytes.

* 256KB / 1MB
```
Total time reading read-file(s): 211.97 s.
```

* 1MB / 4MB
```
Total time reading read-file(s): 109.14 s.
```

* 2MB / 8MB
```
Total time reading read-file(s): 82.67 s.
```

* 4MB / 16MB
```
Total time reading read-file(s): 85.07 s.
```

* 8MB / 32MB
```
Total time reading read-file(s): 88.06 s.
```

* 16MB / 64MB
```
Total time reading read-file(s): 85.09 s.
```

* 32MB / 128MB
```
Total time reading read-file(s): 90.15 s.
```

* 64MB / 256MB
```
Total time reading read-file(s): 91.16 s.
```

* 128MB / 512MB
```
Total time reading read-file(s): 93.04 s.
```

* 256MB / 1GB
```
Total time reading read-file(s): 92.55 s.
```

On my machine, with the proper buffer size, it seems to generally finish in around 80sec. 256KB unit decompression does not seem to cover up the decompression time. So I did the decompression in 2MB increments and changed the buffer size to 8MB for decompressed data.

marcelm:
Hi, I’m back from traveling, sorry for the delay. Thanks for the update, and great that you could even fix the macOS issue.

I’m trying to understand `UncompressedReader` and `IsalGzipReader`, but am a bit slow. I think a couple of comments would have helped. But can you help me for now by explaining a couple of things:
- What are `read_buffer` and `read_buffer_work` and what is the difference?
- Why does `UncompressReader` use a separate thread?
- Is it correct that a thread is created and joined for each read chunk? (I guess it’s ok because thread creation is efficient.)

I also wonder whether using `mmap()` for the input is appropriate. I’d tend to use that for files that need random access, but the input FASTQ needs to be streamed only once, so I don’t see why it would be necessary to map it in its entirety into memory. It also appears to make the code more complex as you now need to find out what the file size is in advance.

src/iowrap.hpp
```diff
@@ -0,0 +1,139 @@
...
+
+    int64_t read(void* buffer, size_t length) override;
+    [[maybe_unused]] std::string name() const override { return "GzipReader"; }
+
+   private:
+    gzFile file;
+    void open(const std::string& filename) override;
+};
+
+class UncompressReader : public Reader {
```
marcelm, line 44: `UncompressReader` sounds as if this were a reader that uncompresses. I suggest to rename to `UncompressedReader` to avoid the ambiguity.

src/iowrap.hpp
```diff
@@ -0,0 +1,139 @@
...
+   public:
+    UncompressReader(const std::string& filename)
+        : Reader(filename)
+        , fd(-1)
+        , preload_size(64ull * 1024 * 1024)
```
marcelm, line 49: I noticed that the tests still run a bit slower than before, and this is due to this buffer: If I reduce its size to 1 MB, then the tests are as fast as before.

One option would be a dynamic buffer size: Start with 1 MiB and double it in each `read()` until it has reached 64 MiB.

But on the other hand, I wonder whether this class couldn’t be much simpler if you just avoid the threading and choose a smaller buffer size. Does the threading actually help? Since reading uncompressed files is so much faster than reading gzip files, I wouldn’t expect reading the data in a thread to help that much.

marcelm:
This has been sitting here for a while, but it would still be nice to have in strobealign. I made the modifications I think are necessary to merge this and opened #418 instead.

@telmin I squashed your contribution to a single commit (preserving authorship information).

I’m closing this now in favor of #418.

Issue #216, opened 2023-01-30
Try other syncmer hash functions


marcelm:
Strobealign currently uses [xxh64](https://github.com/ksahlin/strobealign/blob/8a0e047a451f4ff0ae74988f32695de0a0c80ea4/src/randstrobes.cpp#L38) to hash syncmers.

We could switch to a different, faster hash function as long as accuracy is at least as good as now.

These are some options for hash functions. The first four are those listed on https://github.com/ksahlin/strobemers/tree/main/randstrobe_implementations:
1. No hash (2-bit encoded nucleotides)
2. Thomas Wang hash
3. [xxhash](https://github.com/Cyan4973/xxHash)
4. [wyhash](https://github.com/wangyi-fudan/wyhash)
5. [FNV-1a](https://en.wikipedia.org/wiki/Fowler_Noll_Vo_hash)
6. [fxhash](https://github.com/cbreeden/fxhash)
7. [xxh3](https://github.com/Cyan4973/xxHash)

When looking at published benchmark results, we need to keep in mind that our use case is hashing a single 64-bit value. Benchmarking results obtained on larger amounts of data may not be representative.

A first glance at the ["small data" benchmark](https://github.com/Cyan4973/xxHash#small-data) by the xxhash authors makes it appear as if FNV-1a ("fnv64") was the slowest, but it is actually quite ok when applied to 8 input bytes. The [wyhash benchmark](https://github.com/wangyi-fudan/wyhash) shows wyhash ("wyhash_final3") as being the fastest among the tested ones, but I think it doesn’t show how it would perform on a single 64 bit input number.

There is the great [smhasher](https://github.com/rurban/smhasher) hash benchmark page. They list xxh3 and wyhash in their [summary](https://github.com/rurban/smhasher#summary) among the fastest hash functions without quality problems. But they also say:
> When used in a hash table the instruction cache will usually beat the CPU and throughput measured here. In my tests the smallest FNV1A beats the fastest crc32_hw1 with [Perl 5 hash tables](https://github.com/rurban/perl-hash-stats). Even if those worse hash functions will lead to more collisions, the overall speed advantage and inline-ability beats the slightly worse quality.

That’s why I listed FNV-1a above. Its implementation is also super simple, just ~6 lines of code compared to the 5000-line xxhash.h file.

[...]

PR #278, opened 2023-05-08
Using two vectors to represent the hash table when indexing (similar running time and much lower memory usage)

psj1997:
We do not add the count in the top N bits.  We searched the count with O(log N). In my benchmarking, it can achieve similar running time. 

It means that we can get even smaller indices later by removing the top of the hash. 

Sincerely
Shaojun

### DIFF
--- a/src/arguments.hpp
+++ b/src/arguments.hpp
@@ -24,10 +24,11 @@ struct SeedingArguments {
             "It is recommended not to change this parameter unless you have a good "
             "understanding of syncmers as it will drastically change the memory usage and "
             "results with non default values.", {'s'}}
+        , n{parser, "INT", "use top n bits to store hash [28]", {'n'}}
     {
     }
     args::ArgumentParser& parser;
-    args::ValueFlag<int> r, m, k, l, u, c, s;
+    args::ValueFlag<int> r, m, k, l, u, c, s, n;
 };
 
 #endif
--- a/src/index.cpp
+++ b/src/index.cpp
@@ -46,8 +46,10 @@ unsigned int StrobemerIndex::find(uint64_t key) const{
         if (randstrobes_vector[position_start].hash == key){
             return position_start;
         }
+        else{
+            return -1;
+        }
     }
-
     // Binary research to find the hash
     int firstOccur = -1;
     while (position_start <= position_end){
--- a/src/index.cpp
+++ b/src/index.cpp
@@ -24,13 +24,13 @@ static const uint32_t STI_FILE_FORMAT_VERSION = 1;
 
 unsigned int StrobemerIndex::find(uint64_t key) const{
     unsigned int top_N = key >> (64 - N);
-    unsigned int position_start = hash_positions[top_N] - 1;
+    int position_start = hash_positions[top_N] - 1;
 
     if (position_start == -1){
         return position_start;
     }
 
-    unsigned int position_end = -1;
+    int position_end = -1;
     for (unsigned int i = top_N + 1; i < 1 << N; i++){
         if (hash_positions[i] != 0){
             position_end = hash_positions[i] - 2;
--- a/src/index.cpp
+++ b/src/index.cpp
@@ -22,8 +22,6 @@
 static Logger& logger = Logger::get();
 static const uint32_t STI_FILE_FORMAT_VERSION = 1;
 
-bool cmp(const RefRandstrobeWithHash lhs, const RefRandstrobeWithHash rhs) { return (lhs.hash & StrobemerIndex::hash_mask) < (rhs.hash & StrobemerIndex::hash_mask); }
-
 const int MAX_LINEAR_SEARCH = 4;
 unsigned int StrobemerIndex::find(uint64_t key) const{
     const unsigned int top_N = key >> (64 - N);
@@ -40,6 +38,8 @@ unsigned int StrobemerIndex::find(uint64_t key) const{
           }
           return -1;
       }
+
+    auto cmp = [this](const RefRandstrobeWithHash lhs, const RefRandstrobeWithHash rhs) {return (lhs.hash & hash_mask) < (rhs.hash & hash_mask); };
 
     auto pos = std::lower_bound(randstrobes_vector.begin() + position_start,
                                                randstrobes_vector.begin() + position_end,
@@ -77,15 +77,8 @@ void StrobemerIndex::write(const std::string& filename) const {
     write_int_to_ostream(ofs, filter_cutoff);
     parameters.write(ofs);
 
-    write_vector(ofs, flat_vector);
-
-    // write mers_index
-    auto size = uint64_t(randstrobe_map.size());
-    ofs.write(reinterpret_cast<char*>(&size), sizeof(size));
-    for (auto& p : randstrobe_map) {
-        ofs.write(reinterpret_cast<const char*>(&p.first), sizeof(p.first));
-        ofs.write(reinterpret_cast<const char*>(&p.second), sizeof(p.second));
-    }
+    write_vector(ofs, randstrobes_vector);
+    write_vector(ofs, hash_positions);
 }
 
 void StrobemerIndex::read(const std::string& filename) {
@@ -123,29 +116,8 @@ void StrobemerIndex::read(const std::string& filename) {
         throw InvalidIndexFile("Index parameters in .sti file and those specified on command line differ");
     }
 
-    read_vector(ifs, flat_vector);
-
-    uint64_t sz;
-    // read mers_index:
-    randstrobe_map.clear();
-    ifs.read(reinterpret_cast<char*>(&sz), sizeof(sz));
-    randstrobe_map.reserve(sz);
-    // read in big chunks
-    const uint64_t chunk_size = pow(2,20);//4 M => chunks of ~10 MB - The chunk size seem not to be that important
-    auto buf_size = std::min(sz, chunk_size) * (sizeof(RandstrobeMap::key_type) + sizeof(RandstrobeMap::mapped_type));
-    std::unique_ptr<char> buf_ptr(new char[buf_size]);
-    char* buf2 = buf_ptr.get();
-    auto left_to_read = sz;
-    while (left_to_read > 0) {
-        auto to_read = std::min(left_to_read, chunk_size);
-        ifs.read(buf2, to_read * (sizeof(RandstrobeMap::key_type) + sizeof(RandstrobeMap::mapped_type)));
-        //Add the elements directly from the buffer
-        for (size_t i = 0; i < to_read; ++i) {
-            auto start = buf2 + i * (sizeof(RandstrobeMap::key_type) + sizeof(RandstrobeMap::mapped_type));
-            randstrobe_map[*reinterpret_cast<RandstrobeMap::key_type*>(start)] = *reinterpret_cast<RandstrobeMap::mapped_type*>(start + sizeof(RandstrobeMap::key_type));
-        }
-        left_to_read -= to_read;
-    }
+    read_vector(ifs, randstrobes_vector);
+    read_vector(ifs, hash_positions);
 }
 
 int estimate_randstrobe_hashes(const std::string& seq, const IndexParameters& parameters) {
@@ -200,7 +172,6 @@ void StrobemerIndex::populate(float f, size_t n_threads) {
     auto randstrobe_hashes = estimate_randstrobe_hashes_parallel(references, parameters, n_threads);
     stats.elapsed_unique_hashes = estimate_unique.duration();
     logger.debug() << "Estimated number of randstrobe hashes: " << randstrobe_hashes << '\n';
-    // randstrobe_map.reserve(randstrobe_hashes);
 
     Timer randstrobes_timer;
     add_randstrobes_to_vector(randstrobe_hashes);
@@ -212,7 +183,6 @@ void StrobemerIndex::populate(float f, size_t n_threads) {
     stats.elapsed_sorting_seeds = sorting_timer.duration();
 
     Timer hash_index_timer;
-    // stats.flat_vector_size = randstrobes_vector.size();
 
     unsigned int offset = 0;
     unsigned int tot_high_ab = 0;
@@ -222,11 +192,13 @@ void StrobemerIndex::populate(float f, size_t n_threads) {
     unsigned int randstrobe_hash_size = 0;      
     uint64_t prev_hash = -1;
     
-    // stats.flat_vector_size = 0;
     stats.tot_occur_once = 0;
     // calculate stats.flat_vector_size, randstrobe_hash_size
     // add the top N bits of hash to the hash_positions
     // calculate the count of hash that exists more than one time
+
+    hash_positions.reserve(1 << N);
+
     unsigned int count = 0;
     unsigned int position = 0;
     unsigned int prev_hash_N = 0;
@@ -237,9 +209,8 @@ void StrobemerIndex::populate(float f, size_t n_threads) {
             randstrobe_hash_size += 1;
             occur_once = true;
             unsigned int existing_hash_N = hash_value >> (64 - N);
-            hash_positions[existing_hash_N] = position;
-            for (unsigned int index_temp = 0; index_temp < existing_hash_N; index_temp++){
-                hash_positions[index_temp] = position;
+            for (unsigned int index_temp = 0; index_temp <= existing_hash_N; index_temp++){
+                hash_positions.push_back(position);
             }
             prev_hash_N = existing_hash_N;
             position += 1;
@@ -274,9 +245,8 @@ void StrobemerIndex::populate(float f, size_t n_threads) {
             count = 1;
             unsigned int existing_hash_N = hash_value >> (64 - N);
             if (existing_hash_N != prev_hash_N){
-                hash_positions[existing_hash_N] = position;
-                for (unsigned int index_temp = prev_hash_N + 1; index_temp < existing_hash_N; index_temp++){
-                    hash_positions[index_temp] = position;
+                for (unsigned int index_temp = prev_hash_N + 1; index_temp <= existing_hash_N; index_temp++){
+                    hash_positions.push_back(position);
             }
             prev_hash_N = existing_hash_N;
             }
@@ -298,7 +268,7 @@ void StrobemerIndex::populate(float f, size_t n_threads) {
             }
             randstrobes_vector[position - (count - 1)].hash = (prev_hash & hash_mask) | ((uint64_t)count << (64 - N));
             for (unsigned int index_temp = prev_hash_N + 1; index_temp < (1 << N); index_temp++){
-                    hash_positions[index_temp] = randstrobes_vector.size();
+                    hash_positions.push_back(randstrobes_vector.size());
             }
         }
     }
@@ -358,7 +328,6 @@ void StrobemerIndex::add_randstrobes_to_vector(int randstrobe_hashes){
                 RefRandstrobeWithHash::packed_t packed = ref_index << 8;
                 packed = packed + (randstrobe.strobe2_pos - randstrobe.strobe1_pos);
                 // try to insert as direct entry
-                // RandstrobeMapEntry entry{randstrobe.strobe1_pos, packed | 0x8000'0000};
                 randstrobes_vector.push_back(RefRandstrobeWithHash{randstrobe.hash, randstrobe.strobe1_pos, packed});
                 }
             chunk.clear();
@@ -366,78 +335,6 @@ void StrobemerIndex::add_randstrobes_to_vector(int randstrobe_hashes){
         }
     // stats.tot_occur_once = tot_occur_once;
 }
-
-/*
- * Generate randstrobes for all reference sequences and add them to the hash
- * table. Only those randstrobes which occur only once have correct, fully
- * filled-in entries in the hash table. For the others (with multiple
- * occurrences), only their count is correct. The offset needs to be filled in
- * later.
- *
- * Fills in
- * - stats.tot_occur_once
- * - stats.tot_strobemer_count
- */
-//  std::vector<RefRandstrobeWithHash> StrobemerIndex::add_randstrobes_to_hash_table() {
-//     std::vector<RefRandstrobeWithHash> randstrobes_with_hash;
-//     size_t tot_occur_once = 0;
-//     for (size_t ref_index = 0; ref_index < references.size(); ++ref_index) {
-//         auto seq = references.sequences[ref_index];
-//         if (seq.length() < parameters.w_max) {
-//             continue;
-//         }
-//         auto randstrobe_iter = RandstrobeIterator2(seq, parameters.k, parameters.s, parameters.t_syncmer, parameters.w_min, parameters.w_max, parameters.q, parameters.max_dist);
-//         std::vector<Randstrobe> chunk;
-//         // TODO
-//         // Chunking makes this function faster, but the speedup is achieved even
-//         // with a chunk size of 1.
-//         const size_t chunk_size = 4;
-//         chunk.reserve(chunk_size);
-
-//         bool end = false;
-//         while (!end) {
-//             // fill chunk
-//             Randstrobe randstrobe;
-//             while (chunk.size() < chunk_size) {
-//                 randstrobe = randstrobe_iter.next();
-//                 if (randstrobe == randstrobe_iter.end()) {
-//                     end = true;
-//                     break;
-//                 }
-//                 chunk.push_back(randstrobe);
-//             }
-//             stats.tot_strobemer_count += chunk.size();
-//             for (auto randstrobe : chunk) {
-//                 RefRandstrobeWithHash::packed_t packed = ref_index << 8;
-//                 packed = packed + (randstrobe.strobe2_pos - randstrobe.strobe1_pos);
-
-//                 // try to insert as direct entry
-//                 RandstrobeMapEntry entry{randstrobe.strobe1_pos, packed | 0x8000'0000};
-//                 auto result = randstrobe_map.insert({randstrobe.hash, entry});
-//                 if (result.second) {
-//                     tot_occur_once++;
-//                 } else {
-//                     // already exists in hash table
-//                     auto existing = result.first;
-//                     auto existing_count = existing->second.count();
-//                     if (existing_count == 1) {
-//                         // current entry is a direct one, convert to an indirect one
-//                         auto existing_randstrobe = existing->second.as_ref_randstrobe();
-//                         randstrobes_with_hash.push_back(RefRandstrobeWithHash{randstrobe.hash, existing_randstrobe.position, existing_randstrobe.packed()});
-//                         tot_occur_once--;
-//                     }
-//                     // offset is adjusted later after sorting
-//                     existing->second.set_count(existing_count + 1);
-
-//                     randstrobes_with_hash.push_back(RefRandstrobeWithHash{randstrobe.hash, randstrobe.strobe1_pos, packed});
-//                 }
-//             }
-//             chunk.clear();
-//         }
-//     }
-//     stats.tot_occur_once = tot_occur_once;
-//     return randstrobes_with_hash;
-// }
 
 void StrobemerIndex::print_diagnostics(const std::string& logfile_name, int k) const {
     // Prins to csv file the statistics on the number of seeds of a particular length and what fraction of them them are unique in the index:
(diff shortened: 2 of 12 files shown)


luispedro:
The `log N` method to get the count is implemented here:

https://github.com/psj1997/strobealign/blob/2a024094524625855cc3e26452fb204f2a422109/src/index.cpp#L344

As with the `find` method, we first try a small linear search (https://github.com/psj1997/strobealign/blob/2a024094524625855cc3e26452fb204f2a422109/src/index.hpp#L134) and if that fails, we switch to the `log N` method

ksahlin:
Thanks for the great work behind this PR. I was going to benchmark this version right now but I run into build error, see below.

```
(base) [kris@rackham4 strobealign]$ module add cmake gcc
(base) [kris@rackham4 strobealign]$ cmake -B build -DCMAKE_C_FLAGS="-march=native" -DCMAKE_CXX_FLAGS="-march=native"
-- The C compiler identification is GNU 13.1.0
-- The CXX compiler identification is GNU 13.1.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /sw/comp/gcc/13.1.0_rackham/bin/gcc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /sw/comp/gcc/13.1.0_rackham/bin/g++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found ZLIB: /home/kris/bin/lib/libz.so (found version "1.2.11") 
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - found
-- Found Threads: TRUE  
-- Found OpenMP_C: -fopenmp (found version "4.5") 
-- Found OpenMP_CXX: -fopenmp (found version "4.5") 
-- Found OpenMP: TRUE (found version "4.5")  
-- Build type: Release
-- Found ZLIB: /home/kris/bin/lib/libz.so (found suitable version "1.2.11", minimum required is "1.2.3") 
-- zstr - found ZLIB (version: 1.2.11)
-- zstr - added INTERFACE target 'zstr::zstr'
          includes : /home/kris/source/strobealign_lowmem/strobealign/build/_deps/zstrgitrepo-src/src;/home/kris/bin/include
          libraries: ZLIB::ZLIB
          features : cxx_std_11
-- Configuring done
-- Generating done
-- Build files have been written to: /home/kris/source/strobealign_lowmem/strobealign/build
(base) [kris@rackham4 strobealign]$ make -j -C build
make: Entering directory `/domus/h1/kris/source/strobealign_lowmem/strobealign/build'
make[1]: Entering directory `/domus/h1/kris/source/strobealign_lowmem/strobealign/build'
make[2]: Entering directory `/domus/h1/kris/source/strobealign_lowmem/strobealign/build'
make[2]: Leaving directory `/domus/h1/kris/source/strobealign_lowmem/strobealign/build'
make[2]: Entering directory `/domus/h1/kris/source/strobealign_lowmem/strobealign/build'
Setting version to 0.9.0
make[2]: Leaving directory `/domus/h1/kris/source/strobealign_lowmem/strobealign/build'
[  0%] Built target version
make[2]: Entering directory `/domus/h1/kris/source/strobealign_lowmem/strobealign/build'
make[2]: Leaving directory `/domus/h1/kris/source/strobealign_lowmem/strobealign/build'
make[2]: Entering directory `/domus/h1/kris/source/strobealign_lowmem/strobealign/build'
[  6%] Building CXX object CMakeFiles/salib.dir/src/refs.cpp.o
[  6%] Building CXX object CMakeFiles/salib.dir/src/cmdline.cpp.o
[ 12%] Building CXX object CMakeFiles/salib.dir/src/cigar.cpp.o
[ 15%] Building CXX object CMakeFiles/salib.dir/src/paf.cpp.o
[ 21%] Building CXX object CMakeFiles/salib.dir/src/index.cpp.o
[ 27%] Building CXX object CMakeFiles/salib.dir/src/fastq.cpp.o
[ 27%] Building CXX object CMakeFiles/salib.dir/src/indexparameters.cpp.o
[ 39%] Building CXX object CMakeFiles/salib.dir/src/sam.cpp.o
[ 45%] Building CXX object CMakeFiles/salib.dir/src/randstrobes.cpp.o
[ 45%] Building CXX object CMakeFiles/salib.dir/src/readlen.cpp.o
[ 45%] Building CXX object CMakeFiles/salib.dir/src/version.cpp.o
[ 48%] Building CXX object CMakeFiles/salib.dir/src/nam.cpp.o
[ 54%] Building CXX object CMakeFiles/salib.dir/src/aligner.cpp.o
[ 57%] Building CXX object CMakeFiles/salib.dir/src/pc.cpp.o
[ 57%] Building CXX object CMakeFiles/salib.dir/ext/ssw/ssw_cpp.cpp.o
[ 57%] Building C object CMakeFiles/salib.dir/ext/xxhash.c.o
[ 57%] Building CXX object CMakeFiles/salib.dir/src/io.cpp.o
[ 57%] Building CXX object CMakeFiles/salib.dir/src/aln.cpp.o
[ 57%] Building C object CMakeFiles/salib.dir/ext/ssw/ssw.c.o
In file included from /home/kris/source/strobealign_lowmem/strobealign/src/io.cpp:1:
/home/kris/source/strobealign_lowmem/strobealign/src/io.hpp: In function ‘void read_vector(std::istream&, std::vector<T>&)’:
/home/kris/source/strobealign_lowmem/strobealign/src/io.hpp:21:5: error: ‘uint64_t’ was not declared in this scope
   21 |     uint64_t size;
      |     ^~~~~~~~
/home/kris/source/strobealign_lowmem/strobealign/src/io.hpp:6:1: note: ‘uint64_t’ is defined in header ‘<cstdint>’; did you forget to ‘#include <cstdint>’?
    5 | #include <vector>
  +++ |+#include <cstdint>
    6 | 
/home/kris/source/strobealign_lowmem/strobealign/src/io.hpp:23:38: error: ‘size’ was not declared in this scope; did you mean ‘std::size’?
   23 |     is.read(reinterpret_cast<char*>(&size), sizeof(size));
      |                                      ^~~~
      |                                      std::size
In file included from /sw/comp/gcc/13.1.0_rackham/include/c++/13.1.0/string:53,
                 from /sw/comp/gcc/13.1.0_rackham/include/c++/13.1.0/bits/locale_classes.h:40,
                 from /sw/comp/gcc/13.1.0_rackham/include/c++/13.1.0/bits/ios_base.h:41,
                 from /sw/comp/gcc/13.1.0_rackham/include/c++/13.1.0/ios:44,
                 from /sw/comp/gcc/13.1.0_rackham/include/c++/13.1.0/ostream:40,
                 from /sw/comp/gcc/13.1.0_rackham/include/c++/13.1.0/iostream:41,
                 from /home/kris/source/strobealign_lowmem/strobealign/src/io.hpp:4:
/sw/comp/gcc/13.1.0_rackham/include/c++/13.1.0/bits/range_access.h:274:5: note: ‘std::size’ declared here
  274 |     size(const _Tp (&)[_Nm]) noexcept
      |     ^~~~
/home/kris/source/strobealign_lowmem/strobealign/ext/ssw/ssw_cpp.cpp: In member function ‘bool StripedSmithWaterman::Aligner::ReBuild(const int8_t*, const int&, const int8_t*, const int&)’:
/home/kris/source/strobealign_lowmem/strobealign/ext/ssw/ssw_cpp.cpp:553:19: warning: unused parameter ‘score_matrix_size’ [-Wunused-parameter]
  553 |     const int&    score_matrix_size,
      |     ~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~
make[2]: *** [CMakeFiles/salib.dir/src/io.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
/home/kris/source/strobealign_lowmem/strobealign/src/nam.cpp: In function ‘std::vector<Nam> {anonymous}::merge_hits_into_nams(robin_hood::unordered_map<unsigned int, std::vector<Hit> >&, int, bool)’:
/home/kris/source/strobealign_lowmem/strobealign/src/nam.cpp:105:27: warning: comparison of integer expressions of different signedness: ‘int’ and ‘unsigned int’ [-Wsign-compare]
  105 |             if (h.query_s > prev_q_start + k) {
      |                 ~~~~~~~~~~^~~~~~~~~~~~~~~~~~
/home/kris/source/strobealign_lowmem/strobealign/src/nam.cpp: In function ‘std::vector<Nam> find_nams_rescue(const QueryRandstrobeVector&, const StrobemerIndex&, unsigned int)’:
/home/kris/source/strobealign_lowmem/strobealign/src/nam.cpp:209:22: warning: comparison of integer expressions of different signedness: ‘unsigned int’ and ‘int’ [-Wsign-compare]
  209 |         if (position != -1){
      |             ~~~~~~~~~^~~~~
/home/kris/source/strobealign_lowmem/strobealign/src/aln.cpp: In function ‘alignment align_segment(const Aligner&, const std::string&, const std::string&, int, int, int, bool, bool)’:
/home/kris/source/strobealign_lowmem/strobealign/src/aln.cpp:177:26: warning: comparison of integer expressions of different signedness: ‘int’ and ‘long unsigned int’ [-Wsign-compare]
  177 |     if (ref_segm_len_ham == read_segm_len && !aln_did_not_fit) {
      |         ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~
/home/kris/source/strobealign_lowmem/strobealign/src/aln.cpp: In function ‘alignment get_alignment(const Aligner&, const Nam&, const References&, const Read&, bool)’:
/home/kris/source/strobealign_lowmem/strobealign/src/aln.cpp:243:49: warning: comparison of integer expressions of different signedness: ‘int’ and ‘std::__cxx11::basic_string<char>::size_type’ {aka ‘long unsigned int’} [-Wsign-compare]
  243 |     if (projected_ref_end - projected_ref_start == query.size() && fits) {
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~
/home/kris/source/strobealign_lowmem/strobealign/src/aln.cpp: In function ‘alignment get_alignment_unused(const Aligner&, const Nam&, const References&, const Read&, int, bool)’:
/home/kris/source/strobealign_lowmem/strobealign/src/aln.cpp:309:23: warning: comparison of integer expressions of different signedness: ‘int’ and ‘const long unsigned int’ [-Wsign-compare]
  309 |     if (ref_segm_size == read_len && fits) {
      |         ~~~~~~~~~~~~~~^~~~~~~~~~~
/home/kris/source/strobealign_lowmem/strobealign/src/aln.cpp:335:62: warning: comparison of integer expressions of different signedness: ‘int’ and ‘long unsigned int’ [-Wsign-compare]
  335 |     int break_point = (global_max_bp >= T) && (global_max_bp <= (read_len - T)) ? global_max_bp : -1;
      |                                                ~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~
/home/kris/source/strobealign_lowmem/strobealign/src/aln.cpp:369:86: warning: comparison of integer expressions of different signedness: ‘long unsigned int’ and ‘const int’ [-Wsign-compare]
  369 |         ref_tmp_segm_size = right_ref_start_bp + (read_len + diff - right_region_bp) < ref_len ? (read_len + diff - right_region_bp) : ref_len - right_ref_start_bp;
      |                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~
/home/kris/source/strobealign_lowmem/strobealign/src/index.cpp: In member function ‘void StrobemerIndex::read(const std::string&)’:
/home/kris/source/strobealign_lowmem/strobealign/src/index.cpp:79:31: warning: comparison of integer expressions of different signedness: ‘std::vector<unsigned int>::size_type’ {aka ‘long unsigned int’} and ‘int’ [-Wsign-compare]
   79 |     if (hash_positions.size() != 1 << parameters.n) {
      |         ~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~
/home/kris/source/strobealign_lowmem/strobealign/src/index.cpp: In member function ‘void StrobemerIndex::populate(float, size_t)’:
/home/kris/source/strobealign_lowmem/strobealign/src/index.cpp:199:34: warning: comparison of integer expressions of different signedness: ‘std::vector<unsigned int>::size_type’ {aka ‘long unsigned int’} and ‘int’ [-Wsign-compare]
  199 |     while (hash_positions.size() < (1 << parameters.n)) {
      |            ~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
/home/kris/source/strobealign_lowmem/strobealign/src/index.cpp: In member function ‘void StrobemerIndex::print_diagnostics(const std::string&, int) const’:
/home/kris/source/strobealign_lowmem/strobealign/src/index.cpp:283:25: warning: comparison of integer expressions of different signedness: ‘int’ and ‘std::vector<RefRandstrobeWithHash>::size_type’ {aka ‘long unsigned int’} [-Wsign-compare]
  283 |     for (int it = 0; it < randstrobes_vector.size(); it++) {
      |                      ~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /home/kris/source/strobealign_lowmem/strobealign/ext/hyperloglog/hyperloglog.hpp:16,
                 from /home/kris/source/strobealign_lowmem/strobealign/src/index.cpp:17:
/home/kris/source/strobealign_lowmem/strobealign/ext/hyperloglog/murmur3.h: In function ‘void MurmurHash3_x86_32(const void*, int, uint32_t, void*)’:
/home/kris/source/strobealign_lowmem/strobealign/ext/hyperloglog/murmur3.h:134:16: warning: this statement may fall through [-Wimplicit-fallthrough=]
  134 |     case 3: k1 ^= tail[2] << 16;
      |             ~~~^~~~~~~~~~~~~~~~
/home/kris/source/strobealign_lowmem/strobealign/ext/hyperloglog/murmur3.h:135:5: note: here
  135 |     case 2: k1 ^= tail[1] << 8;
      |     ^~~~
/home/kris/source/strobealign_lowmem/strobealign/ext/hyperloglog/murmur3.h:135:16: warning: this statement may fall through [-Wimplicit-fallthrough=]
  135 |     case 2: k1 ^= tail[1] << 8;
      |             ~~~^~~~~~~~~~~~~~~
/home/kris/source/strobealign_lowmem/strobealign/ext/hyperloglog/murmur3.h:136:5: note: here
  136 |     case 1: k1 ^= tail[0];
      |     ^~~~
make[2]: Leaving directory `/domus/h1/kris/source/strobealign_lowmem/strobealign/build'
make[1]: *** [CMakeFiles/salib.dir/all] Error 2
make[1]: Leaving directory `/domus/h1/kris/source/strobealign_lowmem/strobealign/build'
make: *** [all] Error 2
make: Leaving directory `/domus/h1/kris/source/strobealign_lowmem/strobealign/build'
(base) [kris@rackham4 strobealign]$ 

```

src/index.hpp
```diff
@@ -151,36 +79,92 @@ struct StrobemerIndex {
...
-        randstrobe_map[key] = RandstrobeMapEntry{offset, count};
+    int reference_index(unsigned int position) const {
+        return randstrobes_vector[position].packed >> bit_alloc;
+    }
+
+    unsigned int get_count(const unsigned int position) const {
+        constexpr unsigned int MAX_LINEAR_SEARCH = 8;
+        const auto hash = randstrobes_vector[position].hash;
+
+        // For 95% of cases, the result will be small and a brute force search
```
marcelm, line 138: Unless this number was obtained by measuring, I’d suggest to just say something noncommital such as "In most cases" or so.

src/index.hpp
```diff
@@ -151,36 +79,92 @@ struct StrobemerIndex {
...
 private:
-    std::vector<RefRandstrobeWithHash> add_randstrobes_to_hash_table();
+    void add_randstrobes_to_vector(int randstrobe_hashes);
+    unsigned int get_count_line_search(unsigned int) const;
 
     const IndexParameters& parameters;
     const References& references;
-    RandstrobeMap randstrobe_map; // k-mer -> (offset in flat_vector, occurence count )
+    std::vector<RefRandstrobeWithHash> randstrobes_vector;
+    std::vector<unsigned int> hash_positions;
```
marcelm, line 165: When reading this and other code using the "position" terminology, I constantly had to remind myself that "position" does not refer to a location on the reference sequence. So at least to me, it doesn’t give the right connotation and I’d prefer a different term, probably something involving "index". It’s also not the position of a hash, but the position of a RefRandstrobeWithHash.

Maybe something like `randstrobe_start_indices`?

src/nam.cpp
```diff
@@ -152,20 +154,23 @@ std::pair<float, std::vector<Nam>> find_nams(
     robin_hood::unordered_map<unsigned int, std::vector<Hit>> hits_per_ref;
     hits_per_ref.reserve(100);
 
+    /*
+    1. Find the hash in the vector
+    2. the occur times of the hash value, use a flag 
+    3. need to know reference index, strobe1 position, storbe2 - strobe1
```
marcelm, line 160: The grammar is a bit off. What do these comments mean?

src/nam.cpp
```diff
@@ -152,20 +154,23 @@ std::pair<float, std::vector<Nam>> find_nams(
...
+    */
+
     int nr_good_hits = 0, total_hits = 0;
     for (const auto &q : query_randstrobes) {
-        auto ref_hit = index.find(q.hash);
-        if (ref_hit != index.end()) {
+        int position = index.find(q.hash);
+        if (position != -1){
```
marcelm, line 166: I think it would be nice if this could work unchanged, that is, have `index.end()` return -1.

src/nam.cpp
```diff
@@ -152,20 +154,23 @@ std::pair<float, std::vector<Nam>> find_nams(
...
             total_hits++;
-            if (ref_hit->second.count() > index.filter_cutoff) {
-                continue;
-            }
+            if (index.get_hash(position + index.filter_cutoff) == q.hash) { continue; }
```
marcelm, line 168: This line can stop working when the memory layout changes. I’d suggest to factor this out into a StrobemerIndex method.

src/index.cpp
```diff
@@ -279,33 +251,13 @@ std::vector<RefRandstrobeWithHash> StrobemerIndex::add_randstrobes_to_hash_table
             for (auto randstrobe : chunk) {
                 RefRandstrobeWithHash::packed_t packed = ref_index << 8;
                 packed = packed + (randstrobe.strobe2_pos - randstrobe.strobe1_pos);
-
                 // try to insert as direct entry
```
marcelm, line 254: This comment is now outdated

src/index.hpp
```diff
@@ -151,36 +79,92 @@ struct StrobemerIndex {
...
+        //
+        unsigned int count = 1;
+        while (get_hash(position + count) == hash
+                && count < MAX_LINEAR_SEARCH) {
+            ++count;
+        }
+        if (get_hash(position + count) != hash) {
+            return count;
+        }
+        return count + get_count_line_search(position + count);
```
marcelm, line 151: Could this just use `std::upper_bound` instead?

marcelm (commented):
Hi, thanks a lot for this! I’ve now done a first round of review and mostly have code-style comments (but a lot of those). There’ll probably be another round, so I’ll leave it at that for the moment.

src/arguments.hpp
```diff
@@ -24,10 +24,11 @@ struct SeedingArguments {
             "It is recommended not to change this parameter unless you have a good "
             "understanding of syncmers as it will drastically change the memory usage and "
             "results with non default values.", {'s'}}
+        , n{parser, "INT", "use top n bits to store hash [28]", {'n'}}
```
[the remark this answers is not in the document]
ksahlin: I agree, `-b` is an intuitive and good suggestion.

src/index.hpp
luispedro: You could, it's a very similar idea, but you'd have to find the top-most end point anyway to call it (or you use `end()`, which is a worst case scenario, which is not good either and ruins cache locality).

src/index.hpp
luispedro: It's an informed guess. In the human genome, if I assume that the frequency a hash will be queried is proportional to the frequency it appears in the table, with `MAX_LINEAR_SEARCH=8`, the actual value is `96%`.

src/index.cpp
```diff
@@ -98,59 +74,44 @@ void StrobemerIndex::read(const std::string& filename) {
...
+        throw InvalidIndexFile("hash_positions vector is of the wrong size");
     }
 }
 
-hll::HyperLogLog estimate_unique_randstrobe_hashes(const std::string& seq, const IndexParameters& parameters) {
-    hll::HyperLogLog hll(10);
+int estimate_randstrobe_hashes(const std::string& seq, const IndexParameters& parameters) {
+    int num = 0;
 
     auto randstrobe_iter = RandstrobeIterator2(seq, parameters.k, parameters.s, parameters.t_syncmer, parameters.w_min, parameters.w_max, parameters.q, parameters.max_dist);
```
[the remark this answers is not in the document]
ksahlin: Another suggestion/PR: This count could be used to set `-b` dynamically if it is not specified.

src/index.hpp
marcelm: I meant doing it similar to how `StrobemerIndex::find` determines the interval to search:
```c++
        int position_start = hash_positions[top_N];
        int position_end = hash_positions[top_N + 1];
```
Here, one could use just `position` as the start index, but shouldn’t then `std::upper_bound` work?

src/index.hpp
marcelm: Maybe we could introduce some terminology to make this a little bit easier to grasp. When thinking about how this method works, I’ve been using the term "bucket" for an interval in `randstrobes_vector` that contains entries whose top n hash bits are identical. If we use that term, this variable could be named `bucket_starts`.

marcelm (commented):
Sorry for the amount of review comments. Many should be easy to answer or fix. I’m also sorry if it appears that I’m holding your code to a higher standard than the one we already have. It’s totally fine if you decide not to address some of these comments (especially the ones where I ask you to do a refactoring). Just say so, and then I can do those "touchups" separately after the PR is merged.

src/index.hpp
luispedro: Ah, yes, sorry, I hadn't thought of that: one can use the hash itself to look up at the `hash_positions` vector and get the full interval.

With the current version of index, that'd indeed work (at the minor cost of a likely cache miss when looking at `hash_positions`) This would no longer be possible if we implement the idea of removing redundant hash values, though, while the current code would remain valid

src/nam.cpp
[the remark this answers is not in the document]
luispedro: Yes, but `get_hash` is robust to that

src/nam.cpp
psj1997: I write these notes for coding. I will delete it.

src/arguments.hpp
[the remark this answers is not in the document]
psj1997: OK. I agree.

psj1997:
I have updated the code based on @marcelm's comments.

src/unused.cpp
```diff
@@ -13,6 +13,86 @@
 int main() {
 }
 
+namespace {
```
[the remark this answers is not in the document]
marcelm: What I meant here is: Just leave `unused.cpp` as it is.

marcelm (commented):
Hi, thanks for the updates. A couple of comments haven’t been addressed and I get this compilation error at the moment:
```
/home/marcel/scm/strobealign/src/nam.cpp: At global scope:
.../strobealign/src/nam.cpp:138:1: error: expected declaration before ‘}’ token
  138 | } // namespace
      | ^
```

src/index.cpp
```diff
@@ -98,59 +74,44 @@ void StrobemerIndex::read(const std::string& filename) {
...
+    read_vector(ifs, randstrobes_vector);
+    read_vector(ifs, hash_positions);
+    if (hash_positions.size() != 1 << parameters.n) {
+        throw InvalidIndexFile("hash_positions vector is of the wrong size");
     }
 }
 
-hll::HyperLogLog estimate_unique_randstrobe_hashes(const std::string& seq, const IndexParameters& parameters) {
-    hll::HyperLogLog hll(10);
+int estimate_randstrobe_hashes(const std::string& seq, const IndexParameters& parameters) {
```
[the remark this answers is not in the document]
psj1997: I will open another PR for this.

src/index.cpp
```diff
@@ -327,27 +279,23 @@ void StrobemerIndex::print_diagnostics(const std::string& logfile_name, int k) c
...
+            log_count_squared[seed_length] += count;
+            tot_seed_count ++;
+            tot_seed_count_sq += count;
+            if (count <= 1000){
+                log_count_1000_limit[seed_length] ++;
+                tot_seed_count_1000_limit ++;
             }
+        } else {
+            // TODO This function should not log anything
+            logger.info() << "Detected seed size over " << max_size << " bp (can happen, e.g., over centromere): " << seed_length << std::endl;
```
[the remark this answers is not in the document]
psj1997: Sorry, I do not understand.

src/index.cpp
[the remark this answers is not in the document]
marcelm: Doing the removal in a separate PR is fine, but since this PR should be self-contained, please still rename the function in this PR. (To `count_randstrobe_hashes`, for example.)

src/index.cpp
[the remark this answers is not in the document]
marcelm: Originally, the `logger.info()` line is commented out:

https://github.com/ksahlin/strobealign/blob/6909225fdadd9092c1a79195a42a91568458f084/src/index.cpp#L348-L350

src/nam.cpp
```diff
@@ -16,18 +18,18 @@ void add_to_hits_per_ref(
...
     bool is_rc,
     const StrobemerIndex& index,
-    RandstrobeMapEntry randstrobe_map_entry,
+    unsigned int position,
     int min_diff
 ) {
-    randstrobe_map_entry.for_each(index.flat_vector, [&](const RefRandstrobe& rr) {
-        int ref_s = rr.position;
-        int ref_e = rr.position + rr.strobe2_offset() + index.k();
+    for (const auto hash = index.get_hash(position); index.get_hash(position) == hash; ++position) {
```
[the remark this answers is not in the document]
psj1997: For this, I am not sure how to deal with it.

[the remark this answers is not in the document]
marcelm: That’s ok, just leave it as it is. I will then have a look at refactoring this after the PR is merged.

src/index.cpp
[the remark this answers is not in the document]
psj1997: Agree that open another PR for this.

psj1997:
I have checked and updated the code again. Can you have a look for this?

ksahlin:
To @marcelm : as discussed over emails, I think function `count_randstrobe_hashes` can be removed entirely and we can use `S = (size(reference) / (param.k-param.s+1)) + C`  where `k-s+1` is the expected sampled seed density and `C` is a constant for  ‘a little extra’ to account for variation around the mean. I think `C` could be set to `2*size(reference)^(1/2)`or something. 

This can be done in a separate PR after merge.

marcelm:
Thanks a lot for the changes! This is looking quite good now. I noticed a couple of things while experimenting with your branch and have taken the liberty of pushing some fixes to your fork.

I noticed was that the tests took a very long time to complete. This was because the default `-b 28` meant that even for the tiny phiX test genome, a 1 GiB vector was allocated and filled, and the on-disk index was also sized accordingly. And also for the fruit fly, which we often test on, memory usage both on disk and RAM was larger than before.

To fix this, I made it so that the `-b` parameter (no. of bits) is chosen automatically depending on the genome size using the formula $b=\text{floor}(\log_2(\text{no. of strobemers})) - 1$. This chooses the bucket size so that there are about two strobemers in each bucket and results in $b=28$ for the human genome, just as before. The no. of strobemers is computed using Kristoffer’s estimation formula from above, but without the constant $C$ for the moment because it overestimates the actual number of strobemers in all cases that I checked. But this is something that we should revisit (and because $b$ is only needed after all strobemers have been generated, we don’t have to use an estimate at that point and could use the actual number).

This fixed the memory and speed issue for small genomes, but also brought to light an out-of-bounds access to the `randstrobe_start_indices` vector, which I fixed in 0171adf.

Additionally, I fixed compilation on macOS and adjusted the Python bindings so that they would compile again.

I still need to do some measurements to see whether accuracy, memory usage and runtime are as expected. Also, I have more changes that I would like to make, but I will open an issue with a to-do list so that this can be done separately.

I’ll get back with the measurements, but otherwise, this PR is ready to be merged in my opinion! Thanks a lot for your work, this is a really nice improvement.

ksahlin:
Agree with @marcelm , this PR is a great improvement to strobealign.

> The no. of strobemers is computed using Kristoffer’s estimation formula from above, but without the constant 
 for the moment because it overestimates the actual number of strobemers in all cases that I checked. But this is something that we should revisit (and because is only needed after all strobemers have been generated, we don’t have to use an estimate at that point and could use the actual number).

Just chiming in to say that this is because we don't hash the s-mer within the k-mer making the selection non-random and more sparse (and likely suboptimal, see https://github.com/ksahlin/strobealign/issues/216#issuecomment-1557403783 and https://github.com/ksahlin/strobealign/issues/216#issuecomment-1558166175. It won't affect things here as I agree that we don’t have to use an estimate at that point.

marcelm:
I’ve measured accuracy between `main` and this branch on a couple of datasets and the worst I could find was a drop in 0.01 percentage points, so this looks good.

Here is a table that shows how maximum memory usage on our test genomes (read length 100) changed.

genome | before (GiB) | after (GiB) |
-|-|-
fruit fly | 0.848 | 0.684 |
CHM13 | 22.46 | 12.62 |
maize | 12.27 | 8.84 |
rye | 44.92 | 29.80 |

This means that I’ll be able to run the rye test datasets on my laptop :smile:.

As I said above, there’s lots of follow-up work, but this PR is ready to be merged IMO. The baseline commit needs to be updated, but that can be done only after the merged, so it’s fine if that test does not pass at the moment. @ksahlin What do you think?

luispedro:
Thanks. I like the auto-estimation for `b` as well. It makes perfect sense and, if my quick intuitive reasoning is correct, it means that the overhead of the indexing array becomes ~4 Bytes per entry, which is very reasonable.

ksahlin:
> the worst I could find was a drop in 0.01 percentage points, so this looks good.

I don't see how this commit should result in any positive or negative difference at all, but sure, 0.01% is not worrying. I am happy to merge this.

marcelm:
> I don't see how this commit should result in any positive or negative difference at all, but sure, 0.01% is not worrying. I am happy to merge this.

I haven’t thought this through, but I was assuming that the order in which the hits are found can be different. Anyway, I’ll merge this now.

ksahlin:
My understanding is that the order of the sorted flat_vector should remain the same. Only the index (bucket) vector is new, and it should point to the same intervals in the flat_vector as the old hash table did.

luispedro:
IIRC (from a conversation with @psj1997), the order of the sorted `flat_vector` is not strictly defined because it depends only on the hash (see the definition of [operator <](https://github.com/ksahlin/strobealign/blob/7985847aacd9e75b3ec3f8298c767dd2a4a47350/src/randstrobes.hpp#L26)).

Because a non-stable sort is used and now we sort everything while before only the duplicated ones were sorted (the others were on the hash), the results are not identical.

ksahlin:
>  if my quick intuitive reasoning is correct, it means that the overhead of the indexing array becomes ~4 Bytes per entry, which is very reasonable.

If you with 'entry' mean a strobemer seed, is it not 2 Bytes per entry? (But with 64bit pointers (TBI), needed for all references over ~20Gb, it would be 4Bytes per entry.)

luispedro:
Oh, yes, 2 Bytes. I was thinking that the index array had 8 Byte elements.

PR #279, opened 2023-05-17
Update io.hpp to fix MinGW build

teepean:
Building on MinGW gets the following error without this fix.

`io.hpp:21:5: error: 'uint64_t' was not declared in this scope`


marcelm:
Thanks! This will keep popping up until we have CI tests on MinGW.

PR #282, opened 2023-05-27
Fix tests to be compatible with MSVC/MinGW

teepean:


marcelm:
Thanks, that’s what I was hoping!

Issue #212, opened 2023-01-28
Follow recommended practice for SAM output


marcelm:
Section 2 in the [SAM specification](http://samtools.github.io/hts-specs/SAMv1.pdf) has a list of recommendations for SAM output. We should check the following ones:

- [x] The HD line should be present, with either the SO tag or the GO tag (but not both) specified.
- [ ] No alignments should be assigned mapping quality 255.
- [x] For a unmapped paired-end or mate-pair read whose mate is mapped, the unmapped read should
have RNAME and POS identical to its mate.
- [x] If all segments in a template are unmapped, their RNAME should be set as ‘*’ and POS as 0.
- [x] If POS plus the sum of lengths of M/=/X/D/N operations in CIGAR exceeds the length specified in the LN field of the SQ header line (if exists) with an SN equal to RNAME, the alignment should be unmapped, unless the reference sequence is circular (see below).
- [x] Unmapped reads should be stored in the orientation in which they came off the sequencing machine
and have their reverse flag bit (0x10) correspondingly unset.
- [x] SEQ and QUAL of secondary alignments should be set to ‘*’ to reduce the file size.
- [x] The NM tag should be present.

valentynbez:
Consider interoperability with `samtools` and `pysam`, as they're the most popular libraries for parsing SAM files.

PR #290, opened 2023-06-07
Fix secondary alignments not having CIGAR strings

### LINKED ISSUE #66: `-N 1` outputs no extra alignment
By default, StrobeAlign outputs no secondary alignments (`-N 0`). I’ve tried to see what happens on the test dataset when using `-N 1` and `-N 2`. The sixth read (`SRR1377138.6`) is the first to have a secondary alignment. I show an excerpt from the SAM file for it below. This is done on single-end reads.

With `-N 0`:
```
SRR1377138.6    0       NC_001422.1     649     8 ...
```
With `-N 1`:
```
SRR1377138.6    0       NC_001422.1     649     0 ...
```
With `-N 2`:
```
SRR1377138.6    0       NC_001422.1     649     0 ...
SRR1377138.6    256     NC_001422.1     649     255 ...
```
There are some things that I didn’t expect:
* No extra alignment is shown with `-N 1`. `-N 2` is required for that (as if `-N` would limit the total number of alignments).
* The mapping quality of the primary alignment changes from 8 to 0, but perhaps that is another manifestation of #52.

(The mapping quality for the second alignment is 255, but that is ok since 255 means "unavailable".)

### LINKED ISSUE #289: CIGAR is not recognized by `samtools`
Hello,

I compiled the development version of `strobealign` with cmake. 
I run the command: 
```
strobealign/build/strobealign -U -R 4 -N 1000 -M 1000 -t 32 genomes_chunked/FENG15-1_1.fa.gz crc_phages.100mer.fa.gz 
> FENG15-1_1.strobealign.N1000.sam
```
The output looks as following:
```
FENG15-1_SAMEA3136768_METAG-scaffold_777_phage_45_virsorter_4500-4556   272     FENG15-1_SAMEA3136632_MAG_00000057-scaffold_3   1       255     54M3S   *       0  0AGCCAGTAATGGGTTAAGTGATAACAGGTGTCTGGAAATATAGGGGCAAATCCAGCA       *       NM:i:0  AS:i:118
```

But the CIGAR is not recognized by `samtools`:
```
[W::sam_parse1] mapped query must have a CIGAR; treated as unmapped
```

marcelm:
This was a regression introduced in 51eab56e07dbbb88c67772e7cafa4b777646946f:
After `best_sam_aln = std::move(sam_aln)`, sam_aln should not be used, but we did use it when appending it to the alignments vector.

The  actual bugfix is in 6fb332684574a76569bbd29fcce0f787eb4aae51. This PR also sets SEQ and QUAL to `*` for secondary alignments as recommended by the SAM spec (see #212). (I verified that this is what BWA-MEM does as well.)

Closes #289
Closes #66

### DIFF
--- a/src/aln.cpp
+++ b/src/aln.cpp
@@ -80,15 +80,14 @@ static inline void align_SE(
     int max_tries,
     unsigned max_secondary
 ) {
-    Read read(record.seq);
-
     if (all_nams.empty()) {
         sam.add_unmapped(record);
         return;
     }
 
+    Read read(record.seq);
     std::vector<alignment> alignments;
-    int cnt = 0;
+    int tries = 0;
     float score_dropoff;
     Nam n_max = all_nams[0];
 
@@ -102,7 +101,7 @@ static inline void align_SE(
     for (auto &n : all_nams) {
         score_dropoff = (float) n.n_hits / n_max.n_hits;
 
-        if ( (cnt >= max_tries) || best_align_dist == 0 || score_dropoff < dropoff){ // only consider top 20 hits as minimap2 and break if alignment is exact match to reference or the match below droppoff cutoff.
+        if (tries >= max_tries || best_align_dist == 0 || score_dropoff < dropoff) { // only consider top 20 hits as minimap2 and break if alignment is exact match to reference or the match below droppoff cutoff.
             break;
         }
         statistics.tot_all_tried ++;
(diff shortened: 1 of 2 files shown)


ksahlin:
Great! Does this PR ([ca4febf](https://github.com/ksahlin/strobealign/pull/290/commits/ca4febf342b1444824197c6dfaeeefcdcb19058e)) also fix the off-by-one error in setting `-N` reported in issue https://github.com/ksahlin/strobealign/issues/66?

marcelm:
Ah, I think I just forgot to close #66: I think the off-by-one error has already been fixed by #235 (07dc9e42a3b0e73541eb3ae332c277b34a3cd1a5) by just passing `map_params.max_secondary + 1` to the function instead of `map_params.max_secondary`. The variable name was still misleading at that point and commit ca4febf now makes it so that the variable actually contains the maximum number of secondary hits (this makes `align_SE` consistent with `align_PE` consistent in that regard).

PR #291, opened 2023-06-07
Set RNAME and POS of unmapped read identical to mate

marcelm:
As recommended in the SAM specification:
"For a unmapped paired-end or mate-pair read whose mate is mapped,
the unmapped read should have RNAME and POS identical to its mate."

No reason is given, but this is probably to make the unmapped read be
adjacent to its partner in a coordinate-sorted BAM file.

See #212


ksahlin:
I had a quick look and was first confused by `mate_name1` and `mate_name2`, then I realised that this is probably the reference chromosome ID that the mate is mapped to, right? perhaps it should be named `ref_id_mate1` or something instead?

marcelm:
Good point; I just refactored the code to use somewhat more consistent variable names. The code in the Sam class uses the term "name" to describe reference ids and query ids because that’s consistent with what the SAM spec uses (RNAME, QNAME), so I opted for `mate_reference_name1`.

ksahlin:
Great, happy to merge!Sent from my iPhoneOn 8 Jun 2023, at 10:06, Marcel Martin ***@***.***> wrote:﻿
Good point; I just refactored the code to use somewhat more consistent variable names. The code in the Sam class uses the term "name" to describe reference ids and query ids because that’s consistent with what the SAM spec uses (RNAME, QNAME), so I opted for mate_reference_name1.

PR #295, opened 2023-06-12
Backport SSW fix

### LINKED ISSUE #294: `NM` tag might be not calculated correctly in some situation
I noticed that strobealign might not calculate NM tags correctly for some special cases. I generated an example as follow:

The reference file:
```
>ref
CTCCAAAGATGTGATCGAGTGCTTGGTCGAGTGATGGTATTGGTGGTGAGTGATGATACTCGACCAGGGGGTCGAGTAATATGGTGGAG
```

The query sequence:
```
>qry
CTAGGTATTGGTGGTGGTCTT
```

And tested it use this command:

```bash
trobealign -k 8 -A 4 -O 3 -L 100 ref.fa debug.fa 2>/dev/null | samtools calmd - ./ref.fa | grep different
```

**[bam_fillmd1] different NM for read 'qry': 17 -> 18**

The error message from samtools indicates that the `NM` tag is not correct, at least not in a stadard way. 

I count the mismatch manually, and the number is 18 rather than 17

```
CTA-GGTATTGGTGGTG-GTCTT
.|...|...|.|..|........
GTGATGGTATTGGTGGTGAGTGA
```

`|` for match and `.` for mismatch

marcelm:
SSW sometimes computes wrong alignments. This has been reported and fixed in https://github.com/mengyao/Complete-Striped-Smith-Waterman-Library/pull/78

This PR backports just makes the same change that was made in that PR.

Our copy of SSW still has some other bugs that have been fixed in more recent versions of SSW. I’ll look into updating SSW.

Closes #294

### DIFF
--- a/ext/ssw/ssw.c
+++ b/ext/ssw/ssw.c
@@ -674,7 +674,7 @@ static cigar* banded_sw (const int8_t* ref,
 	l = 0;	// record length of current cigar
 	op = prev_op = 'M';
 	temp2 = 2;	// h
-	while (LIKELY(i > 0)) {
+	while (LIKELY(i >= 0 && j > 0)) {
 		set_d(temp1, band_width, i, j, temp2);
 		switch (direction_line[temp1]) {
 			case 1:


marcelm:
Closing in favor of #296

PR #296, opened 2023-06-12
Update SSW to v1.2.5

### LINKED ISSUE #294: `NM` tag might be not calculated correctly in some situation

marcelm:
Obtained from
https://github.com/mengyao/Complete-Striped-Smith-Waterman-Library

The previously included copy of SSW had been modified by adding a score_size parameter to Align(). This customization has been removed. This does not change behavior as the passed value passed was the default used by the library anyway.

This is an alternative to #295

- [x] SSW can now fail to find an alignment. If that happens, it prints a message to stderr and sets a flag. We need to check the flag and handle the error case.
- [ ] ~~Also, we should patch SSW to not print out the error message (we want to control terminal output).~~ Decided to ignore this as it happens so rarely
- [x] Bump baseline commit as one of the alignments actually changes slightly
- [x] Measure whether accuracy changes in our test datasets

Closes #294

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,8 @@
 * #289: Fixed missing CIGAR for secondary alignments.
 * #212: SEQ and QUAL are set to `*` for secondary alignments as recommended
   by the SAM specification.
+* #294: Updated the alignment library (SSW), which fixes some incorrect
+  alignments.
 
 ## v0.10.0 (2023-06-05)
 
--- a/ext/ssw/README.md
+++ b/ext/ssw/README.md
@@ -1,0 +1,261 @@
+# SSW Library
+
+## An SIMD Smith-Waterman C/C++/Python/Java/R Library for Use in Genomic Applications
+
+License: MIT
+
+Copyright (c) 2012-2015 Boston College
+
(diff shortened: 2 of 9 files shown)


marcelm:
> So I guess this is the PR to be merged instead of #295 ?

Yes, that was just the first fix until I decided to do it properly by updating all of SSW.

Regarding the above To Do list:

I added some code that checks whether an alignment failed, but I could not trigger it. Since it seems to occur so rarely, I decided to leave in the message printed to the terminal. It’s just a cosmetic thing anyway.

I have checked  a couple of datasets and couldn’t find any difference in accuracy, which I guess is expected because the interval to which the query is mapped on the reference doesn’t change.

ksahlin:
Ok great, happy to merge.

PR #297, opened 2023-06-13
Fix: Allocated buffer was not correctly deleted

marcelm:
GCC 13 has a new warning that caught this:
```
warning: ‘void operator delete(void*, std::size_t)’ called on pointer returned from a mismatched allocation function [-Wmismatched-new-delete]
```

Background: Memory allocated with the `new[]` operator (distinct from just `new`) must be deallocated with the `delete[]` operator (distinct from just `delete`). The `unique_ptr` hides the `delete`, but the principle still applies.


ksahlin:
approved.

marcelm:
Whoops, I was in the wrong branch when I noticed this - the problematic code is now gone after PR #278 has been merged.

Issue #298, opened 2023-06-13
To do after merging new hash table memory layout


marcelm:
After merging #278, a couple of things need to be done. I’ll add them here as individual bullet points. We can convert anything that is not super easy to do into separate issues.

- [x] Update baseline commit
- [x] Update changelog
- [x] "Time estimating number of unique hashes" message is now incorrect
- [x] Add accessors for RefRandstrobeWithHash
- [x] Rename `RefRandstrobeWithHash` to `RefRandstrobe` since we no longer need a RefRandstrobe without hash
- [x] Add one or two comments describing the memory layout
- [x] Add error handling when invalid bits/`-b` values are provided
- [x] Some refactoring so that `nam.cpp` doesn’t need to know the memory layout of the index
- [x] #304
- [x] Remove leftover `get_count_line_search` declaration

marcelm:
Closing this issue and moving the last to-do item into a separate one.

PR #299, opened 2023-06-13
Some touchups after #278

marcelm:
See #298


ksahlin:
Had a look and all looks good from what I can see.

PR #300, opened 2023-06-13
Log index statistics in a nicer way

marcelm:
This is also part of #298, but I took the opportunity to make the index statistic logging a bit nicer, which I had wanted to do for a while.
Before:
```
Unique strobemers: 23324540
Total strobemers count: 26446802
Total strobemers occur once: 22560298
Fraction Unique: 0.97
Total strobemers highly abundant > 100: 1562
Total strobemers mid abundance (between 2-100): 762681
Total distinct strobemers stored: 23324540
```

After:
```
Index statistics
  Total strobemers:          26446802
  Unique strobemers:         23324540
    1 occurrence:            22560298 ( 96.72%)
    2..100 occurrences:        762681 (  3.27%)
    >100 occurrences:            1562 (  0.01%)
```

### DIFF
--- a/src/index.cpp
+++ b/src/index.cpp
@@ -140,8 +140,7 @@ void StrobemerIndex::populate(float f, size_t n_threads) {
 
     Timer count_hash;
     auto randstrobe_hashes = count_randstrobe_hashes_parallel(references, parameters, n_threads);
-    stats.elapsed_unique_hashes = count_hash.duration();
-    logger.debug() << "Count number of randstrobe hashes: " << randstrobe_hashes << '\n';
+    stats.elapsed_counting_hashes = count_hash.duration();
 
     Timer randstrobes_timer;
     add_randstrobes_to_vector(randstrobe_hashes);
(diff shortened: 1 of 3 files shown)


marcelm:
I thought about adding a hard-coded `(100%)` after "Unique strobemers" to clarify what the other percentages refer to. What do you think? Like this:
```
Index statistics
  Total strobemers:          26446802
  Unique strobemers:         23324540 (100.00%)
    1 occurrence:            22560298 ( 96.72%)
    2..100 occurrences:        762681 (  3.27%)
    >100 occurrences:            1562 (  0.01%)
```
But maybe it’s confusing that this is always 100%?

ksahlin:
I like the clarification

> I thought about adding a hard-coded (100%) after "Unique strobemers"

Okay, but in that case perhaps we should change to `Distinct strobemers` or `Total distinct strobemer hash values` instead of Ùnique`?

marcelm:
> Okay, but in that case perhaps we should change to `Distinct strobemers` or `Total distinct strobemer hash values` instead of `Unique`?

Do you suggest this because "unique" could also refer to the strobemers with one occurrence? That’s a good point. I have changed it to "distinct" now. The longer version doesn’t fit well and the index statistics are only shown when `-v` is used, so I think it’s fine. (We could at some point add a section to the documentation explaining what the numbers mean and how to interpret them.)

ksahlin:
> Do you suggest this because "unique" could also refer to the strobemers with one occurrence?

Yes, that is what I was referring to.

Ok great. good to merge.

PR #301, opened 2023-06-14
Fail somewhat gracefully for too large references

marcelm:
This improves handling of large references by hopefully crashing with a somewhat helpful error message (instead of just `std::length_error` or so).

Also, this limits the bits (`-b`) to the range 8..31. It may work up to 32, but this needs some further checking.

* Fix counting of randstrobes (this was just a 32 bit unsigned int and would therefore overflow on large references).
* Introduce a `bucket_index_t` type alias. The intention is for this to become a template parameter at a later stage, which can then be instantiated to be either a `uint32_t` or `uint64_t`.


ksahlin:
great, looks good to merge.

PR #302, opened 2023-06-15
Speed up index generation slightly by forcing inlining of XXH64

marcelm:
See https://fastcompression.blogspot.com/2018/03/xxhash-for-small-keys-impressive-power.html

This speeds up index generation by ~3%.


ksahlin:
super! approved

PR #312, opened 2023-06-20
Various refactorings

marcelm:
- Use a `std::vector<Syncmer>` instead of one vector for hashes and one for positions
- Rename `RandstrobeIterator2` to `RandstrobeGenerator` and add comments to explain what the difference is (to make it  clear that we need both at the moment)
- Tiny bugfix: Remove the `v` prefix from the version in the SAM PG header (only relevant when the version number is derived from a Git tag)
- Rename `alignment` struct to `Alignment`
- Move indexing parameters into separate `SyncmerParameters` and `RandstrobeParameters` structs

### DIFF
--- a/GitVersion.cmake
+++ b/GitVersion.cmake
@@ -5,7 +5,9 @@ execute_process(
   ERROR_QUIET
   OUTPUT_STRIP_TRAILING_WHITESPACE
 )
+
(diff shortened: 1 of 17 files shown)


ksahlin:
Great, makes it much clearer. Approved!

PR #316, opened 2023-06-22
Add --count option to dumpstrobes command

marcelm:
I found this useful for experiments: This adds a `--count` option to `dumpstrobes`, which will make it count all randstrobes (or syncmers if `--syncmers` is also provided).

This can also be used for benchmarking the `RandstrobeGenerator` and `SyncmerIterator` (only the overhead from reading the reference has to be subtracted from the measured runtime).


ksahlin:
Nice! approved.

PR #323, opened 2023-07-04
Fix strobemer count statistics

marcelm:
There was a race condition in updating `stats.tot_strobemer_count`, which led to the statistics being off. The strobemer count is already available from the first (counting-only) pass over the reference; we can just use that number instead.

This PR also includes a couple of refactoring commits.


ksahlin:
Great, approved.

PR #326, opened 2023-08-21
Refactoring

marcelm:
I worked a little bit on strobealign during my vacation and accumulated these refactorings. Most of them rename something to increase readability (hopefully).

One functional change is here as well: We now ensure that the value provided for `-M` is at least 1.


ksahlin:
Much clearer, approved.

PR #327, opened 2023-08-21
Avoid MAPQ=60 for single-end multimappers

### LINKED ISSUE #293: Reads get MAPQ=60 although they are multimappers
Apparently some (possibly many) reads get MAPQ 60 although they actually are multimappers that should get MAPQ 0.

I noticed this when running the `samdiff.py` script to compare two strobealign commits. In one commit, I had changed the randstrobe hash to an asymmetric one in order to avoid getting the spurious hits on the reverse complement. The exact change doesn’t matter so much because I now realize I have seen the following earlier when tweaking the mapping parameters or the algorithm.

The samdiff script often outputs changes that look like this:
```
simulated.61
           MAPQ: 60
          score: 120
             NM: 0
          CIGAR: 50M
            ref: 2L -> 2R
            pos: 1680535 -> 19717413
```
So in one version of the program, the read `simulated.61` is mapped to 2L:1680535 and in the other, it is mapped to `2R:19717413`. In both cases, however, the alignment score is 120, which means that the loci are actually equally good (and in this case, they are even identical). That is, since the read maps equally well to both locations, MAPQ should be 0.

One explanation could be that the program simply doesn’t find the respective other locus, but at least for the above read, both versions find both loci at the same score. The set of produced NAMs is essentially the same if one looks only at the highest-scoring ones, they just are found in a different order.

Possibly this is a regression introduced by one of the changes I made; I will need to check this.

See also #25

marcelm:
by ensuring that we compute at least two alignments even if the first that is found has edit distance 0.

Closes #293

Because I still need to test whether this actually fixes #293, I’m marking this as draft.

### DIFF
--- a/src/aln.cpp
+++ b/src/aln.cpp
@@ -100,7 +100,7 @@ static inline void align_SE(
     int min_mapq_diff = best_edit_distance;
     for (auto &nam : nams) {
         float score_dropoff = (float) nam.n_hits / n_max.n_hits;
-        if (tries >= max_tries || best_edit_distance == 0 || score_dropoff < dropoff_threshold) {
+        if (tries >= max_tries || (max_tries > 1 && best_edit_distance == 0) || score_dropoff < dropoff_threshold) {
             break;
         }
         bool consistent_nam = reverse_nam_if_needed(nam, read, references, k);
--- a/src/aln.cpp
+++ b/src/aln.cpp
@@ -98,9 +98,10 @@ static inline void align_SE(
     best_alignment.score = -100000;
     best_alignment.is_unaligned = true;
     int min_mapq_diff = best_edit_distance;
+
     for (auto &nam : nams) {
         float score_dropoff = (float) nam.n_hits / n_max.n_hits;
-        if (tries >= max_tries || (max_tries > 1 && best_edit_distance == 0) || score_dropoff < dropoff_threshold) {
+        if (tries >= max_tries || (tries > 1 && best_edit_distance == 0) || score_dropoff < dropoff_threshold) {
             break;
         }
         bool consistent_nam = reverse_nam_if_needed(nam, read, references, k);
@@ -125,7 +126,6 @@ static inline void align_SE(
         }
         tries++;
     }
-
     if (max_secondary == 0) {
         best_alignment.mapq = std::min(min_mapq_diff, 60);
         sam.add(best_alignment, record, read.rc, true, details);


ksahlin:
approved.

marcelm:
(Note that I had a typo/logic error in the PR and pushed a fix.) 

I tried to do some measurements.

It’s a bit hard to inspect multimappers only, so I looked at incorrectly mapped reads instead, which should all be multimappers: Assuming a couple of things about how the read mapper works, these should have at least two roughly equivalent mapping locations (from the read mapper’s point of view): The correct one and the one the read mapper actually outputs. So they should get MAPQ=0.

So issue #293 is essentially that we get many incorrectly mapped reads that are assigned MAPQ>0, which should actually have been assigned MAPQ=0.

I generated a test dataset with know mapping locations by picking every tenth 100-mer from the *Drosophila* reference (about 14.4 million reads). Then I could count incorrectly mapped reads that have MAPQ>0 and MAPQ=0, respectively. For good measure, I also counted the *correctly* mapped reads with MAPQ>0 and MAPQ=0. The results are as follows:

| Read category | main (40748fe) | this PR (097807b5cf9c4f8fb00c2e1d20f29ac8be776da4) | BWA-MEM | comment |
|-|-:|-:|-:|
Incorrect and MAPQ>0 (bad) |   9.87% |  **0.54%** |0% |
Incorrect and MAPQ=0 (good) |   0.03% |  **9.36%** |9.94% |
Correct and MAPQ>0   |  90.08% | 87.20% |86.63% |
Correct and MAPQ=0   |      0% |  2.88% |3.41% |

I think the relevant numbers are the two bold ones: The "bad" alignments are much closer to zero and the number of "good" ones has gone up. Also, all four percentages are now closer to BWA-MEM.

Note that the percentage of "correct and MAPQ>0" goes down (and "correct and MAPQ=0" up accordingly). I assume these are mostly multimappers that strobealign by chance placed into the "correct" position (the one from which the k-mer was sampled). They previously got MAPQ=60, but since they are multimappers, they should get MAPQ=0, which they now do.

ksahlin:
great improvement, approved!

marcelm:
I forgot to add a changelog entry in this PR and took the liberty of pushing that directly to main without a PR.

marcelm:
I realized I hadn’t check the performance impact of this change, so here are some measurements on 100 bp reads. 

* Drosophila: ~3% slower
* Maize: ~9% slower
* CHM13: ~5% slower
* Rye: ~5% slower

These measurements were quite noisy, especially for rye.

ksahlin:
Ok, worth it.

marcelm:
Agreed!

PR #330, opened 2023-08-26
Get rid of add_to_hits_per_ref() min_diff parameter

marcelm:
The intention of this parameter is to disallow hits that span too different lengths on the query compared to the reference.

The parameter has a bit of a misleading name because it is actually the *maximum* allowed difference. (This can come from some of my refactoring. Within the function, it represents the minimum observed difference so far. The value provided as argument is merely its initial value.)

But mainly, the parameter has no effect at all and just initializing it to `std::numeric_limits<int>::max()` leads to the same decisions being made. (I temporarily added some code tested this.)

### DIFF
--- a/src/nam.cpp
+++ b/src/nam.cpp
@@ -16,9 +16,9 @@ void add_to_hits_per_ref(
     int query_end,
     bool is_rc,
     const StrobemerIndex& index,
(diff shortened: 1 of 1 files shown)


ksahlin:
approved.

Issue #321, opened 2023-06-29
Scoring of read pairs on same vs. different chromosome


marcelm:
> do you think it would make sense / be worth to score read pairs on the same chromosome (but far away) higher than reads on separate chromosomes? Currently I think there is only the binary 'close and proper' or 'not close and proper' classification.

_Originally posted by @ksahlin in https://github.com/ksahlin/strobealign/issues/319#issuecomment-1611405859_

ksahlin:
I think we have detected a bug if my reasoning below is correct, nice!

The score was implemented so that the further away from mean the isize is, the lower score.

As seen here https://github.com/ksahlin/strobealign/blob/main/src/aln.cpp#L1185 read pairs that map more than 5*sigma away gets a score lower than disjoint reads. So in the scoring that happens here: https://github.com/ksahlin/strobealign/blob/main/src/aln.cpp#L1169 , we should probably have something like:

`S = (double)a1.sw_score + (double)a2.sw_score + std::max( log(normal_pdf(x, mu, sigma)), -20 + epsilon_pair );` (`log(normal_pdf(x, mu, sigma))` is always negative) 

In the proposed line above `epsilon_pair` is a small value to prefer pairs mapping 'together' rather than on, e.g., different chromosomes. 

The reason I claim that our current scoring is a bug is because we will actually penalize proper but far away mates more than completely separated reads (if sw scores identical). This is likely bad for detecting larger deletions.

ksahlin:
To add a bit of data, I get below penalties libraries with mu=300, sigma=30 (which I think we use in our simulations?). 

So, we get that proper pairs with distance higher than about 470 would be penalized more than individual best alignments (e.g. on different chromosomes). 

Also, we set `x = std::abs(a1.ref_start - a2.ref_start);`. Is this the difference of 5' starts in both reads? If so, we penalize close-by pairs harshly.

```
x: 30 penalty: -44.820135914866825
x: 60 penalty: -36.320135914866825
x: 90 penalty: -28.820135914866828
x: 120 penalty: -22.320135914866828
x: 150 penalty: -16.820135914866828
x: 180 penalty: -12.320135914866828
x: 210 penalty: -8.820135914866828
x: 240 penalty: -6.320135914866828
x: 270 penalty: -4.820135914866828
x: 300 penalty: -4.320135914866828
x: 330 penalty: -4.820135914866828
x: 360 penalty: -6.320135914866828
x: 390 penalty: -8.820135914866828
[...]

PR #331, opened 2023-08-26
Refactor align_PE

marcelm:
I’m working my way through `align_PE` in order to understand how pairs are scored (so that I can work on #321). I came up with these refactorings/cleanups along the way.

### DIFF
--- a/src/aln.cpp
+++ b/src/aln.cpp
@@ -407,8 +407,8 @@ static inline std::vector<NamPair> get_best_scoring_nam_pairs(
     }
 
     // Find high-scoring R1 NAMs that are not part of a proper pair
-    Nam dummy_nan;
-    dummy_nan.ref_start = -1;
+    Nam dummy_nam;
+    dummy_nam.ref_start = -1;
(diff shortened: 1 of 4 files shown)


src/aln.cpp
```diff
@@ -371,62 +376,54 @@ static inline std::vector<std::tuple<int,Nam,Nam>> get_best_scoring_nam_location
...
+                joint_nam_scores.emplace_back(joint_hits, nam1, nam2);
+                added_n1.insert(nam1.nam_id);
+                added_n2.insert(nam2.nam_id);
+                best_joint_hits = std::max(joint_hits, best_joint_hits);
             }
         }
     }
 
+    // Find high-scoring R1 NAMs that are not part of a proper pair
     Nam dummy_nan;
```
ksahlin, line 410: supposed to be dummy_nam

ksahlin:
approved

src/aln.cpp
marcelm: Indeed! Didn’t even notice ... (fixed now)

PR #337, opened 2023-09-06
More refactorings

marcelm:
* Rename the last three classes/structs that did not have CamelCase names.
* Rename `get_alignment` to `extend_seed`
* etc.


ksahlin:
Nice, approved.

Issue #272, opened 2023-04-27
Consider reducing k for read length 50


marcelm:
We discussed this during our recent meeting: Reducing $k$ for read length 50 from the default of $k=20$ gives higher mapping rate and accuracy on the test datasets.

Here are the results for 2 x 50bp paired-end reads. ~~I will edit the table as soon as I have all results for rye.~~ Numbers added.

I don’t have good numbers for mapping times, yet. They appear to go up by <10%. I will get some numbers as soon as I have a quiet system.

## Mapping rate

Reference | $k = 20$ | $k=19$ | $k=18$ | $k=17$ |
-|-|-|-|-
drosophila | 99.46615| 99.56685| 99.64465 | **99.69955** |
CHM13      | 99.316  | 99.39755|99.4694|**99.51295**|
maize      |99.41475|**99.44605**|99.4441|99.4203|
rye  | **99.22915** |99.22355 | 99.22355 | 99.0205 | 

## Accuracy

Reference | $k = 20$ | $k=19$ | $k=18$ | $k=17$ |
-|-|-|-|-
drosophila |90.0774|90.1818|90.3049|**90.3616**|
CHM13      |90.4725|90.6161|90.69315|**90.786**
maize      |71.2683|71.54695|71.72395|**71.9865**
rye  |68.88975 |69.16655|69.37835 |**69.53555**

ksahlin:
Just to double check: Accuracy is calculated as `#correct/total_reads`, and not `#correct/#mapped_reads`, right? (I guess you use a modified jaccard version of the script I used in the paper, In thet case it should be divided on total_reads)

Based on the accuracy and mapping rate, k=17 or 18 looks like interesting alternatives. Lowering k for canonical read length 50 will lower it for read lengths < 91 IIRC.

Is it easy to check the 2*75 nt reads as well? My intent is not to be annoying with extra analyses :) there is a reason behind it:  
 First, 75nt reads may be more common than 50nt(?). Second, as we know there is a tradeoff in k. With lower k we may find more 'non-destroyed' seeds (good). However, seeds also get more repetitive and we lose them to masking (bad; as hinted in the mapping rate table for the k=17 on rye genome).  For 50nt reads the good effect is obviously much more pronounced. However, with longer read length this good effect may not be as large.

ksahlin:
Interesting. Not crystal clear what to do with this. Some suggestions:

[...]

PR #338, opened 2023-09-08
Allow w_min to be zero

marcelm:
There’s nothing inherently problematic with `w_min=0` and, as we see in #272, it even leads to higher accuracy for 50 bp reads.

This requires changing the `l` parameter from -4 to -3 for 50 bp reads if we want w_min to remain equal to 1 as before.


ksahlin:
Approved

PR #342, opened 2023-09-19
Factor out a randstrobe_hash function

marcelm:
Just a tiny refactoring to make it easier to change the way the syncmer hashes are combined into a single randstrobe hash.


ksahlin:
Approved.

PR #344, opened 2023-09-19
Fix read length estimation crash on chromosome-sized "reads"

### LINKED ISSUE #343: Mapping CHM13 against itself results in a crash
When running `strobealign -x CHM13.fa CHM1.fa`, we get:
```
Estimated read length: -1756587512 bp
```
and later on:
```
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
```
(Maybe these are distinct problems.)

marcelm:
Closes #343

(This does not mean that it is actually possible to run CHM13 vs CHM13 in a reasonable amount of time, strobealign just doesn’t crash anymore before getting stuck.)

### DIFF
--- a/src/readlen.cpp
+++ b/src/readlen.cpp
@@ -1,8 +1,8 @@
 #include "readlen.hpp"
 
-int total_read_len(const std::vector<klibpp::KSeq> &records) {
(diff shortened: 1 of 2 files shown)


ksahlin:
Approved.

PR #346, opened 2023-09-21
Replace XXH64 with a simplified version

marcelm:
This function gives the same result as regular XXH64(), but with the difference that it only works on single 64-bit values, which makes it possible to reduce it to just a couple of lines; mostly the "bit mixing" part (functions finalize()/avalanche()) is left.

This performs a little bit better than calling the original function on an 8-byte slice because the compiler (at least GCC) does not fully inline the original function (finalize() remains a function call).

This speeds up index generation by a couple of percent: 163 are reduced to 159 seconds for CHM13.

### DIFF
--- a/src/hash.hpp
+++ b/src/hash.hpp
@@ -1,0 +1,120 @@
+#ifndef STROBEALIGN_HASH_HPP
+#define STROBEALIGN_HASH_HPP
+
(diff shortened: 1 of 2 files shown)


ksahlin:
Approved.

PR #345, opened 2023-09-20
Optimize parameters

marcelm:
This is the result of an exhaustive search for parameter combinations that improve accuracy over the currently used ones. (In total, strobealign ran over 7000 times.)

~~Due to a typo, I need to re-do some measurements for length 200, and length 500 is not finished, yet. I will updated the PR once they are done.~~

These are the changes I suggest so far. The "improvement" columns show how much avergae accuracy increases. As in #339, I have ranked the parameter combinations by PE accuracy plus 1/4 SE accuracy.

Canonical read length | Current parameters | suggested parameters | SE improvement | PE improvement
-|-|-|-|-
50    | (20, 16, -3, 2) | (18, 14, -2, -2) | +2.478 | +1.116
75    | (20, 16, -3, 2) | (20, 16, -2, -2) | +0.185 | +0.288
100   | (20, 16, -2, 2) | (20, 16, -1, -1) | +0.070 | +0.137
125   | (20, 16, -1, 4) | (20, 16, -1,  3) | no data, interpolated
150   | (20, 16, 1,  7) | (20, 16, 1,  7)  | no change, already best
250   | (20, 16, 4, 13) | _(22, 18, 4, 14)_  | _+0.028_ | _+0.020_
~~300~~   | ~~(22, 18, 2, 12)~~ | ~~(23, 19, 6, 12)~~ | ~~+0.042~~ | ~~+0.020~~
300*  | (22, 18, 2, 12) | (22, 18, 3, 13) | +0.033 | +0.013
400   | (23, 17, 2, 12) | _(23, 17, 5, 11)_ | _+0.040_ | _+0.015_

* The suggested parameters for read length 50 are different from those we agreed on in #272 because I have now tested more parameter combinations. The one given here improves PE accuracy even more (at the cost of some SE accuracy).
* The second suggestion for length 300 is the second-best parameter combination. I kept it because the it has $u-l=10$ like the current parameters. I have no idea whether that has any importance, though.
* Suggestion for canonical read length 250 is based on measurements on the 200 bp dataset.
* Suggestion for canonical read length 400 is based on measurements on the 500 bp dataset.

## To Do

* [X] Decide what to do with canonical read length 125
* [x] Decide which 300-bp-parameters to use
* [x] Add changelog entry

I have the accuracy measurements in an SQLite database file and can relatively easy produce tables to show more detailed statistics if needed.

* Closes #272
* Closes #339 

*Edit:* Added numbers for read lengths 200 and 500 (canonical 250 and 400).

<details>

[...]

PR #347, opened 2023-09-27
Reduce memory usage when indexing

marcelm:
A full copy of each reference sequence was inadvertently made in each worker thread.

Reduces memory usage, especially when indexing genomes with large scaffolds in parallel.

When indexing rye with 8 threads, memory usage is reduced from 35.2 GiB to 33.4 GiB.

See #345


ksahlin:
Perfect, approved!

PR #348, opened 2023-09-28
Do not require strobealign to be in PATH when running tests

marcelm:
This just removes a minor inconvenience.


ksahlin:
approved

PR #353, opened 2023-10-02
Add dumpstrobes --seeds option for writing out the full seed vector

marcelm:
With this, `dumpstrobes --seeds genome.fa` dumps the sorted randstrobes vector to standard output in a comma-separated format.

Columns:
- hash
- position
- reference index
- strobe 2 offset

I also had to move two functions in `index.cpp` into an anonymous namespace as otherwise the linker complains about a duplicate name (`count_randstrobes` exists twice).


ksahlin:
Thanks! Approved.

Issue #359, opened 2023-11-13
Assign multimappers randomly to one of the candidate mapping positions


marcelm:
(As discussed by e-mail.)

For reads with multiple equally good mapping locations, we currently always pick the first location.

To reduce biases, we should instead pick one of the locations randomly.

PR #360, opened 2023-11-13
Assign single-end multimappers randomly to one of the candidate mapping locations

marcelm:
As before, we iterate over NAMs, compute alignments and keep track of the currently best one.

This PR also handles alignments that have the same score as the currently best one. When found, we use reservoir sampling to uniformly pick one of them.

This does not use the NAM score, but the alignment scores, which means we need to compute the alignments. We compute the alignments anyway, but it may slow things down in a certain case: I had to restrict the optimization that allowed us to stop computing alignments when we have found two exact matches (i.e. edit distance zero). Now we need to compute all exact matches until we find a nonexact one because we need to pick among the exact ones.

I will need to measure how much this slows things down.

See #359

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,8 @@
 * #318: Added a `--details` option mainly intended for debugging. When used,
   some strobealign-specific tags are added to the SAM output that inform about
   things like no. of seeds found, whether mate rescue was performed etc.
+* #359: Assign single-end multimappers randomly to one of the candidate
+  mapping locations. This should reduce biases.
 
 ## v0.11.0 (2023-06-22)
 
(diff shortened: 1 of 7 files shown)


ksahlin:
As discussed over email:

- I would prefer to solve this at the NAM stage. The goal is alignment to many similar genomes (which btw we should at a standard test instance with alignment to 10 or 50 E. coli genomes), and this would cause a drastic slowdown in such cases.

- One way to address the issue is to shuffle all top scoring NAMs, but keep all other functionality as is. I predict this would solve the large majority of cases without offering a noticeable slowdown. The cases it wouldn't work would be:  we get three NAMs that have (NAM-) scores 100, 100, 90, but the alignments all have the same score. Then just shuffling the top-scoring NAMs gives us the first or second alignment, but not the third. I am hoping such cases are rare.

marcelm:
Superseded by #364.

PR #362, opened 2023-11-14
Let baseline comparison fail if no reads were in the input

### LINKED ISSUE #361: Baseline comparison CI job passes unexpectedly
Commit a290e7b from PR #350 changed mapping results slightly, but this was not caught in CI.

For some reason, the [input file was empty when strobealign mapped the reads](https://github.com/ksahlin/strobealign/actions/runs/6339997543/job/17220433185#step:12:19). Then the BAM output was empty and the `samdiff.py` script apparently succeeds when one of the input files is empty.

marcelm:
This should ensure that the baseline comparison CI job doesn’t pass inadvertently as it did previously (see #361).

This also updates the baseline commit. To ensure we didn’t miss any changes in mapping results just because the samdiff script didn’t work, I ran it manually for each commit since the problem was introduced and can confirm that no other commit caused any changes.

Closes #361

### DIFF
--- a/tests/baseline-commit.txt
+++ b/tests/baseline-commit.txt
@@ -1,1 +1,1 @@
-baseline_commit=91f22234baf192a3167354a7f2bbfefb49cb2162
+baseline_commit=94c20bec1731b1c06d8090503cc4a0a402f6efb2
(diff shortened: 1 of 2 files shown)


marcelm:
This is blocking some other PRs and should be uncontroversial, so I’m taking the liberty of merging this without explicit approval.

ksahlin:
Yes absolutely. Please take the liberty in such occasions in general.

marcelm:
Thank you! I will do so in the future.

PR #363, opened 2023-11-14
Fix matches ending too early in PAF output

### LINKED ISSUE #333: End coordinate of match in PAF output
PAF output in strobealign uses these two values for the end position of the match on the query and reference:

* `nam.query_prev_hit_startpos + k`
* `nam.ref_prev_hit_startpos + k`

The `*_prev_hit_startpos` attributes describe the start position of the last hit in the NAM. Adding *k* to that gives the end position of the first strobe of the last strobemer. So apparently, the intention here is to not include the gap between the two strobes in the match. Asked about why this i the case, Kristoffer replied:

> This is probably a bug. If there was some intent for this I have forgotten it. Also, I cannot find a reason for it being like that. If you cannot think of any reason either, I vote for changing it to end_pos.

If the attributes are not needed for PAF output, it would be possible (with some other adjustments) to remove them entirely from the `Nam` struct.

marcelm:
Closes #333


ksahlin:
Approved.

PR #364, opened 2023-11-16
Choose random mapping locations for multimappers

marcelm:
This is an alternative to #360 that works by shuffling the top NAMs that have the same score as the best one.

See #359

`samdiff.py` outputs this when comparing to main:
```
       9643 reads were unmapped before and after
          0 reads became mapped
          0 reads became unmapped
      82741 reads were mapped to same locus before and after
*      7421 reads were multimapper before and after, same alignment score (AS)
*        17 reads were multimapper before and after, better alignment score (AS)
*         5 reads were multimapper before and after, worse alignment score (AS)
*       173 reads changed in another way
     100000 total reads
```
The high number of reads that get the same alignment score before and after is a sign that this works as intended. There are some reads that get a different score, but this is as expected and the price to pay for this faster variant that only shuffles the NAMs.

Accuracy changes in an unremarkable way (I ran this comparison only on a subset of the datasets):

dataset           | main (8c74210) | this PR | difference
-|-|-|-
drosophila-50-se  | 82.5175 | 82.5108 | -0.0067
drosophila-100-se | 89.2626 | 89.255 | -0.0076
drosophila-150-se | 90.937 | 90.9448 | +0.0078
drosophila-200-se | 91.9463 | 91.9559 | +0.0096
drosophila-300-se | 93.2153 | 93.2187 | +0.0034
maize-50-se       | 47.4622 | 47.4417 | -0.0205
maize-100-se      | 70.5148 | 70.503 | -0.0118
maize-200-se      | 86.7001 | 86.7083 | +0.0082
maize-300-se      | 91.8516 | 91.8522 | +0.0006
CHM13-50-se       | 81.7231 | 81.7321 | +0.0090
CHM13-100-se      | 90.6468 | 90.6401 | -0.0067
CHM13-150-se      | 92.4066 | 92.4081 | +0.0015
CHM13-200-se      | 93.221 | 93.2167 | -0.0043
CHM13-300-se      | 94.1508 | 94.1472 | -0.0036

Average difference se: -0.0015

We don’t have a way at the moment to test whether this results in the intended outcome, that is, whether locations for multimappers are less biased.

PR #365, opened 2023-11-17
Test secondary alignments

marcelm:
An improvement I plan to make for #364 affects the code that outputs the secondary alignments. We didn’t have tests covering that, so this adds a couple.

There’s also a change in here related to how empty reads are encoded in SAM output. `samtools` insists it should be `*`, so this is changed here as well.


ksahlin:
Approved

PR #368, opened 2023-11-22
Refactor aln.cpp

marcelm:
This renames:

* `align_SE` → `align_single`
* `align_PE` → `align_paired`
* `align_SE_read` → `align_or_map_single`
* `align_PE_read` → `align_or_map_paired`
* `joint_nam_scores` → `nam_pairs` (since we now use the `NamPair` struct)
* `rescue_mate` → `rescue_align` (could never remember what it does differently from `rescue_read`)
* Rename `get_mapq` to `proper_pair_mapq` (since this is only used to compute MAPQ for uniquely mapping proper pairs)

Also, `InsertSizeDistribution` is moved to a separate file.

The main change is to factor out an output_aligned_pairs() function. The rescue_read() and align_paired() functions previously did two things:
- Compute a list of paired alignments
- Output them to SAM

When working on random alignment selection for #367, I noticed that the code outputting to SAM was duplicated (and had slightly diverged already). This commit moves that part to a
new function output_aligned_pairs(). Now rescue_read() and align_paired() just
return a vector of ScoredAlignmentPairs.

This should fix at least two bugs:
- The details for R1 and R2 were swapped when R1 was rescued
- MAPQ of secondary alignments was inconsistently set to 0 or 255 depending on whether it was a rescued alignment or not.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,6 +63,7 @@ add_library(salib STATIC ${SOURCES}
   src/readlen.cpp
   src/version.cpp
   src/io.cpp
+  src/insertsizedistribution.cpp
   ext/xxhash.c
   ext/ssw/ssw_cpp.cpp
   ext/ssw/ssw.c
--- a/src/aln.cpp
+++ b/src/aln.cpp
@@ -410,9 +410,6 @@ static inline std::vector<NamPair> get_best_scoring_nam_pairs(
         }
     }
 
-    added_n1.clear();
-    added_n2.clear();
-
     std::sort(
         nam_pairs.begin(),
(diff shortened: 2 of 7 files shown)


ksahlin:
I haven't had the time to look at all the changes in this PR in detail. But I agree with the name changes and with what you describe, without reservation. 

If this PR is holding up any other development, or if you think its mostly non-controversial changes, I suggest a merge without a careful check.

marcelm:
Do you want to review this PR more closely or shall I just merge it?

ksahlin:
Won't have time anytime soon unfortunately . Yes please go ahead and merge.

PR #370, opened 2023-11-27
With --details, add a SAM tag that shows the no. of identically scored best alignments

marcelm:
This is the number of alignments among which we (pseudo-randomly) choose one to output.

Is this useful? I thought I needed this for debugging, but then never actually looked at the numbers.

Regarding the name of the tag: The old `bwa aln` used `X0` for "Number of best hits", which I think is the same thing. There is a `H0` tag in the SAM specification with the description "Number of perfect hits", which could be something similar, but I could not find what it is actually intended to be used for, so I went with `X0`.

This requires #368 to be merged first.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -68,7 +68,7 @@ style guide, but new code should follow it.
 
 ## Detailed SAM output (`--details`)
 
-When `--details` or `-d` is provided, the following additional SAM tags are output for
+When `--details` is provided, the following additional SAM tags are output for
 mapped reads.
 
 `na`: Number of NAMs (seeds) found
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -77,3 +77,6 @@ mapped reads.
   the expected region given its mate)
 `al`: Number of times an attempt was made to extend a seed (by gapped or ungapped alignment)
 `ga`: Number of times an attempt was made to extend a seed by gapped alignment
+`X0`: Number of equally scored best alignments (greater 1 for multimappers).
```
ksahlin, line 80: greater 'than' 1 ?

ksahlin:
Could be nice for development. Perhaps it would be informative for downstream applications as well?

ksahlin:
approved

PR #371, opened 2023-11-28
Add --no-PG option for not outputting the PG SAM header

marcelm:
This is useful for our tests where we need output to be consistent even if the version number of the program changes.

The name of the option is chosen to be consistent with `samtools view`.


ksahlin:
Approved

PR #374, opened 2023-12-04
Fix first index of randstrobe start index to be always the 0th randstrobe

hasin-abrar:
Previously 0th index was always skipped at the randstrobe start index. Also, count variable was initialized to 0, so we would end up in the ++tot_mid_ab case, even if this was a single occurrence.

### DIFF
--- a/src/index.cpp
+++ b/src/index.cpp
@@ -180,9 +180,10 @@ void StrobemerIndex::populate(float f, size_t n_threads) {
 
     uint64_t unique_mers = randstrobes.empty() ? 0 : 1;
     randstrobe_hash_t prev_hash = randstrobes.empty() ? 0 : randstrobes[0].hash;
-    unsigned int count = 0;
-
-    for (bucket_index_t position = 0; position < randstrobes.size(); ++position) {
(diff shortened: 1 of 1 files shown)


src/index.cpp
```diff
@@ -180,9 +180,10 @@ void StrobemerIndex::populate(float f, size_t n_threads) {
 
     uint64_t unique_mers = randstrobes.empty() ? 0 : 1;
     randstrobe_hash_t prev_hash = randstrobes.empty() ? 0 : randstrobes[0].hash;
-    unsigned int count = 0;
-
-    for (bucket_index_t position = 0; position < randstrobes.size(); ++position) {
+    unsigned int count = 1;
+    // first randstrobe index will always be the 0th idex
```
marcelm, line 184:
```suggestion
    // first randstrobe index will always be 0
```

marcelm (commented):
Thanks! Just some comments regarding formatting

hasin-abrar:
Hi Marcel, thanks for the comments. I have updated the formatting accordingly.

marcelm:
Looks good now, thanks again!

PR #379, opened 2023-12-12
Use optimized indexing parameters

marcelm:
Also, remove canonical read length 300 because after changing the parameters, 250 and 300 were using the same.


ksahlin:
Approved.

PR #389, opened 2024-01-19
Fix drop in mapping-only accuracy

marcelm:
Commit 1fe3e34 made a buggy refactoring: I thought that the `score` attribute of `NamPair` objects is the same as the sum of the scores of the two constituent NAMs. But that is incorrect: The `score` is actually the sum of the `n_hits`.

This PR fixes that and renames `NamPair::score` to `NamPair::n_hits` to eliminate this confusion.

### DIFF
--- a/src/aln.cpp
+++ b/src/aln.cpp
@@ -15,7 +15,7 @@ using namespace klibpp;
 namespace {
 
 struct NamPair {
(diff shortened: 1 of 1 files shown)


ksahlin:
Nice, approved!

PR #391, opened 2024-01-22
Simplify baseline comparison on CI

marcelm:
Do not let GitHub Actions do the work, but rely on the baseline comparison script. Only ensure caches are filled before starting it.

This reduces a little bit of code in the GitHub Actions workflow file (reduces essentially duplicate code).

We will only be able to see whether retrieving the cached baseline file works correctly on the next PR.


ksahlin:
approved

PR #392, opened 2024-01-22
Vendor zstr instead of fetching it from Git at build time

marcelm:
This way, the strobealign sources are more self-contained and it should not be necessary to have internet access at build time.


ksahlin:
Approved

PR #394, opened 2024-02-13
strobealign-aemb module for metagenomic binning

psj1997:
Hi,

This is the aemb (abundance estimation for metagenomic binning) module,  the usage is :

strobealign ref.fa read1.fq read2.fq --aemb > abun.txt 

In my benchmarking, aemb can output similar results, but decrease more than 90% running time.

Sincerely
Shaojun


ksahlin:
Thanks Shaojun. I think Marcel is a bit busy right now but I am sure he will have a look when there is sufficient time.

src/aln.cpp
```diff
@@ -903,6 +910,73 @@ inline void get_best_map_location(
...
+            if (!nams1.empty()){
+                for (auto &t : nams1){
+                    if (t.score == nams1[0].score){
+                        best_ref1.push_back(t);
+                    }else{
+                        break;
+                    }
+                }
+
+                int ref_size1 = best_ref1.size();
```
marcelm, line 953:
```suggestion
                size_t ref_size1 = best_ref1.size();
```

src/aln.cpp
```diff
@@ -903,6 +910,73 @@ inline void get_best_map_location(
...
+                    break;
+                }
+            }
+
+            int ref_size1 = best_ref1.size();
+            for (auto &t: best_ref1){
+                if (t.ref_start < 0) {
+                    continue;
+                }
+                abundance[t.ref_id] += float(read1_len) / float(ref_size1);
```
marcelm, line 929: Please don’t call this variable `ref_size`. This was quite confusing as I thought it represented the reference size.

src/aln.cpp
```diff
@@ -903,6 +910,73 @@ inline void get_best_map_location(
...
+
+        if (output_abundance){
+            for (auto &[score, n1, n2] : nam_pairs){
+                if ((n1.score + n2.score) == score_joint){
+                    best_ref1.push_back(n1);
+                    best_ref2.push_back(n2);
+                }else{
+                    break;
+                }
+            }
```
marcelm, line 922: This needs a comment describing what the point of this code is. It looked initially to me as if you were just re-finding `n1_joint_max` and `n2_joint_max`, but I guess you want to find all NAM pairs that have the same best score.

marcelm (commented):
Here are some more comments. Also, this needs documentation and tests.

marcelm (commented):
Just two more tiny comments. Also, I will probably rearrange the documentation a little bit (clarifying the different output modes a bit) after this has been merged, but it’s probably easiest if it’s not part of this PR. I’ll also add a changelog entry later.

src/main.cpp
```diff
@@ -105,7 +105,7 @@ InputBuffer get_input_buffer(const CommandLineOptions& opt) {
     }
 }
 
-void output_abundance(std::vector<double> abundances, References references){
+void output_abundance(const std::vector<double> abundances, const References references){
```
marcelm, line 108:
```suggestion
void output_abundance(const std::vector<double>& abundances, const References& references){
```

marcelm:
This looks good now! I think this deserves a slightly longer section in the documentation, but I’ll do that after merging.

PR #396, opened 2024-02-16
Introduce canonical read length 75

### LINKED ISSUE #395: Optimized parameters for read length 50 decrease accuracy for read length 75
Switching from (20, 16, -3, 2) to (18, 14, -2, 1) improved accuracy for read length 50, but had the unintended sideeffect of reducing it for read length 75, which is mapped to canonical read length 50 and therefore uses the same parameter settings.

The data to see this was already available in [this table](https://github.com/ksahlin/strobealign/issues/376#issue-2031100374), which says that (20, 16, -3, 2) is optimal for read length 75.

Do we need to add canonical read length 75?

marcelm:
Closes #395


ksahlin:
Approved

PR #403, opened 2024-03-04
Set default number of threads to 1

### LINKED ISSUE #401: Change default number of threads
The default number of worker threads is currently set to 3:
https://github.com/ksahlin/strobealign/blob/79cdd96df52fe3b5c9be88a1f11736b543e3d888/src/cmdline.hpp#L9

Since few CPUs have three cores, this is probably never the right number of threads to use. (I assume this was copied over from minimap2.)

I think there are three alternatives that make more sense (to me):
- Set the default to 4
- By default, use all available cores
- Set the default to 1

I suggest setting the default to 1, mainly because other tools do it this way and I would argue that this is the expected behavior.

Tools such as `bwa mem`, `bowtie2` and `make` use one thread by default. Typically, only tools explicitly meant to run things in parallel (`pigz`, `parallel`) use all available cores.

Even Snakemake expects a tool to use one thread by default. Currently, if strobealign is run without `-t` and the rule does not contain `threads: 3`, snakemake may incorrectly schedule too many parallel strobealign jobs.

One downside to reducing the default to 1 is that an inexperienced user may not know that they could get their results faster by using the `-t` option. To remedy this a bit, I suggest to make it clearer in the log output which number of threads strobealign is using.

marcelm:
Also, adjusted logging output to make it more obvious how many threads are used (and which mode). Example:

```
...
Total time indexing: 3.79 s
Processing paired-end reads in mapping-only mode using 1 worker thread
 Mapped         1.00 M reads @    11.02 us/read
Done!
```
Other example messages:
```
Processing single-end reads in abundance estimation mode using 4 worker threads
...
Processing single-end reads using 4 worker threads
```

Closes #401

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 # Strobealign Changelog
 
+## development version
+
+* #401: The default number of threads is now 1 instead of 3.
+
 ## v0.13.0 (2024-03-04)
 
 * #394: Added option `--aemb` (abundance estimation for metagenomic binning),
--- a/src/cmdline.cpp
+++ b/src/cmdline.cpp
@@ -19,7 +19,7 @@ CommandLineOptions parse_command_line_arguments(int argc, char **argv) {
     args::ActionFlag version(parser, "version", "Print version and exit", {"version"}, []() { throw Version(); });
 
     // Threading
-    args::ValueFlag<int> threads(parser, "INT", "Number of threads [3]", {'t', "threads"});
+    args::ValueFlag<int> threads(parser, "INT", "Number of threads [1]", {'t', "threads"});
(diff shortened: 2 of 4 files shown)


ksahlin:
Great! About your comment on making it 'easily interpretable' for less experienced users (https://github.com/ksahlin/strobealign/issues/401); what do you think about removing 'worker' in `1 worker thread` and just state `using 1 thread`? 

One may think that `worker threads` are not directly related to number of threads the program is using. For example, in the producer/consumer solution there is one producer and several consumers that sometimes might be confused with 'workers'.

marcelm:
> One may think that `worker threads` are not directly related to number of threads the program is using.

I somehow misremembered that the main thread does the I/O, so it was actually my intention to convey that there are more than $n$ threads, but checking the code, I see that that isn’t the case: The main thread only prints the progress indicator, so it shouldn’t be counted. I have changed the text as you suggested.

ksahlin:
Great, approved!

Issue #409, opened 2024-03-20
Invalid @SQ header


drtconway:
Hi strobealign.

We're checking out this very promising tool for our pipelines, but we have run into the following problem.

The HLA contigs in our HG38 reference (from the Broad, I think) have labels like

```
>HLA-A*01:01:01:01      HLA00001 3503 bp
```

which `strobealign` is turning in to the SAM header

```
@SQ     SN:HLA-A*01:01:01:01    HLA00001        LN:3503
```

which violates the SAM spec, and `samtools` complains about.

To reproduce, use the attached files, and run the command

```bash
strobealign ref.fa SRR11321731_one_read_{1,2}.fastq | samtools view -b -o test.bam -
```

Happy to discuss the "correct" behaviour. For comparison, `bwa` produces

```
@SQ     SN:HLA-A*01:01:01:01        LN:3503
```

[files.zip](https://github.com/ksahlin/strobealign/files/14661973/files.zip)

drtconway:
Looking at the data and the source code, the problem arises because although the code (`refs.cpp` line 48) cuts at the first space, the reference sequence name uses a tab as a separator.

A bit of googling confirms the lack of a single consistent standard or specification for the sequence names, so probably the best option is to work backwards from the [SAM format specification](https://samtools.github.io/hts-specs/SAMv1.pdf) which says the SN should match the regular expression `[:rname:∧*=][:rname:]*`.

In section 1.2.1 of the specification, it says reference sequence names should match

```
[0-9A-Za-z!#$%&+./:;?@^_|~-][0-9A-Za-z!#$%&*+./:;=?@^_|~-]*
```

which it uses to define `:rname:` as `[0-9A-Za-z!#$%&*+./:;=?@^_|~-]`.

So, I guess it would be idea if `strobealign` makes sure the names it prints in the `@SQ` header, and in the RNAME column conform to that.

drtconway:
Thanks for adding the tests. My bad! I'm happy.

PR #410, opened 2024-03-20
Add a fix for #409

drtconway:
This fix makes sure the reference sequence names produced conform to the SAM specification.

There is also code to check that the reference sequence does not contain duplicate names.

### DIFF
--- a/src/refs.cpp
+++ b/src/refs.cpp
@@ -1,5 +1,4 @@
 #include <vector>
-#include <set>
 #include <fstream>
 #include <sstream>
 #include <algorithm>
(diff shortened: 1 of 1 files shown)


src/refs.cpp
```diff
@@ -36,16 +89,31 @@ References references_from_stream(T& stream) {
...
         if (eof || (!line.empty() && line[0] == '>')) {
             if (seq.length() > 0) {
                 to_uppercase(seq);
                 sequences.push_back(seq);
                 names.push_back(name);
+                seen_names.insert(name);
             }
             if (!eof) {
-                name = line.substr(1, line.find(' ') - 1); // cut at first space
+                name = get_rname_prefix(line.begin() + 1, line.end());
```
marcelm, line 104: I’m wondering whether just extracting the longest prefix that matches the regex is the best thing to do here. If you have a header such as `>my_impörtant_sequence`, this would silently truncate the name to `my_imp`. I think splitting at tab or space and then erroring out if the part before the whitespace doesn’t match the regex would be a little bit less surprising.

src/refs.cpp
```diff
@@ -17,10 +18,62 @@ void to_uppercase(std::string& s) {
 
 namespace {
 
+/* Following the SAM specification, section 1.2.1,
+   return true if a given character is allowed in a reference sequence name. */
+bool is_rname_char(char c) {
```
marcelm, line 23: Would it perhaps be a bit nicer to use [`std::regex`](https://en.cppreference.com/w/cpp/regex)?

marcelm (commented):
Thanks! I had a short look, but will not be able to get back to you before next week. Please remind me if necessary!

src/refs.cpp
drtconway: Yes, it would be fine to do it that way. I hadn't checked which revision of the C++ standard you were coding against.

src/refs.cpp
drtconway: Sure! That would be reasonable also. It might also be good to still emit a warning or an error if the name you get does not satisfy the rules for SAM.

PR #411, opened 2024-03-21
Better fix for #409

drtconway:
I've reworked the change I proposed earlier to use `std::regex` which makes it neater.

It now splits on whitespace, then checks that the name is SAM-compatible.


marcelm:
Nice, thanks! This looks fine now.

Just a note: It is more customary to modify an existing PR until it is satisfactory because otherwise the previous discussion (as in #410) is lost. But no worries, this is not a problem here.

PR #425, opened 2024-05-14
Fix: Do not report failed alignments

marcelm:
* Refactor `Aligner::align` to return an `std::optional<Alignment>`. This way, we don’t have to rely on special vales for the attributes of an alignment to represent failed alignments.
* Fix the problem that we did not detect failed calls to ssw_align

Marked as draft because `extend_seed` and `rescue_align` don’t quite work, yet, with this change.

Closes #424

Issue #423, opened 2024-04-30
Optimized parameters for multi-context seeds


marcelm:
Branch: mcs-optimized-parameters (commit 7fe07b498e274567050cfe888bb91d1752e961c9).

The same command as in #407 was used:
```
./search.py -c ${commit} -x --accuracy-slack 0.1 --mapping-rate-slack 1 -r ${read_length}
```

## Suggested changes

Readl. | Before | Optimized | maponly SE | maponly PE | extalign SE | extalign PE
-|-|-|-|-|-|-
50 | (18, 14, -2, 1)  | (16, 12, -2, -1) | +0.1477 | +0.3609 | +0.2171 | +0.1236
75 | (20, 16, -3, 2)  | (22, 18, -2, -1) | +0.3006 | +0.2672 | +0.0232 | +0.1401
100 | (20, 16, -2, 2) | (20, 16, 0, 3) | +0.6108 | +0.3681 | +0.1630 | +0.1153
100 |                 | (23, 19, 0, 1) | +0.4422 | +0.3745 | +0.0729 | +0.1442
150 | (20, 16, 1, 7)  | (23, 19, 4, 7) | +0.1915 | +0.2323 | +0.0257 | +0.0748
150 |                 | (22, 18, 3, 7) | +0.2030 | +0.1779 | +0.0536 | +0.0679
200 | (22, 18, 2, 12) | (24, 20, 4, 12) | +0.1054 | +0.1024 | +0.0550 | +0.0386
300 | (22, 18, 2, 12) | (24, 20, 6, 13) | +0.2251 | +0.1039 | +0.0769 | +0.0368
500 | (23, 17, 2, 12) | still running

marcelm:
I’ve now started to run the same optimization as above on the new "Sim5" dataset (the above was done on the "Sim3" dataset, which has less variation).

Here are the results for the read lengths that have finished.

Readl. | Before | Optimized | maponly SE | maponly PE | extalign SE | extalign PE
-|-|-|-|-|-|-
50  | (16, 12, -2, 0) | (16, 12, -2, -1) | +0.2479 | +0.2580 | +0.1281 | +0.1493
75  | (20, 16, -3, -1)| (19, 15, -1, -1) | +0.2116 | +0.0701 | +0.0817 | +0.0682
100 | (16, 12,  1, 3) | (17, 13, 1, 4)  | +0.0891 | +0.0328 | -0.0007 | +0.0070
[...]

PR #426, opened 2024-05-22
Multi-context seeds plus fixes and optimized parameters

marcelm:
I realized only now that #388 is incomplete and that @ksahlin’s updates were made in a separate branch. We need a place to review this mcs-optimized-parameters branch and someplace where we have a press "Merge", so this is a separate PR that supersedes #388.


src/randstrobes.cpp
```diff
@@ -45,8 +45,11 @@ static inline syncmer_hash_t syncmer_smer_hash(uint64_t packed) {
     return xxh64(packed);
 }
 
-static inline randstrobe_hash_t randstrobe_hash(syncmer_hash_t hash1, syncmer_hash_t hash2) {
-    return hash1 + hash2;
+static inline randstrobe_hash_t randstrobe_hash(syncmer_hash_t hash1, syncmer_hash_t hash2, size_t aux_len) {
+    if (hash1 > hash2) {
+        std::swap(hash1, hash2);
+    }
+    return ((hash1 >> aux_len) << aux_len) ^ (hash2 >> (64 - aux_len));
```
marcelm, line 52: I think this would be clearer (and requires fewer operations) if a mask was used instead: `(hash1 & mask) | (hash2 & ~mask)`.

src/index.hpp
```diff
@@ -80,18 +82,66 @@ struct StrobemerIndex {
         return end();
     }
 
+    //Returns the first entry that matches the main hash
+    size_t partial_find(randstrobe_hash_t key) const {
```
marcelm, line 86: Code duplication: `find` is a special case of `partial_find` with `aux_len=0`.

src/index.hpp
```diff
@@ -80,18 +82,66 @@ struct StrobemerIndex {
...
+
     randstrobe_hash_t get_hash(bucket_index_t position) const {
         if (position < randstrobes.size()) {
             return randstrobes[position].hash;
         } else {
             return end();
         }
     }
+
+    randstrobe_hash_t get_partial_hash(bucket_index_t position) const {
```
marcelm, line 124: Should probably be named `get_main_hash`.

src/index.hpp
```diff
@@ -80,18 +82,66 @@ struct StrobemerIndex {
...
+        if (position < randstrobes.size()) {
+            return randstrobes[position].hash >> parameters.randstrobe.aux_len;
+        } else {
+            return end();
+        }
+    }
+
+    bool first_strobe_is_main(bucket_index_t position) const {
+        return randstrobes[position].main_is_first();
```
marcelm, line 133: Both methods should use the same name. `first_strobe_is_main` sounds more descriptive.

src/nam.cpp
```diff
@@ -171,11 +228,28 @@ std::pair<float, std::vector<Nam>> find_nams(
                 continue;
             }
             nr_good_hits++;
-            add_to_hits_per_ref(hits_per_ref[q.is_reverse], q.start, q.end, index, position);
+            add_to_hits_per_ref_full(hits_per_ref[q.is_reverse], q.start, q.end, index, position);
+        }
+        else {
+            PartialSeed ph = {q.hash >> aux_len, q.partial_start, q.is_reverse};
+            bool already_queried = std::find(partial_queried.begin(), partial_queried.end(), ph) != partial_queried.end();
+            if ( !already_queried ){
```
marcelm, line 236:
```
if (already_queried) {
    continue;
}
```

marcelm (commented):
Here are a couple of comments, mostly directed at myself.

src/randstrobes.cpp
ksahlin: As substitute for line 52, yes. I don't see how we would get around the necessary canonicity operation `if (hash1 > hash2)`  though. My benchmarks indicate than mcs (current implementation) are about 10% slower to create than main's seeds.

src/arguments.hpp
```diff
@@ -26,10 +26,11 @@ struct SeedingArguments {
             "results with non default values.", {'s'}}
         , bits{parser, "INT", "No. of top bits of hash to use as bucket indices (8-31)"
             "[determined from reference size]", {'b'}}
+        , aux_len{parser, "INT", "Number of bits to be selected from the second strobe hash [24]", {"aux-len"}}
```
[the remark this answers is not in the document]
ksahlin: Sure. However, that's an implementation specific thing which of the hashes should me main/primary/base (we need to decide on a consistent notation). I suggest something relating to the mcs method and not implementation. For example: "Number of bits selected for the secondary-level context of the seed."

src/nam.cpp
ksahlin: yes!

src/randstrobes.cpp
```diff
@@ -154,6 +158,15 @@ Randstrobe RandstrobeIterator::get(unsigned int strobe1_index) const {
     uint64_t min_val = std::numeric_limits<uint64_t>::max();
     Syncmer strobe2 = strobe1;
 
+    if (syncmers.size() < w_start) {
```
marcelm, line 161: Adding this only here breaks one of our tests:
https://github.com/ksahlin/strobealign/blob/f59bcb4efbf038d148184cb2a95dbcb581203f21/tests/tests.cpp#L87

ksahlin: I probably already mentioned this, but we can still make use of the last couple of base/main seeds in the case the window is already outside the read.

Not sure why I didn't add this for the references, but perhaps because the gain on genomes is very small (bigger gain with many small contigs), or that the indexing implementation is more complicated because it builds on parallelisation.

marcelm: Yes, you said so in the commit message. I made both implementations return the same result now. It is correct that the `count_randstrobes` function also had to be fixed.

marcelm:
Commit 21599d9f0d1bd26223d43ffc6403bb1e838d74bf "Possibly fix redundant alignment sites for symmetric multi context seeds." also breaks a test. After it, the `rescuable.43` read in the tiny test dataset is no longer rescuable. Shortened diff of read 1:

```diff
-rescuable.43   83  NC_001422.1  3137  60  120S14=1X4=1X9=1X4=1X4=1X9=1X4=1X4=1X4=1X9=1X19=1X4=1X4=1X4=1X4=1X4=1X4=1X9=1X4=1X9=1X4=1X9=1X4=1X4=1X4=1X      =       2955    -362    (sequence)       (qual)       NM:i:25 AS:i:120        RG:Z:1
+rescuable.43   69  NC_001422.1  2955  0  *  =  2955  0       (sequence)  (qual)  RG:Z:1
```

src/randstrobes.cpp
```diff
@@ -154,6 +159,15 @@ Randstrobe RandstrobeIterator::get(unsigned int strobe1_index) const {
...
+        return Randstrobe{
+            randstrobe_hash(strobe1.hash, 0, aux_len),
```
marcelm, line 164: Would it be ok to use `strobe1.hash` as the second parameter?
```suggestion
            randstrobe_hash(strobe1.hash, strobe1.hash, aux_len),
```
We already generate randstrobes that represent a syncmer paired with itself when the second syncmer is too far downstream; this would make both cases consistent with each other.

(Actually, the code would become much simpler also because the entire `if (syncmers.size() < w_start)` block could be deleted and we would rely on the for loop being executed zero times, just like when the second syncmer is further away than `max_position`.)

src/arguments.hpp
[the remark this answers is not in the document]
marcelm: Ok, I agree that’s an implementation detail. I suggest to leave the wording as it is and explain what this option is about (in the README). This is an advanced option that should usually not be changed anyway, so it is ok if it is somewhat cryptic.

src/randstrobes.cpp
marcelm: I’ve now made that change in commit 9c3768364b3f3e9ef981516b2c4fdac2d8a5bf53. It was necessary to avoid some spurious NAMs (which made `unmappable.41` mappable). See the commit message for an explanation.

src/randstrobes.cpp
marcelm: I’ve started converting `aux_len` to a mask `aux_mask`, but since it is a bit intrusive, I decided to wait with this change until the tests pass and perhaps until after this PR is merged.

src/randstrobes.cpp
ksahlin: Oh I see! Passing a 0 was a bug that I introduced! Good catch

marcelm:
> Commit [21599d9](https://github.com/ksahlin/strobealign/commit/21599d9f0d1bd26223d43ffc6403bb1e838d74bf) "Possibly fix redundant alignment sites for symmetric multi context seeds." also breaks a test. After it, the `rescuable.43` read in the tiny test dataset is no longer rescuable. [...]

I looked into this. The problem is that this heuristic check now returns false:
https://github.com/ksahlin/strobealign/blob/3a97f6b817824235a36ca8f9c5710ee533d3ad56/src/aln.cpp#L482

This happens because $k$ was increased from 22 to 25. With $k=22$, there were shared $k$-mers, but there are no longer any with $k=25$.

I think the easiest fix for now is to pass `-k 22` on the command-line when running that particular test, but maybe it also makes sense to reconsider how the heuristic works (perhaps it could depend less on $k$). But that would be something independent of this PR.

ksahlin:
Yes, sounds good to reconsider the heuristic or make it independent of k.

marcelm:
I’m not sure where, but @ksahlin has some measurements that show that the multi-context seeds branch is slower than `main`. I did some profiling to find out why that is the case. In short, I think the simple explanation seems to be that this branch simply generates more hits.

I used the Linux `perf` tool for profiling, which has a `perf diff` subcommand that can compare two profiles. Here is how the first couple of lines in its output look:
```
              +20.54%  strobealign-mcs       [.] find_nams
              +13.15%  strobealign-mcs       [.] SyncmerIterator::next
              +12.84%  strobealign-mcs       [.] (anonymous namespace)::merge_hits_into_nams
               +6.38%  strobealign-mcs       [.] sw_sse2_byte
```
That is, the multi-context seeds version from this PR spends 21% more time in the `find_nams` function, 13% more in the `SyncmerIterator::next` function etc.

The above was done with read length 150. I am going to check 50 bp as well. (Since one idea was to disable mcs at read lengths ~100bp and above.)

I then changed strobealign to keep track of a couple more statistics, in particular the number of hits and NAMs generated per read. This is the output.

Before (3a97f6b):
```
Number of reads: 1000000
Number of non-rescue hits: 25160733 total. Per read: 25.16
Number of rescue hits: 583028 total. Per rescue attempt: 23.07
Number of non-rescue NAMs: 27321159 total. Per read: 27.32
Number of rescue NAMs: 11457402 total. Per rescue attempt: 453.27
Total mapping sites tried: 1883792
Total calls to ssw: 184179
Inconsistent NAM ends: 1154
Tried NAM rescue: 25277
```

After (this PR):
```
Number of reads: 1000000
Number of non-rescue hits: 38371991 total. Per read: 38.37
Number of rescue hits: 843076 total. Per rescue attempt: 30.16
Number of non-rescue NAMs: 39108199 total. Per read: 39.11
Number of rescue NAMs: 14522183 total. Per rescue attempt: 519.54
Total mapping sites tried: 1898603
Total calls to ssw: 193726
Inconsistent NAM ends: 1205
Tried NAM rescue: 27952
```

So is the explanation that mcs is slower because it generates more hits and NAMs?

ksahlin:
Pasted below are the plots you're referring too. I was interested mainly in reducing, or at least finding the cause for, the overhead for longer reads 200 and above (see figs). This may not be so relevant if we disable the mcs for reads over 125bp or so.

[time_plot_SE.pdf](https://github.com/user-attachments/files/16997398/time_plot.pdf)

[time_plot_PE.pdf](https://github.com/user-attachments/files/16997401/time_plot.pdf)

marcelm:
I tested all commits in this branch to see which ones contribute to the slowdown (using the sim3 datasetand drosophila-200 reads, single-end, mapping only). This table shows the commits at which runtimes change:

Commit  | Time (s) | Overall slowdown | Description
-|-|-|-
f59bcb4 | 12.9 | 0% | Merge pull request #425 from ksahlin/optional
379a1fc | 13.9 | 8% | (Added mcs, lumped together with earlier commits)
3902f41 | 14.7 | 14% | Prevent looking up and, thus, adding the same partial hits twice
caea5c2 | 15.4 | 19% | Updated parameters according to optimization in issue #423
d5125b6 | 16.8 | 30% | Make RandstrobeIterator/-Generator yield the same results
9c37683 | 16.5 | 28% | Ensure unmappable.41 really is unmappable

* Commit 3902f41 adds a linear search in a `vector`; this can probably be optimized a bit (using a set).
* I don’t understand, yet, why d5125b6 slows things down so much. It should only add a couple of strobemers at the end of each contig (syncmer paired with itself). That is, it only affects the reference. There was a problem here that the second hash value was set to zero (see discussion above), but that was fixed in commit 9c37683. However, runtime only goes down a bit with that commit.

marcelm:
Here are also runtimes for disabling multi-context seeds (by just commenting out the appropriate `else` in `find_nams` or `find_nams_rescue`:

* 13.1s when disabling mcs only in `find_nams`
* 13.0s when disabling mcs also in `find_nams_rescue`

Since runtime in `main` was 12.9s, this means that all the other changes done in this branch affect runtime very little. (For sim3-drosophila-200, that is.)

ksahlin:
Great detective work! 

I am not sure I have any great input on the above yet. As you mention, speed increase from  https://github.com/ksahlin/strobealign/commit/d5125b68e915804225b3dfff8d6a042e8ebeb061 is very strange indeed.

> Commit https://github.com/ksahlin/strobealign/commit/3902f4171473f317f1161926e7837108d5ae52cf adds a linear search in a vector; this can probably be optimized a bit (using a set).

I wrote that as a TODO in the code, but my first guess was that small vectors (up to 100 elements or so) are faster to search and maintain than small sets. but maybe worth to try.

ksahlin:
We do `bool first_strobe_is_main = strobe1.hash < strobe2.hash;` in `make_randstrobe`. Then 

```
    if (hash1 > hash2) {
        std::swap(hash1, hash2);
    }
```

In the cases we set `Syncmer strobe2 = strobe1;` (several times in both fw and rc ends in each query), it becomes an unnecessary swap.

Also, I don't think we need `assert(i <= w_max);` and `assert(i < syncmers.size());` in the `RandstrobeIterator` and `RandstrobeGenerator::next`.

Could one of these be responsible for the slowdown in https://github.com/ksahlin/strobealign/commit/d5125b68e915804225b3dfff8d6a042e8ebeb061 ?

ksahlin:
I think my first statement is also not correct. It only results in that strobe2 is main, which is not harmful since they  are set to be the same at start.. I will blame my lame attempt on that I am both on vacation and sick -- trying to do something meaningful from an hotel room.. I'll let you do the detective work instead :)

marcelm:
Oh no, sorry to hear! Yeah, the assertions are ignored when compiling in Release mode. And I think the swap is very cheap; that should be something like two machine instructions. I’ll investigate further.

marcelm:
One of the remaining to-do items here is to enable multi-context seeds only when the read length is below a certain threshold. I just pushed commit 47f92575467569ecf75e13866f81f291b947a85a, which makes it possible to enable or disable mcs dynamically by setting a boolean parameter, but for the moment, I hard-coded this parameter to `false`. That is, with this commit, all the other changes in this PR are still enabled, but actually using multi-context seeds at mapping time is disabled.

I ran a full evaluation comparing `main`, multi-context seeds and this most recent commit, please see the most relevant plots here:
* SE [time.pdf](https://github.com/user-attachments/files/17190788/time.pdf)
* SE [accuracy.pdf](https://github.com/user-attachments/files/17190790/accuracy.pdf)
* PE [time.pdf](https://github.com/user-attachments/files/17192444/time.pdf)
* PE [accuracy.pdf](https://github.com/user-attachments/files/17192446/accuracy.pdf)

We can use this to decide which criteria to use for enabling mcs. But I’m having a hard time right now to make such a decision because the slowdown feels relatively large for some datasets. I’ll try to improve the runtime a bit, maybe that makes it easier.

ksahlin:
Nice!

Would the difference between green and pink mostly be from new seeding parameters?

From your plots it seems favourable to use mcs for read lengths 100 and below. Actually, I think I mentioned around 110-125 somewhere earlier. For 50-100bp, the mapping accuracy is substantially better at reasonable (sometimes no) cost in runtime. Drosophila seems most affected by mapping slowdown for these short read lengths (relative terms).

Assuming there would be a significant improvement in runtime for mcs (e.g., lowering some of the increases we observed in your table), I'd say we would want to use it up to lengths of 200. Particularly, it seems mcs can help the mapping only and, thus, the `--aemb` setting quite a bit.

marcelm:
I’m finding it difficult to work on some aspects of strobealign at the moment because many changes would actually be in conflict with the multi-context seeds PR.

Also, as we saw in the table above, this PR contains multipe things that change strobealign’s behavior. I think it would be good to tease them apart into separate PRs where each one is much smaller than this PR.

I’m mainly thinking:
1. Change the way randstrobes are hashed (that is, introduce `aux_len`)
2. Use multi-context seeds when looking up randstrobes
3. Add syncmers at the ends of reads
4. Switch to optimized parameters

marcelm:
> Would the difference between green and pink mostly be from new seeding parameters?

I was wondering that as well, so I ran an evaluation for this PR but without parameter optimization (mcs lookups disabled, parameters from main). It’s "mcs-nooptim" (the olive lines) in these plots (some read lengths missing so I wouldn’t have to wait that long):

* SE [time.pdf](https://github.com/user-attachments/files/17213027/time.pdf)
* SE [accuracy.pdf](https://github.com/user-attachments/files/17213029/accuracy.pdf)

Accuracy is identical to main. Runtime is is consistently slightly higher than main. Due to the observation in #447 that only changing the hash function actually reduces runtime slightly, I think this is due to the extra syncmers at the end of reads.

ksahlin:
> Also, as we saw in the table above, this PR contains multipe things that change strobealign’s behavior. I think it would be good to tease them apart into separate PRs where each one is much smaller than this PR.

Very good suggestion! I agree.

> Due to the observation in https://github.com/ksahlin/strobealign/pull/447 that only changing the hash function actually reduces runtime slightly, I think this is due to the extra syncmers at the end of reads.

Ok, makes sense!

src/index.hpp
marcelm: Fixed in PR #450

marcelm:
Everything in this PR has now been split out into other, smaller PRs.

The discussion about improving mcs speed is still ongoing, so I’ll leave this PR open for the moment.

>> Commit https://github.com/ksahlin/strobealign/commit/3902f4171473f317f1161926e7837108d5ae52cf adds a linear search in a vector; this can probably be optimized a bit (using a set).
> 
> I wrote that as a TODO in the code, but my first guess was that small vectors (up to 100 elements or so) are faster to search and maintain than small sets. but maybe worth a try.

I tried a couple of things to make multi-context seeds faster:

- Use `std::unordered_set` and `robin_hood::unordered_set` for `partial_queried`: This is actually slower.
- Do not call `partial_queried.reserve(10)` (or use different values other than 10): This makes a small difference, which is why I disabled the call to reserve() entirely if multi-context seeds are disabled. If they are enabled, reserving a couple of elements seems to be a bit beneficial.
- In the `query_randstrobes` vector, we first get all forward randstrobes and then all reverse randstrobes. I tried to truncate the `partial_queried` vector when we see the first reverse randstrobe because we do not need to compare them to the forward ones. This does not make a difference.

Also, when looking up a full hash, the interval that we find is a subinterval of the interval for the partial hash. We currently look up the subinterval first and then the larger interval encompassing it, which requires that we do the full lookup again. I tried to swap this around, that is, to always look up the partial hash first. The advantage should be:
- If we don’t find the partial hash, we can stop because we won’t find the full hash either.
- In the majority of the cases, the first randstrobe that matches the partial hash also matches the full hash. Then we don’t need to do a second lookup.
- If the position we found does not match the full hash, we could just search forward from that position and would not have to re-lookup.
Unfortunately, this wasn’t actually faster. I presume this is because everything is in the cache already when we do the second lookup of the partial hash. I’m not quite ready to give up on this because I find it a bit cleanerconceptually, so I’ll invest a little bit of time to ensure I haven’t overlooked anything.

marcelm:
I have also tried to vary the `MAX_LINEAR_SEARCH` parameter (currently set to 4). Interestingly, it makes no measurable difference in total runtime whether I set it to 0 or 10000 (at least on Drosophila).

ksahlin:
> Also, when looking up a full hash, the interval that we find is a subinterval of the interval for the partial hash. We currently look up the subinterval first and then the larger interval encompassing it, which requires that we do the full lookup again. I tried to swap this around, that is, to always look up the partial hash first. 

I like this idea(!) and also find it conceptually cleaner. However, I don't think drosophila is a good test for this as most partial seeds are already unique. I would benchmark this on CHM13.

> I have also tried to vary the MAX_LINEAR_SEARCH parameter (currently set to 4). Interestingly, it makes no measurable difference in total runtime whether I set it to 0 or 10000 (at least on Drosophila).

Makes sense to try! But I would also benchmark this on CHM13.

marcelm:
Regarding varying `MAX_LINEAR_SEARCH`: I tried it on CHM13 and set it to various powers of two. There’s no discernible difference in runtime when set to 0, 4 or 256. It gets a little bit (1%) slower at 512 and then it gets worse with higher values. So I think we’re fine with leaving it at 4.

marcelm:
Documenting another failed attempt at improving speed: In commit 6a5eca7 I tried to make it so that the `partial_queried` vector is not re-created every time the `find_nams` function is run. Instead, there’s just one vector that gets cleared. This avoids the memory allocation (which I thought was somewhat expensive because the call to `reserve()` takes some time). But alas, there’s no difference in speed.

PR #427, opened 2024-06-05
Add a debugging compilation option for showing list of NAMs

marcelm:
If `cmake` is run with `-DTRACE=ON`, then some `#ifdef`s become active that make strobealign print the list of found NAMs for each query.

I find myself adding this type of output quite often for debugging purposes, and think it would be nice to have this as part of the code, disabled by default, but easily enabled.

`CHUNK_SIZE` is forced to 1 when tracing is on to ensure that SAM/PAF output comes right after the corresponding tracing output. (Otherwise you get the tracing output for all queries within one chunk follewd by all the SAM/PAF output for all queries in the chunk.)


ksahlin:
approved

PR #428, opened 2024-06-05
Make Python bindings work with nanobind 2+

marcelm:
Nanobind changelog
(https://nanobind.readthedocs.io/en/latest/changelog.html#version-2-0-0-may-23-2024):

> The nb::enum_<T>() binding declaration is now a wrapper that creates either
> a enum.Enum or enum.IntEnum-derived type. Previously, nanobind relied on a
> custom enumeration base class that was a frequent source of friction for
> users.
>
> This change may break code that casts entries to integers, which now only
> works for arithmetic (enum.IntEnum-derived) enumerations.


marcelm:
Should be uncontroversial, merging this now.

PR #439, opened 2024-09-13
Do not compile unused.cpp

marcelm:
I would like to suggest no longer compiling the `unused.cpp` file by default for a couple of reasons:
- Because it includes some header files, it takes extra effort to make it compilable if those header files change.
- While building, many warning messages are printed just for unused.cpp while for strobealign itself almost all warnings have been eliminated. This makes it harder to see whether a change introduces a new warning.
- It takes extra time to compile (just a couple of seconds each time, but it adds up).

This PR removes the file as a target from `CMakeLists.txt` so that it no longer gets compiled. I haven’t deleted the file itself for now.


ksahlin:
Oh, I didn't even know we compiled it. Approved!

PR #446, opened 2024-10-01
Rename Hit struct to Match

marcelm:
We use the term "hit" both for
- a query strobemer successfully looked up in the index
- a locus on the query paired with a locus on the reference (this type of hit is merged into NAMs)

I think choosing the name for the latter to be "match" makes sense because the M in NAM stands for "match". That is, we can then say that we merge multiple overlapping matches into non-overlapping, approximate matches (=NAMs).

The flow would then be: hit -> match -> NAM.

Marking as draft because this should only be merged after #426 as it will otherwise cause conflicts.

### DIFF
--- a/src/aln.cpp
+++ b/src/aln.cpp
@@ -118,7 +118,7 @@ inline void align_single(
     best_alignment.is_unaligned = true;
 
     for (auto &nam : nams) {
-        float score_dropoff = (float) nam.n_hits / n_max.n_hits;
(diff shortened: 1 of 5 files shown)


ksahlin:
Great, approved for merge after https://github.com/ksahlin/strobealign/pull/426.

marcelm:
After merging most of #426, this PR no longer conflicts, so merging.

PR #448, opened 2024-10-02
Implement multi-context seed lookups (but leave them disabled)

marcelm:
This adds everything necessary for using multi-context seeds during lookups, but also adds a configuration option that is set to `false` so that multi-context seeds are never actually looked up.

The idea is that merging this PR should have not impact at all, but prepare everything for enabling mcs lookups. However, there was initially a slowdown compared to main due to the following:

- When `find_nams` calls `merge_hits_into_nams`, it previously had the `sort` parameter set to `false`, but with mcs, it is necessary to sort the hits because they may not be generated in ascending order. So this was set to `true` unconditionally and causes a couple of percent of slowdown. This was remedied in 07954dab1ece82c93e77f7e433a664de5bd17978 (only sort if `use_mcs` is true).
- Calling `partial_queried.reserve(10)` causes a slowdown. In 37bf55f4cf3b3d5bf42d32b7a5f576718803d136, this was changed to call it only when `use_mcs` is true.

Also, there are some minor changes in `merge_hits_into_nams` that change the results a tiny bit. This made it necessary to update the baseline commit.


ksahlin:
Good that you found the explanation for those extra percents! Approved.

marcelm:
I added one last commit to this PR that adds a `--mcs` command-line option. That way, we can enable multi-context seeds for experiments without having to recompile. My thinking at the moment is that we remove that option again before the next release, but maybe it can remain or become a `--no-mcs` option or so.

marcelm:
The baseline commit hash for tests was wrong; I pushed a fix directly to main.

PR #449, opened 2024-10-03
Prepare release: Bump version to 0.14.0

marcelm:
I think this is a good time for a release, with the ISA-L improvements and just after we switched to multi-context ~~seeds~~ hashes (which gives a slight speedup), but before actually using them.

(Edited)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,6 @@
 # Strobealign Changelog
 
-## development version
+## v0.14.0 (2024-10-03)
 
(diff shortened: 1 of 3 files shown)


ksahlin:
I agree! approved

marcelm:
Done, including the Bioconda package:
- https://github.com/ksahlin/strobealign/releases/tag/v0.14.0
- https://github.com/bioconda/bioconda-recipes/pull/51125

PR #450, opened 2024-10-07
Multi-context seeds refactoring and docs

marcelm:
Mainly, this removes one of the currently duplicated `StrobemerIndex::find*()` functions.

Instead, have a common function that has a parameter which tells it how many bit of the hash to ignore (0 for finding a full hash and aux_len for a partial one).

Same for `get_count()`.


ksahlin:
Great, approved!

PR #451, opened 2024-10-07
Add a test for multi-context seeds

marcelm:
That is, run `strobealign` with `--mcs` as one of the tests.


marcelm:
This should be uncontroversial.

PR #452, opened 2024-10-07
Add syncmers at ends of reads

marcelm:
Split out from #426.

To Do

- [ ] Maybe add these syncmers only when mcs is enabled
- [ ] Fix tests

### DIFF
--- a/src/index.cpp
+++ b/src/index.cpp
@@ -27,18 +27,15 @@ static const uint32_t STI_FILE_FORMAT_VERSION = 3;
 namespace {
 
 uint64_t count_randstrobes(const std::string& seq, const IndexParameters& parameters) {
+    // To reduce runtime, this function actually counts *syncmers* and
+    // assumes that one randstrobe is generated per syncmer
     uint64_t n_syncmers = 0;
     SyncmerIterator syncmer_iterator(seq, parameters.syncmer);
     Syncmer syncmer;
     while (!(syncmer = syncmer_iterator.next()).is_end()) {
         n_syncmers++;
     }
-    // The last w_min syncmers do not result in a randstrobe
-    if (n_syncmers < parameters.randstrobe.w_min) {
-        return 0;
-    } else {
-        return n_syncmers - parameters.randstrobe.w_min;
-    }
+    return n_syncmers;
 }
 
 std::vector<uint64_t> count_all_randstrobes(const References& references, const IndexParameters& parameters, size_t n_threads) {
--- a/src/randstrobes.cpp
+++ b/src/randstrobes.cpp
@@ -202,10 +202,21 @@ Randstrobe RandstrobeGenerator::next() {
         }
         syncmers.push_back(syncmer);
     }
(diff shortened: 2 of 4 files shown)


marcelm:
In https://github.com/ksahlin/strobealign/pull/426#issuecomment-2379082142, I measured that the commit "Make RandstrobeIterator/-Generator yield the same results" caused a significant slowdown in strobealign when multi-context seeds are enabled. I’ve tried to analyze that commit only, but doing so doesn’t make sense because it contains the buggy hashing (where the second hash was always set to zero), which was fixed in a subsequent commit. This PR here contains that particular commit and the fix, so I think it is better to just look at this entire PR and discuss the slowdown here.

First, the slowdown when multi-context seeds are *disabled* is relatively small, about 2%, which is presumably just because more randstrobes are generated per read. Diff (main vs this PR):

fruit fly 100 bp reads:
```
-Number of randstrobes: 56477476 total. Per read: 28.24
+Number of randstrobes: 64477476 total. Per read: 32.24
```
fruit fly 200 bp reads:
```
-Number of randstrobes: 59178572 total. Per read: 59.18
+Number of randstrobes: 71178572 total. Per read: 71.18
```

I don’t think there should be any gains from generating these extra randstrobes when we are not using multi-context seeds, so we may want to not generate them in that case.

Second, the slowdown when multi-context seeds are *enabled* is about 10%. Profiling shows that strobealign spends 36% more in the `find_nams()` function and 22% more in `merge_matches_into_nams()`. This is a bit more than what I had expected given that there aren’t 36% more randstrobes.

I made another observation (which may be obvious to Igor and you), but which I wanted to write down. Regarding these extra randstrobes at the end of queries that this PR adds.

- These randstrobes are usually not found in full in the index because the strobes are closer together than w_min (wich was used when indexing).
- That means the full lookup is theoretically unnecessary. They will always trigger a partial lookup.
- The partial lookup only succeeds when the strobe that we see in the read was chosen as *main* during indexing, so it fails half the time.

marcelm:
I’ve rebased this PR on top of main (v0.16.0) and ran some measurements on sim3 and sim5. The differences in accuracy are very small and average to close to +-0. The exception is CHM13, read length 50 in mapping mode: On sim3, accuracy improves by 0.2 percentage points and on sim5, it improves by 0.4 percentage points. All other changes are smaller than 0.1 percentage points. Given that this PR also makes mapping slower by ca. 5%, I’d tend towards not making this change as-is.

What I could still try:
- Run the measurements on the new sim6 dataset
- Run strobealign with `--mcs` when doing measurements
- Look up the extra syncmers only if no regular randstrobes were found

ksahlin:
Tough call. It will probably matter more for split-read mapping and be noticed more on SIM6. 

> The exception is CHM13, read length 50 in mapping mode

Single-end reads I presume?

> Look up the extra syncmers only if no regular randstrobes were found

Do you mean only adding them only if we enter the partial strobe rescue (MCS-R) step? That sounds appealing, if there is a clean way in the code to to do this.

PR #453, opened 2024-10-07
Use optimized parameters according to optimization in issue #423

marcelm:
The last PR to be split out from #426.

If we activate multi-context seeds only for certain read lengths, should we also only use optimized parameters for those read lengths?

To Do:

- [ ] Rebase once #452 has been merged.


ksahlin:
> If we activate multi-context seeds only for certain read lengths, should we also only use optimized parameters for those read lengths?

Good point

PR #457, opened 2024-11-19
Fix shuffle_top_nams

marcelm:
We would not shuffle the NAMs if all found NAMs had the same score.

I found this a couple of weeks ago but noticed now I never sent this as a PR.


marcelm:
This doesn’t improve mapping accuracy, BTW.

ksahlin:
Nice! Approved

PR #455, opened 2024-10-22
Allow references with up to 2**31 contigs 

marcelm:
~~Note, this is not quite the layout suggested in https://github.com/ksahlin/strobealign/issues/421#issuecomment-2429388998. I implemented it this way before that comment, but will adjust it.~~

This changes the RefRandstrobe struct in the following way.

Previous layout:
- 40 bits base hash
- 24 bits auxiliary hash
- 32 bits position
- 23 bits reference/contig index
- 8 bits strobe2 offset
- 1 bit first_strobe_is_main flag

New layout:
- 38 bits base hash
- 17 bits auxiliary hash
- 8 bits strobe2 offset
- 1 bit first_strobe_is_main flag
- 32 bits position
- 32 bits reference/contig index

Closes #421

*Edited to reflect actual new layout*

PR #458, opened 2024-11-20
Let --aux-len work correctly

marcelm:
With PR #455, `--aux-len N` changed its meaning and needs to be interpreted as "use N minus 9 bits of auxiliary hash" (because 9 bits are used for the mcs flag and strobe2 offset), so it no longer works as intended. This PR fixes that.

This includes also some refactoring. In particular, we no longer use aux_len directly in the `randstrobe_hash()` function, but a "main_hash_mask". Note that the way the bits are combined also changes a little bit (which bits from hash2 are used is different).


marcelm:
Added a check to ensure that aux_len is reasonable (at least 6 and such that B is within the main hash).

marcelm:
Merging this now.

PR #459, opened 2024-11-25
Report number of generated randstrobes

marcelm:
With `-v`, this reports the number of generated randstrobes in total and per read.


marcelm:
Should be uncontroversial, merging.

PR #461, opened 2024-11-26
Experiment: Fast path for exact matches

marcelm:
Thinking about ways to alleviate the multi-context seeds slowdown, I tried to add a "fast path" for exact matches (this has nothing to do with mcs, but if we add an unrelated optimization at the same time, it feels less bad): The first randstrobe from the query is looked up in the index, and if it results in a unique hit (only one entry for that hash in the index), the query and reference are compared and if they match exactly, the match is output and everything else is skipped (that is, looking up other hits in the index, merging them into NAMs, ungapped alignments for the top NAMs).

This is a just hack for the moment (with some code duplication) that is also incomplete as it does not check the reverse-complemented sequence.

I tested this on sim5-drosophila-100 first, and there was no improvement (although ~30% of all CIGARs are `100=`), but it does help a little bit with the sim3-drosophila-100 library, which appears to gets 3% faster.

I’ll need to do some more measurements.

### DIFF
--- a/src/aln.cpp
+++ b/src/aln.cpp
@@ -1179,6 +1179,43 @@ void align_or_map_single(
     auto query_randstrobes = randstrobes_query(record.seq, index_parameters);
     statistics.tot_construct_strobemers += strobe_timer.duration();
 
+    // Try an optimization if the first randstrobe is unique
+    if (query_randstrobes.size() > 0) {
+        auto qr = query_randstrobes.front();
+
+        size_t pos = index.find_full(qr.hash);
+        if (pos != index.end() && index.is_unique(pos)) {
+            size_t ref_pos = index.get_strobe1_position(pos);
+            if (ref_pos >= qr.start) {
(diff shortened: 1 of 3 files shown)


ksahlin:
Copy comment on an individual commit to be in the PR discussion instead: 

_This optimization could be dangerous as randstrobes (similar to any other approximate seeds) are not guaranteed to match the smallest edit distance, so _one_ unique hit could actually be suboptimal (see fig below). Happens when mutations permute the hash values._ 

_I'm not sure how common this is in practice, but I am sure it happens occasionally for large datasets - but is typically solved by looking ate some more seeds._

_Maybe to solve this in practice is to look at two seeds in instead - because the probability that this happens (p) would be p^2 for two seeds under a good hash function._ 

Image

ksahlin:
I see, your commit also includes a hamming/extension alignment and requires a single exact match? Then I guess there is no real limitation introduced. But still good to keep my above comment in mind.

PR #465, opened 2024-12-04
Test: SyncmerIterator returns canonical syncmers

marcelm:
See also #463


marcelm:
Added this to PR #464 instead.

PR #466, opened 2024-12-04
Experiment: Use a bit in RefRandstrobe to indicate whether it is filtered

marcelm:
Profiling suggested that the `Index::is_filtered()` call is a bit slow. It checks whether a randstrobe occurs more often than `filter_cutoff` by accessing `randstrobes[i]` and `randstrobes[i + filter_cutoff]` and comparing the hashes.

The slowness could come from two cache misses because two quite far apart memory locations are read. To get rid of the second access, the idea is to use one bit within `RefRandstrobe` to store whether the item is filtered.

Somewhat unexpectedly, this does not improve speed. It does reduce cache misses according to `perf stat -d`, but this does not translate to a shorter runtime.


marcelm:
After a couple of measurements on a different (10 years younger) machine, I *can* measure a difference - this PR makes mapping-only mode about 2% faster. (This comes at the expense of one less bit available for the hash, but this has very little impact.)

ksahlin:
Great! Don't we anyway have B top bits available to store other things because of our prefix vector?  This depends of course on that the bit is added after the sorted vector has been produced.

marcelm:
> Great! Don't we anyway have B top bits available to store other things because of our prefix vector? This depends of course on that the bit is added after the sorted vector has been produced.

Right, good point! I have the impression the filter bit would better fit in those upper bits anyway. Let me update the PR later.

marcelm:
I’m going to close this PR as well because it’s outdated. I have the idea still in my separate to-do list and it may be worth pursuing at some other time, but given that this does not improve speed so much and that I’m currently going in the opposite direction of reducing the index size, this is not a high priority anymore.

PR #467, opened 2024-12-10
ENH Make reference reading faster

luispedro:
This was not a problem for databases with few large sequences, but when using many sequences, small inefficiencies in header parsing add up.

There is a slight loss of features here as duplicated names will only be checked later and it is not possible to report the line number.

Parsing the greengenes database (2022.10.seqs.fna, with 21,090,001 sequences) goes from 19-20s to 7s.

### DIFF
--- a/src/refs.cpp
+++ b/src/refs.cpp
@@ -3,7 +3,7 @@
 #include <sstream>
 #include <algorithm>
 #include <regex>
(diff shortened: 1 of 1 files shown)


marcelm:
Thanks, this looks good IMO!

PR #471, opened 2024-12-16
Be less strict when verifying contig identifiers

### LINKED ISSUE #468: Consider relaxing the check whether contig identifiers match the SAM regex
@luispedro suggested by e-mail that strobealign is too strict when it checks the identifiers of sequences in the reference FASTA. Other tools such as BWA, minimap2 and Bowtie2 are less strict. 
The current check was contributed by @drtconway in PR #410 and fails parsing when an identifier doesn’t match the regex in the SAM specification. Somewhat suprisingly, even SAMtools itself does not complain when these "disallowed" characters occur in identifiers. For example, we currently reject reference identifiers with the `<` or `>` character, but SAMtools doesn’t complain.

#410 was mainly intended to fix the problem that strobealign did not recognize a tab character as separator between sequence identifier and comment (#409). That part of the fix needs to stay of course, but I agree we can relax the strict regex check.

marcelm:
Closes #468

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 # Strobealign Changelog
 
+## development version
+
+* #468: Be less strict when checking reference sequence names.
+
 ## v0.15.0 (2024-12-13)
 
 * #388 and #426: Increase accuracy and mapping rate for reads shorter than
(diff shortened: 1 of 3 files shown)


ksahlin:
approved

PR #473, opened 2024-12-17
Skip alignments that are actually not aligned

marcelm:
`extend_seed` may return alignments with `is_unaligned` set to `true` if `Aligner::align` fails to align a sequence.

If all NAMs result in such "unaligned alignments", there’s no best alignment, and we need to ensure that we output an unmapped read.


ksahlin:
Good catch, approved!

PR #474, opened 2024-12-19
Non-canonical syncmers and randstrobes

marcelm:
This changes to non-canonical randstrobes *and* to non-canonical syncmers. This is a bit incomplete (only single-end works), needs some cleanup, and some improvements because it also reduces the accuracy. I just want to have a single place where this change can be discussed and where I can keep a to-do list.

## To Do

- [ ] Make paired-end mapping work
- [ ] Remove `reverse_nam_if_needed`
- [ ] Remove `first_strobe_is_main` (it’s always true)
- [ ] To alleviate the problem that filtering now works differently for forward and reverse sequences, we could perhaps somehow also look at the reverse complemented sequence when checking whether a randstrobe is filtered. That is, let the filter work as with canonical syncmers/randstrobes. Not sure how.

### DIFF
--- a/src/aln.cpp
+++ b/src/aln.cpp
@@ -1010,6 +1010,7 @@ void align_or_map_paired(
         const auto& record = is_revcomp == 0 ? record1 : record2;
 
         Timer strobe_timer;
+        // FIXME forward only at the moment
         auto query_randstrobes = randstrobes_query(record.seq, index_parameters);
         statistics.n_randstrobes += query_randstrobes.size();
         statistics.tot_construct_strobemers += strobe_timer.duration();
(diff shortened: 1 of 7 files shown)


ksahlin:
Good idea with this PR.

About the fourth point:

Very vague but; if a read has more or equal number of distinct hits found in the index in the masked direction than as the non-masked direction it could be flagged as suspicious. For suspicious reads we could do some extra work (eg collecting some matches in both directions) to assess which direction is best?  

Are the seeds sorted also w.r.t chr ID and pos in the index, or how does the pdq sort work again? If they were we could attempt something like storing the first hit of each masked hit in the hope that they well map to the same beginning of a repetitive region (this is not very robust though)

PR #472, opened 2024-12-17
Always pick first syncmer as main

marcelm:
As discussed: Always pick the first syncmer as main.

~~As suggested in an e-mail by @ksahlin, this changes the way the base syncmer is chosen when multi-context seeds are enabled. The rule is: Pick the syncmer with odd hash value. If there’s a tie, use the one with the smaller hash value (as before).~~

~~I also tested two other rules, but these would make randstrobes non-canonical:~~
- ~~XOR the two hashes together and pick the first if the resulting value is odd~~
- ~~Always pick the first syncmer~~

PR #476, opened 2025-01-15
Rescue using partial hits, even in non-MCS mode

marcelm:
Credit for this idea goes to @Itolstoganov.

PR #472 changes the index so that the first syncmer is always chosen as the base. This improves accuracy in MCS, but reduces mapping rate (and thus accuracy) in non-MCS mode. Apparently, some reads with many errors no longer get any hits at all.

However, we can easily detect when a query doesn’t get any hits and then - even if MCS isn’t explicitly enabled - just use the method we already have of looking up partial seeds, as a different kind of MCS rescue.

To be clear, only using partial seeds when no regular hits can be found was an idea by @Itolstoganov.

This actually works *much better* than I would have expected: Both accuracy and speed get improve!

Details below.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,14 +2,20 @@
 
 ## development version
 
+* #476: Significantly improve speed and accuracy by enabling by default a new
+  variant of multi-context seeds: When no regular seeds - which consist
+  of two strobes - can be found for the entire query, strobealign attempts to find
+  single-strobe ("partial") seeds.
+  The `--mcs` option is still available for now. It is a bit slower, but
+  slightly more accurate.
 * #468: Be less strict when checking reference sequence names.
 
 ## v0.15.0 (2024-12-13)
 
 * #388 and #426: Increase accuracy and mapping rate for reads shorter than
   about 200 bp by introducing multi-context seeds.
-  Previously, seeds always consisted of two k-mers and would only be found if
-  both occur in query and reference.
+  Previously, seeds always consisted of two k-mers ("strobes") and would only
+  be found if both occur in query and reference.
   With this change, strobealign falls back to looking up just one of the k-mers
   when appropriate.
   This feature is currently *experimental* and only enabled when using the
--- a/README.md
+++ b/README.md
@@ -223,6 +223,29 @@ actual mapping:
 - Index files are about four times as large as the reference.
 
 
+## Explanation
+
+### Multi-context seeds
+
+Strobealign uses randstrobes as seeds, which in our case consist of two k-mers
+("strobes") that are somewhat close to each other. When a seed is looked up
+in the index, it is only found if both strobes match. By changing the way in
+which the index is stored in v0.15.0, it became  possible to support
+*multi-context seeds*. With those changes, strobealign falls back to looking
+up only one of the strobes (a "partial seed") if the full seed cannot be found.
+This results in better mapping rate and accuracy for read lengths of up to
+about 200 nt.
+
+Usage of multi-context seeds is enabled by default in strobealign since v0.16.0.
+The strategy is to first search for all full seeds of the query and fall back to
+partial seeds if *no* seeds could be found.
+
+A slightly more accurate, but slower mode of using multi-context seeds is
+available by using option `--mcs`: With it, the strategy is changed to a
+fallback *per seed*: If an individual full seed cannot be found, its partial
+version is looked up in the index.
+
+
 ## Changelog
 
 See [Changelog](CHANGES.md).
--- a/src/cmdline.cpp
+++ b/src/cmdline.cpp
@@ -54,7 +54,7 @@ CommandLineOptions parse_command_line_arguments(int argc, char **argv) {
     args::ValueFlag<int> end_bonus(parser, "INT", "Soft clipping penalty [10]", {'L'});
 
     args::Group search(parser, "Search parameters:");
-    args::Flag mcs(parser, "mcs", "Use multi-context seeds for finding hits", {"mcs"});
+    args::Flag mcs(parser, "mcs", "Use extended multi-context seed mode for finding hits. Slightly more accurate, but slower", {"mcs"});
     args::ValueFlag<float> f(parser, "FLOAT", "Top fraction of repetitive strobemers to filter out from sampling [0.0002]", {'f'});
     args::ValueFlag<float> S(parser, "FLOAT", "Try candidate sites with mapping score at least S of maximum mapping score [0.5]", {'S'});
     args::ValueFlag<int> M(parser, "INT", "Maximum number of mapping sites to try [20]", {'M'});
--- a/src/nam.cpp
+++ b/src/nam.cpp
@@ -264,43 +264,34 @@ std::pair<int, std::vector<Nam>> find_nams_rescue(
         }
     };
     std::array<robin_hood::unordered_map<unsigned int, std::vector<Match>>, 2> matches_map;
-    std::vector<RescueHit> hits_fw;
-    std::vector<RescueHit> hits_rc;
+    std::array<std::vector<RescueHit>, 2> hits;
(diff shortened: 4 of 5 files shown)


marcelm:
(Comparisons are still running, I will update the tables below when the results are in.)

# Comparing accuracy (paf) - sim3

8f79881 Version 0.15.0
5941441 Rescue using partial hits, even in non-MCS mode

library                | 8f79881 | 5941441 | difference
-|-|-|-
sim3-drosophila-50-se  | 83.4170 | 85.9487 | +2.5317
sim3-drosophila-75-se  | 87.3502 | 87.7274 | +0.3772
sim3-drosophila-100-se | 88.5679 | 88.7693 | +0.2014
sim3-drosophila-150-se | 90.3653 | 90.3935 | +0.0282
sim3-drosophila-200-se | 91.4271 | 91.4321 | +0.0050
sim3-drosophila-300-se | 92.7857 | 92.7790 | -0.0067
sim3-maize-50-se       | 43.7357 | 44.4181 | +0.6824
sim3-maize-75-se       | 56.7183 | 56.7283 | +0.0100
sim3-maize-100-se      | 65.1711 | 65.1622 | -0.0089
sim3-maize-150-se      | 77.0898 | 77.1823 | +0.0925
sim3-maize-200-se      | 83.4425 | 83.5704 | +0.1279
sim3-maize-300-se      | 89.1560 | 89.3088 | +0.1528
sim3-CHM13-50-se       | 81.2211 | 82.8875 | +1.6664
sim3-CHM13-75-se       | 87.4537 | 87.8153 | +0.3616

Average difference se: +0.4444

# Comparing accuracy (paf) - sim5

8f79881 Version 0.15.0
5941441 Rescue using partial hits, even in non-MCS mode

library                | 8f79881 | 5941441 | difference
-|-|-|-
sim5-drosophila-50-se  | 76.9207 | 82.8214 | +5.9007
sim5-drosophila-75-se  | 83.4332 | 85.4542 | +2.0210
sim5-drosophila-100-se | 84.9742 | 86.9654 | +1.9912
sim5-drosophila-150-se | 87.5069 | 89.0132 | +1.5063
sim5-drosophila-200-se | 89.2447 | 90.0349 | +0.7902
sim5-drosophila-300-se | 91.2737 | 91.3069 | +0.0332
sim5-maize-50-se       | 39.0947 | 40.5819 | +1.4872
sim5-maize-75-se       | 51.1997 | 51.7045 | +0.5048
sim5-maize-100-se      | 58.7611 | 59.3002 | +0.5391
sim5-maize-150-se      | 70.7927 | 71.3318 | +0.5391
sim5-maize-200-se      | 77.6704 | 78.0476 | +0.3772
sim5-maize-300-se      | 84.6004 | 84.7826 | +0.1822
sim5-CHM13-50-se       | 74.2728 | 78.0569 | +3.7841
sim5-CHM13-75-se       | 82.7277 | 84.3564 | +1.6287
sim5-CHM13-100-se      | 85.3533 | 86.9793 | +1.6260

Average difference se: +1.5274

marcelm:
Here are some initial runtime measurements.

# Comparing user time in seconds (paf) - sim5

8f79881 Version 0.15.0
5941441 Rescue using partial hits, even in non-MCS mode

library                | 8f79881 | 5941441 | difference
-|-|-|-
sim5-drosophila-50-se  | 13 | 11 | -2
sim5-drosophila-75-se  | 16 | 13 | -3
sim5-drosophila-100-se | 17 | 14 | -3
sim5-drosophila-150-se | 20 | 17 | -3
sim5-drosophila-200-se | 24 | 20 | -4
sim5-drosophila-300-se | 34 | 28 | -6
sim5-maize-50-se       | 176 | 159 | -16
sim5-maize-75-se       | 178 | 168 | -10
sim5-maize-100-se      | 174 | 157 | -17
sim5-maize-150-se      | 168 | 161 | -6
sim5-maize-200-se      | 181 | 167 | -13
sim5-maize-300-se      | 219 | 202 | -17
sim5-CHM13-50-se       | 189 | 181 | -7
sim5-CHM13-75-se       | 188 | 187 | -1
sim5-CHM13-100-se      | 199 | 189 | -10

marcelm:
(To clarify: A while ago, you mentioned that Ivan tried this approach in order to speed up MCS mode - at least that is how I understood it at the time. That is where I got it from.)

> Am I interpreting it correctly that:
> 
>     1. This PR enables us to use the same index for non-mcs and mcs, thus we can safely merge this into main.

Yes. (But more accurately, we wouldn’t even have a "pure" non-MCS mode anymore.)

> Also, because of this very interesting result, should we now benchmark three versions in the paper:
> 
>     * v0.15.0
>     * mcs approach 1 (current approach in paper)
>     * mcs approach 2 (@Itolstoganov's suggestion)

My first thought is that this is too much - we already have too many plots. The difference wouldn’t be visible anyway (maybe except for read length 50). We should focus on one MCS approach and present that. Maybe add a sentence or perhaps a single plot that shows what the difference is in results between approach 1 and 2.

Alternatively, maybe we can find a combined approach that is (nearly) as fast as the first approach and as accurate as the second.

>     2. For paper benchmarks, it is then not correct to benchmark latest main in non-mcs and mcs modes, because both versions use the "mcs feature", they are thus two different variants of mcs. So we still need to invoke v0.15.0 in the benchmarks.

Yes, but we could also choose to abandon approach 1 entirely and change `--mcs` to mean "enable approach 2", giving us again the non-MCS and MCS modes.

(However, for the next release, we would definitely want to enable MCS approach 1 by default, but we could do that afterwards.)

ksahlin:
> We should focus on one MCS approach and present that. Maybe add a sentence or perhaps a single plot that shows what the difference is in results between approach 1 and 2.

Ok, I agree.

> Alternatively, maybe we can find a combined approach that is (nearly) as fast as the first approach and as accurate as the second.

You mean the opposite, right? That is, "as fast as the second approach and as accurate as the first"? 

> for the next release, we would definitely want to enable MCS approach 1 by default

hmm, you mean "definitely want to enable MCS approach 2 by default", right? because according to this PR (which is approach 2), approach 2 is both faster and more accurate. Approach 1 (the "previous" mcs in the paper) still has the accuracy-speed tradeoff, making it a more subjective decition.

marcelm:
> > Alternatively, maybe we can find a combined approach that is (nearly) as fast as the first approach and as accurate as the second.
> 
> You mean the opposite, right? That is, "as fast as the second approach and as accurate as the first"?

Yeah, I mixed them up.

> > for the next release, we would definitely want to enable MCS approach 1 by default
> 
> hmm, you mean "definitely want to enable MCS approach 2 by default", right?

Yes, same mixup.

ksahlin:
> My first thought is that this is too much - we already have too many plots. The difference wouldn’t be visible anyway (maybe except for read length 50). We should focus on one MCS approach and present that. Maybe add a sentence or perhaps a single plot that shows what the difference is in results between approach 1 and 2.

Yes, maybe we could, briefly write that we have have an additional approaches -  then show a single plot in supplementary comparing the two where we look at accuracy and runtime (particularly the speed difference might be noticeable?)


> (However, for the next release, we would definitely want to enable MCS approach 1 [TYPO: meant approach 2] by default, but we could do that afterwards.)

Yes, I agree with this.

marcelm:
Shall I start benchmarks for this vs v0.15.0?

ksahlin:
I realize that when I wrote v0.15.0 above I meant whatever was in our paper, which is actually v0.14.0.

Can we benchmark as identical to what we have in the preprint already, i.e. v0.14.0?

ksahlin:
> See my e-mail: I did run benchmarks for v0.15.0 as well, but mapping rate and accuracy is identical to 0.14.0.
Great then I don't mind which one. Let's do v0.15.0 then(?), but as long as Ivan and you are synced @Itolstoganov

PR #478, opened 2025-01-27
Some refactoring

marcelm:
I didn’t get anywhere with the actual change I wanted to make (related to MCS), but here’s at least some refactoring I did along the way. The biggest change is factoring out a `get_nams()` function that contains some code previously duplicated between `align_or_map_single` and
`align_or_map_paired`.


ksahlin:
Excellent, approved!

PR #479, opened 2025-01-28
Revert PR #472 "Always pick first syncmer as base"

marcelm:
... because of a decrease of accuracy in our tests for CHM13 (mostly at telo- and centromers).

As discussed by e-mail, the problem is likely because non-canonical randstrobes, as introduced by #472, lead to biased filtering of overrepresented randstrobes: Before #472, that is, with canonical randstrobes, we would treat a randstrobe and its "reverse complement" (its version with both strobes swapped) the same. A randstrobe in the query would thus be considered overrpresented if the corresponding sequence or its reverse complement were overrepresented. With #472, this was no longer the case and whether a randstrobe is filtered or not now depends on whether it is overrpresented in the orientation that it has in the query.


ksahlin:
Approved!

marcelm:
The reverted changes are now in branch `base-first` and could be re-applied if we solve the accuracy problem.

PR #482, opened 2025-02-25
Fix incorrect match start coordinates when adding partial matches

marcelm:
This is only relevant in the MCS-R code path, that is, when `--mcs` is *not* provided.

The code as written was correct as long as we had the "always use first syncmer as base" variant (I had even eliminated the `partial_start` and `partial_end` attributes in that variant). When I reverted that change, I did not update this line to use the correct partial start and end coordinates.

The number of alignments with score less than 40 goes down by 75% with this change (only tested for one dataset).

Accuracy increases by ca 0.01 percentage points on sim3, ca 0.03 percentage points on sim5 (based on more than one library, but not all).


ksahlin:
Nice -- approved!

PR #489, opened 2025-03-21
Avoid processing hits any further when we would run rescue anyway

marcelm:
In `find_nams`, we find hits, find matches and then merge matches into NAMs. However, if we determine that we need to run rescue, we throw away all of this anyway. To determine whether rescue needs to run, we only need the hits, so this PR changes `find_nams` into `find_hits`, which only finds hits. If it turns out that we do not run rescue, then looking up the hits to find matches and merging matches into NAMs is done later.

This saves approximately 1% runtime on CHM13, about 2% on drosophila and 8% on chrY.

This PR needs to be cleaned up a little bit more, but I won’t have time for that today.

For some reason, the results change very slightly although I think matches and NAMs are computed exactly the same way as before. My guess at the moment is that the iteration over the references in the matches_map changes (it’s just an unordered_map, so I believe the order in which it iterates over its keys is not defined).

### DIFF
--- a/src/aln.cpp
+++ b/src/aln.cpp
@@ -1041,9 +1041,9 @@ std::vector<Nam> get_nams(
 
     // Find NAMs
     Timer nam_timer;
-    auto [nonrepetitive_fraction, n_hits, partial_hits, nams] = find_nams(query_randstrobes, index, map_param.use_mcs);
-    statistics.tot_find_nams += nam_timer.duration();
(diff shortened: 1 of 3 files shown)


ksahlin:
Great refactoring!

marcelm:
The results were slightly different because I called `reserve()` for `robin_hood::unordered_map` in a different place where it did not have an effect. Fixed now.

PR #490, opened 2025-03-25
Let NAM finding functions work on one strand only

marcelm:
This refactors the functions in `nam.cpp` so that they no longer have to be aware of reverse complements. In particular, `find_hits()` no longer returns two `Hit` vectors, but just one. We then call it twice, once with the forward randstrobes and once with the reverse randstrobes.

* `find_nams_rescue` now walso works on one strand only.
* We no longer need `merge_matches_into_nams_forward_and_reverse`
* The `randstrobes_query()` function no longer produces a single vector with `QueryRandstrobes` (where the itmes in the first half have `is_revcomp=true` and the ones in the second half have `is_revcomp=false`). Instead, it produces two vectors, one with the forward and one with the reverse-complemented randstrobes.
* `QueryRandstrobe` no longer needs an `is_revcomp` attribute. 

One consequence at the moment is that `get_nams()` gets a bit longer because it needs to call `find_hits` and `find_matches_rescue` twice and then combine the results. However, I actually find it quite helpful to see how the results are combined explicitly. For example, a line like
```c++
int nonrepetitive_hits = hits[0].size() + hits[1].size();
```
makes it now clear that the number of nonrepetitive hits corresponds to the number on the forward *and* reverse-complemented strand.

To be clear, this PR does not implement the full separation that I mentioned over e-mail where we would search for the forward and reverse-complemented query entirely separately. A couple of further changes would be required for that if we want to go that way. The change in this PR is an improvement in any case IMO.

As it is, the PR does not change how strobealign works, except in one case: The MCS-R rescue step is now performed per strand, that is, it is run for each strand that does not have any hits. Previously, it would only be run if neither strand had any hits. This changes results a tiny bit.


ksahlin:
very good!

> As it is, the PR does not change how strobealign works, except in one case: The MCS-R rescue step is now performed per strand, that is, it is run for each strand that does not have any hits. Previously, it would only be run if neither strand had any hits. This changes results a tiny bit.

Could we distinguish the two cases (1) no hit (e.g. due to mutation/error), (2) filtered hit (>1000 occurrences) when deciding to rescue? Rescue is only viable if there are no or few hits due to case 2.

marcelm:
> Could we distinguish the two 'no hit' cases (1) No occurrence in index (e.g. due to mutation/error), (2) >1000 occurrences in index (i.e. filtered hit) when deciding to rescue?

Yes, this is already done this way: When we look up full seeds, we count `total_hits` (includes both normal and repetitive hits) and `nonrepetitive_hits`. Partial seeds are then only looked up if `total_hits` is zero.

>  Rescue is only viable if there are no or few hits due to case 2.

You bring in the additional aspect that it may be worthwhile to look up partial seeds even when `total_hits` is not zero but just very small. I believe I actually tried this in the beginning, but the results were not better. I don’t have any notes, though, and may be misremembering.

ksahlin:
> Yes, this is already done this way: When we look up full seeds, we count total_hits (includes both normal and repetitive hits) and nonrepetitive_hits. Partial seeds are then only looked up if total_hits is zero.

I don't understand this statement. It seems we want to look up partial hits if `nonrepetitive_hits < T` and `total_hits >> nonrepetitive_hits`? If I interpret you correctly, having `total_hits == 0` means that no hits are found, menaing the read has lots of mutations/errors. 

Or maybe I misinterpret what `repetitive` means in this context? I thought it meant hits above our threshold of 100, but maybe you just mean any hit that occurs twice or more?

marcelm:
Hm, not sure I understand. Are we talking about the same thing? I mean the R in MCS-R, not `find_nams_rescue` (which can be disabled with `-R 0`).

The code is here in any case:
https://github.com/ksahlin/strobealign/blob/41579b9c790530f633a88956abc3757d61526d79/src/nam.cpp#L262

> If I interpret you correctly, having total_hits == 0 means that no hits are found, menaing the read has lots of mutations/errors.

Yes, isn’t that what we want? If we find no hits when we use full seeds, the seeds are apparently too specific and we try with less specific partial seeds instead. If we find repetitive seeds (>100 or whatever the threshold is), we don’t want to fall back to partial seeds because we would just find even more hits.

PR #491, opened 2025-03-26
Use a "trailer" in the commit message to mark when baseline results change

marcelm:
This removes the baseline-commit.txt file, which previously stored the commit that represents the current baseline. It’s a bit annoying to have to maintain this file during development because it has to be updated manually on every rebase (which I do quite a lot locally to bring a branch into shape).

With this, the baseline commit is defined as the most recent commit that contains the line `Is-new-baseline: yes` in its commit message. Lines formatted like this at the end of a commit message are called "trailers" and are a Git feature that is also used for other things (like `Co-authored-by: ...`).

Whenever results change, the commit that introduces that change needs to contain the `Is-new-baseline: yes` trailer.

An earlier commit already used this trailer, so this commit itself does not need to set it.


ksahlin:
approved

PR #494, opened 2025-04-10
Extend Python bindings

marcelm:
The Python bindings were a bit out of date. Previously, strobealign had a single `find_nams` function, but this was split up into `find_hits`, `hits_to_matches` and `merge_matches_into_nams`. The bindings only had the first function. This PR adds the other two.


ksahlin:
great, approved!

PR #497, opened 2025-05-15
Fix out-of-bounds memory access

marcelm:
`find_partial` and `find_full` can return the special value `end()`, which indicates that the hash was not found. We need to check for this value before using it as an index into the randstrobes vector.

This fixes a crash on macOS.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 # Strobealign Changelog
 
+## development version
+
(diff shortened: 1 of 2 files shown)


ksahlin:
Fixed 😌

marcelm:
I suggest we release 0.16.1 with this fix.

ksahlin:
ok

marcelm:
Release done

PR #499, opened 2025-05-23
Sort NAM pairs by score, not by number of hits

marcelm:
(Opening a PR for @ksahlin’s changes)

Also, the score cutoff in SE mode uses actual score and not n_hits. This leaves accuracy mostly unchanged (though there are a couple of differences both up and down in SIM6), but will improve both SE and PE mapping-only score when using chains (this is unmerged).

Is-new-baseline: yes


ksahlin:
Approved

PR #501, opened 2025-06-02
Missing cstdint header

starsareintherose:
fix: g++ 15.1.1


marcelm:
Thanks, good find!

I cannot quite use the PR as-is because the files in the `ext/` directory should be direct copies of the upstream code if at all possible (because custom patches disappear the next time we update the code). Fortunately, the problem has been fixed in zstr already, so I have opened a separate PR that just updates zstr to the newer version. I’ll close this PR, but have added you as a co-author to that other commit so you get credit.

PR #504, opened 2025-06-04
Chaining

NicolasBuchin:
This PR introduces collinear chaining as the new default method for mapping/aligning in Strobealign, replacing NAMs. NAMs are still supported and can be enabled using the `--nams` flag.

It adds new CLI parameters to fine-tune the chaining behavior:
`--nams` Use NAMs instead of collinear chaining
`-H [INT]` Chaining look-back heuristic (default: 50)
`--gd=[FLOAT]` Diagonal gap cost (default: 0.1)
`--gl=[FLOAT]` Gap length cost (default: 0.05)
`--vp=[FLOAT]` Best chain score threshold (default: 0.7)
`--sg=[INT]` Skip distance allowed on the reference

I should add that the CI pipeline currently fails because the output format and alignments produced by collinear chaining differ from those of NAMs, causing test assertions to break, and should be handled at some point.

## Next Steps

- [x] Update test suite to reflect new chaining behavior
- [x] Document changes usage in CHANGES.md
- [x] Benchmark performances comparing chaining vs NAMs
- [x] Fix the reporting of chains that cover the same region of a reference

### DIFF
--- a/src/chain.cpp
+++ b/src/chain.cpp
@@ -11,7 +11,6 @@
 #include <utility>
 
 #include "pdqsort/pdqsort.h"
-
 #include "aln.hpp"
 #include "chain.hpp"
 #include "randstrobes.hpp"
@@ -177,8 +176,8 @@ float collinear_chaining(
     predecessors.assign(n, -1);
     float best_score = 0;
 
-    for (int i = 0; i < n; ++i) {
-        const int lookup_end = std::max(0, i - chaining_params.max_lookback);
+    for (size_t i = 0; i < n; ++i) {
+        const int lookup_end = std::max(0, static_cast<int>(i) - chaining_params.max_lookback);
 
         for (int j = i - 1; j >= lookup_end; --j) {
             const Anchor& ai = anchors[i];
@@ -228,7 +227,7 @@ void extract_chains_from_dp(
     std::vector<bool> used(n, false);
 
     std::vector<std::pair<int, float>> candidates;
-    for (int i = 0; i < n; ++i) {
+    for (size_t i = 0; i < n; ++i) {
         if (dp[i] >= valid_score) {
             candidates.push_back(std::make_pair(i, dp[i]));
         }
@@ -279,24 +278,6 @@ void extract_chains_from_dp(
     }
 }
 
-template <typename T>
-bool by_score(const T& a, const T& b) {
-    return a.score > b.score;
-}
-
-void shuffle_top_chains(std::vector<Nam>& chains, std::minstd_rand& random_engine) {
-    if (chains.empty()) {
-        return;
-    }
-    auto best_score = chains[0].score;
-    auto it = std::find_if(chains.begin(), chains.end(), [&](const Nam& chain) {
-        return chain.score != best_score;
-    });
-    if (it > chains.begin() + 1) {
-        std::shuffle(chains.begin(), it, random_engine);
-    }
-}
-
 std::vector<Nam> get_chains(
     const klibpp::KSeq& record,
     const StrobemerIndex& index,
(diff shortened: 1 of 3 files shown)


marcelm:
Awesome! Let’s get this into shape so it can be merged.

I’m looking at the failing tests at the moment. When I run `tests/run.sh`, I get:
```
Failure running 'diff tests/phix.se.sam phix.se.sam'
```
When I look at the diff, I can see that the alignments are actually all identical before and after but that the mapping quality changes. Whereas before, we had 42 alignments with quality 60 and 3 with quality 0, afterwards we get 24 with quality 60 and 21 with quality zero. Alignment quality 0 means that there are multiple equally good alignments. This may be an indication that multiple chains are found that result in the same alignment. Either we need to detect these duplicates and remove them or the chaining algorithm needs to be changed to not produce them in the first place. The latter would be preferable because computing alignment is expensive and these duplicates seem to happen quite often.

marcelm (commented):
I’ve done a first round of code review. Most of it is about code style, but there’s also a potential bug with `int` vs `unsigned`.

marcelm:
> You are right, we do report chains that cover the same region on a reference.
>
> I fixed it, by preventing reporting chains with an already used anchor.

Nice, good that was a small fix.

> The tests should pass now.

They don’t but the failure is somewhere else (in the PAF output). Don’t you run the tests locally? You should do that before every push.

src/chain.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+#include "index.hpp"
+#include "indexparameters.hpp"
+#include "nam.hpp"
+#include "readlen.hpp"
+
+struct Anchor {
+    int query_start;
+    int ref_start;
+    int ref_id;
```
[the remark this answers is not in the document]
NicolasBuchin: should we also changed the Nam struct to use unsigned ?

CHANGES.md
```diff
@@ -1,5 +1,21 @@
 # Strobealign Changelog
 
+## developpment version
```
marcelm, line 3:
```suggestion
## development version
```

CHANGES.md
```diff
@@ -1,5 +1,21 @@
...
+
+* #504: Introduce collinear chaining as the new default mapping and alignment
```
marcelm, line 5: This should be split up into a short changelog entry and a section in the README

src/chain.cpp
```diff
@@ -0,0 +1,385 @@
...
+        );
+    }
+}
+
+template <typename T>
+bool by_score(const T& a, const T& b) {
+    return a.score > b.score;
+}
+
+void shuffle_top_chains(std::vector<Nam>& chains, std::minstd_rand& random_engine) {
```
marcelm, line 287: This appears to be a copy of `shuffle_top_nams`, just call that function instead.

src/chain.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+std::vector<Nam> get_chains(
+    const klibpp::KSeq& record,
+    const StrobemerIndex& index,
+    const MappingParameters& map_param,
+    const IndexParameters& index_parameters,
+    std::minstd_rand& random_engine
+);
+
+static inline float mg_log2(float x) {
```
marcelm, line 30: Please link to the code in minimap2 where this is copied from.

marcelm:
I have now rebased this PR on top of main and cleaned up the commit history as much as I could. I also ran all the tests for each commit and set the `Is-new-baseline: yes` trailer if necessary.

It is unfortunate that GitHub only shows my avatar next to the commits that Nicolas has authored, making it appear as if I were the author (maybe it is because Nicolas uses the default avatar). However, the commits themselves are correctly attributed to Nicolas as author (while I am the committer).

There are some unaddressed review comments that I will look into next before merging the PR.

ksahlin:
very nice!

marcelm:
I think this deserves to be merged now.

The chaining code as it is here is what we are running in our current benchmarks, so it makes sense to mark that somehow. I’ll open separate, smaller PRs for the few remaining issues.

@NicolasBuchin Great work, congratulations!

PR #506, opened 2025-06-23
Failure more loudly when tests fail

marcelm:
... and print 'Failure' in red. Previously, we would only silently exit with an exit code of 1.


ksahlin:
approved

PR #514, opened 2025-09-05
Stop checking anchors when we chain to one on the same diagonal

marcelm:
If we find a new anchor that we are going to chain to the current one on the same diagonal, then we assume that this is the correct one and skip checking other anchors.

This speeds up chaining significantly at no loss in accuracy.

This makes chaining typically on par speedwise with NAMs and often faster, except for fruitfly SIM0-SIM4, where it is still a little bit slower.


marcelm:
Plots:
[ends-se-accuracy.pdf](https://github.com/user-attachments/files/22173419/ends-se-accuracy.pdf)
[ends-se-time.pdf](https://github.com/user-attachments/files/22173420/ends-se-time.pdf)

ksahlin:
Very clever! You like that @NicolasBuchin ?! (https://www.youtube.com/shorts/-KaR2mmifk0 )

I only had a _very quick_ think, but I cannot come up with a scenario where this heuristic would lead to loss compared to the "full" -h 50 cutoff.

Approved IMO

ksahlin:
All read mappings that changes get a higher "number of matching bases" in the PAF file (I think we report number of matching seeds there). Any idea why that is?

NicolasBuchin:
If it is as good then let use it, good find!

PR #517, opened 2025-09-13
Only keeping strobemer matches with minimum strobes distance difference

NicolasBuchin:
A small change that changes the heuristic on adding strobemer matches, now we just keep the ones with the closest distance to the strobemer hit on the query. Not a very noticeable difference on results but should make chaining a tiny bit faster.

### DIFF
--- a/src/chain.cpp
+++ b/src/chain.cpp
@@ -17,16 +17,29 @@ void add_to_anchors_full(
     size_t position
 ) {
     int min_diff = std::numeric_limits<int>::max();
+    std::vector<size_t> best_positions;
+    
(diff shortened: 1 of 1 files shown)


marcelm:
Hi, I’ve run our benchmark on this commit and the previous one and don’t see any difference, not in accuracy but also not in runtime:
* [ends-se-time.pdf](https://github.com/user-attachments/files/23121879/ends-se-time.pdf)
* [ends-se-accuracy.pdf](https://github.com/user-attachments/files/23121878/ends-se-accuracy.pdf)

I therefore suggest to not merge this PR as it just adds (a little bit of) complexity.

PR #522, opened 2025-10-03
Adjust chunk size depending on read length

marcelm:
Let chunk size be in terms of nucleotides, not number of reads

Note that there’s a hidden command-line option `--chunk-size`. This is now re-intepreted as meaning the number of *nucleotides* in a chunk. This way, there are fewer reads in a chunk if the reads are longer.

This avoids the problem that we do not use all input threads for small input
datasets with long reads (if the input had 50000 reads, we would use at most
5 threads, for example).

Experiments suggest that the chunk size can be relatively small without
adding too much overhead. Even for 100 nt reads and a chunk size of 1 read,
mapping is just a couple of percent slower than with bigger chunk sizes.


ksahlin:
Approved

PR #526, opened 2025-10-30
Minor logging improvements

marcelm:
- When tracing, do not show all single-anchor chains
- Log the number of anchors (with `-v`)
- When tracing, show the number of matches of a `Nam`


ksahlin:
Approved!

PR #530, opened 2025-12-15
Fix a memory leak when -r is used

marcelm:
When option `-r` is *not* provided, we read in the first 500 reads to estimate the average read length and then call `InputBuffer::rewind_reset()` to "rewind" to the beginning of the file. `InputBuffer` uses `RewindableFile`, which keeps all data read from it in memory until `rewind()` is called. (A RewindableFile can be rewound once.)

When `-r` is provided, we do not estimate average read length, never call `InputBuffer::rewind_reset()`, never call `RewindableFile::rewind()` and therefore grow the input buffer indefinitely until the end of file has been reached or we are out of memory.

The solution chosen here is to always estimate the length of the input reads even when -r is provided. This is somewhat of a workaround as it would be nicer to fix InputBuffer/RewindableFile, but it is actually helpful because this allows the user to check whether what they think the read length should be matches what it actually is.

### DIFF
--- a/src/fastq.cpp
+++ b/src/fastq.cpp
@@ -35,8 +35,8 @@ namespace {
 RewindableFile::RewindableFile(const std::string& filename)
     : reader(nullptr),
     rewindable(true),
-    stream_(klibpp::make_ikstream(this, rewind_read, 16384)) {
-
+    stream_(klibpp::make_ikstream(this, rewind_read, 16384))
+{
     reader = make_reader(filename);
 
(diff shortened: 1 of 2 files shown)


SergeWielhouwer:
I’m still on 0.16.1, haha. Saw the new release, so thought I’d ask ;).

I’ll try to switch to 0.17.0 soon, but need to run some quick performance tests first to make sure it still works well in my pipelines (I use strobealign for things like quick reference-based insert size checks and verifying if samples are of potentially of saliva origin).

Thanks again,
Serge

PR #531, opened 2025-12-16
Heuristic that ties to chain on the best current chain, which might not otherwise be tested

NicolasBuchin:
Because of the heuristic that stops looking for a best previous anchor if they are both on the diagonal, we might lose the best current chain. This PR ensures we keep trying it at least once, and it will be prioritized as the best parent.

This will be especially useful for piecewise extension in the future.

Here is an example of a chain that won't try the previous best chain:
chain_id=0_score=349 73
Here is the same chain generated with this heuristic:
chain_id=0_score=353 74

### DIFF
--- a/cpp/chain.cpp
+++ b/cpp/chain.cpp
@@ -144,7 +144,7 @@ float Chainer::collinear_chaining(
             const float score = compute_score(dq, dr);
 
             const float new_score = dp[j] + score;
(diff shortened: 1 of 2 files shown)


NicolasBuchin:
I couldn't see any real impact on performances. But might be worth a benchmark ?

PR #535, opened 2026-01-13
Add a Rust port of strobealign

### LINKED ISSUE #366: Structure of aln.cpp
This shows which functions call which other functions in `aln.cpp`. This is an updated version of #98.

I’m documenting this because I think some functions should be renamed to make it easier to remember what function does what. I will update this over time.

# Paired end

```mermaid
graph LR;
    align_PE_read --> shuffle_top_nams;
    align_PE_read --> output_hits_paf_PE;
    align_PE_read --> align_PE;
    align_PE_read --> get_best_map_location;

    align_PE --> is_proper_nam_pair;
    align_PE --> rescue_read;
    align_PE --> extend_seed;
    align_PE --> get_best_scoring_nam_pairs;
    align_PE --> joint_mapq_from_high_scores;
    align_PE --> is_proper_pair;
    align_PE --> get_mapq;
    align_PE --> top_dropoff;

    extend_seed --> hamming_distance;
    extend_seed --> hamming_align;
    extend_seed --> Aligner.align;
    get_best_scoring_nam_pairs --> is_proper_nam_pair;
    rescue_mate --> has_shared_substring;
    rescue_mate --> Aligner.align;
    rescue_read --> extend_seed;
    rescue_read --> rescue_mate;
    rescue_read --> get_best_scoring_pairs;
    rescue_read --> joint_mapq_from_alignment_scores;
    rescue_read --> is_proper_pair;
    joint_mapq_from_high_scores --> joint_mapq_from_alignment_scores;
    get_best_map_location --> get_best_scoring_nam_pairs;
```

# Single end

```mermaid
graph LR;
    align_SE --> reverse_nam_if_needed;
    align_SE --> extend_seed;
    extend_seed --> hamming_distance;
    extend_seed --> hamming_align;
    extend_seed --> Aligner.align;
    align_SE_read --> shuffle_top_nams;
    align_SE_read --> align_SE;
    align_SE_read --> output_hits_paf;
```

### LINKED ISSUE #481: Usage of mmap for compressed input leads to inflated resident set size
mmapped memory is included in the resident set size (RSS), which means that strobealign appears to use a huge amount of memory when reading a gzipped input file because that is mapped into memory.

### LINKED ISSUE #534: Rust version of strobealign
I have been working on a Rust port of strobealign over the last months, available at https://github.com/marcelm/rstrobes. I’d like to suggest that we switch to this version and discuss here under which conditions this can happen and how.

## Motivation

I won’t go into a detailed discussion of Rust vs C++, only share some selected points. For me, it has been quite pleasant working in Rust. For example, some bugs are prevented by the compiler. It’s sometimes an effort to get the code to compile, but when it does, it often works right away. I rather spend my time interacting with the compiler and reading its (really good) error messages rather than in the debugger with code that compiles but is buggy.

The tooling is so much nicer. No need having to deal with CMake. Just `cargo run`, `cargo build`, `cargo test`. Using external dependencies is very easy (`cargo add`). Also, the documentation is really good.

No more header files – it’s annoying to have to change method signatures in *two* files all the time.

Python-like string formatting (`info!("Estimated read length: {} bp", r);` instead of `logger.info() << "Estimated read length: " << opt.r << " nt";`). (Granted, this would also be available with C++20.)

I like how I can see at every call site whether something is passed by value, as a reference or as a mutable reference. In C++, I have to inspect the function signature to know.

Less boilerplate: Having to add `std::ostream& operator<<(...)` for every struct that I want to print to stdout is annoying. In Rust, an automatic implementation is added by adding `#[derive(Debug)]` before the struct. (Same for `operator<=` etc.)

Finally, since @NicolasBuchin is working on piecewise alignment in #533 using a Rust library, it would be nice to not have to use C++ bindings but be able to use it directly.

I’m happy to go into more detail, but maybe that’s enough for now.

---

The Rust version is very close to being a drop-in replacement for strobealign 0.17, but some work remains.

## Accuracy

My focus was first on getting equivalent results to the C++ version, and this has been achieved: The accuracy of the port is identical to that of strobealign 0.17. Mapping results for multimappers can be different because a different random number generator is used, which means that `shuffle_top_nams` may choose a different NAM.

## Runtime

The Rust version is still up to ~10% slower than the C++ version for read lengths over 200 nt (single-end reads, mapping only). Making the port fast wasn’t my focus initially, so this is not suprising. I have spent some time on optimization over the holidays, though, and have some ideas on how to proceed in order to achieve the same speed as the C++ version even for longer reads. Indexing is actually faster than in the C++ version already, and paired-end mapping is faster for short reads.

## Memory

I noticed just now that memory usage is a little higher for the Rust version; worst is rye with 36.5 GB instead 35 GB. I would guess this is not hard to fix.

## Missing features

The ones that are IMO required (at least the first one) before switching are:
* ~~No `--use-index` option: The Rust version can create on-disk indices with `--create-index` (and the created files are identical to the ones created by the C++ version), but it cannot read them.~~ This is now implemented
* Input reads cannot be in FASTA format (they must be FASTQ).
* Some tests have not been ported
* Some sanity checks for parameters are missing (`-b` and `--aux-len` should not collide, `--aux-len` must be at least 6 and not larger than 27).

These are IMO less important and we could do without 
* There are no Python bindings. I have started on this and it isn’t hard, but the other things seemed more important. Also, the reason I added them was to be able to run experiments without having to modify the C++ code. I haven’t use the bindings that much recently, so we could do without them for now. Those who want bindings can use the C++ version in the meantime.
* I have not implemented the `--index-statistics` option. I have never used it so my interest in that is not so high. I don’t even if it works as intended.

## Plan

As a first step, I suggest to open a PR that adds the Rust code, but retains the C++ code in a separate directory so that one can choose which version to build. After that is merged, I/we implement the missing features listed above (and maybe improve runtime and memory usage). At the same time, the open PRs (in particular #533) can be ported to Rust.

marcelm:
This adds the Rust port of strobealign as suggested in #534.

This moves the C++ sources from `src/` to `cpp/`. Compile with `cmake` as before.

Rust sources are in `src/`. Use `cargo build`, `cargo run`, `cargo test` as normal for Rust projects. Remember to use option `--release` (or the short form `-r`) in order to get an optimized build. See also the new section in `CONTRIBUTING.md`.

For a couple of commits from external contributors that I directly translated from C++ to Rust, I tried to give credit by setting the commit author appropriately.

It’s a lot of commits overall, but this represents work spread out over three years. I don’t think it makes sense to squash everything into a single commit. Especially the later commits are self-contained as I usually try to make them (that is, the code compiles and `cargo test` runs successfully).

Since it is quite impossible to review this code, here’s at least proof that the results are as good as before: [ends.pdf](https://github.com/user-attachments/files/24594613/ends.pdf) (Legend get lost somehow, green is strobealign 0.17.0, black is the Rust version). The runtime looks actually a bit better than I reported in #534; this is probably because this plot also shows extension alignment mode (squeezing the mapping-only mode runtime).

Closes #534
Closes #266
Closes #481
Closes #366

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -6,7 +6,7 @@ env:
   CARGO_TERM_COLOR: always
 
 jobs:
-  build:
+  test:
     # Run for PRs only if they come from a forked repo (avoids duplicate runs)
     if: >-
       github.event_name != 'pull_request' ||
@@ -23,3 +23,12 @@ jobs:
       run: cargo build --verbose
     - name: Run cargo test
       run: cargo test --verbose
+    - name: Install Linux dependencies
+      if: runner.os == 'Linux'
+      run: sudo apt-get install samtools
+    - name: Install macOS dependencies
+      if: runner.os == 'macOS'
+      run: brew install samtools
+    - name: Run shell test
+      run: tests/rust.sh
+
--- a/tests/rust.sh
+++ b/tests/rust.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -xeuo pipefail
+set -euo pipefail
 
 if [[ $OSTYPE = linux-gnu ]]; then
     color="--color=always"
@@ -9,12 +9,11 @@ fi
 
 function strobealign() {
     echo "Testing '${@}'" >&2
-    target/debug/strobealign "${@}"
-#     if ! target/debug/strobealign "${@}" 2> testlog.txt; then
-#         cat testlog.txt
-#         echo "Failure"
-#         exit 1
-#     fi
+    if ! target/debug/strobealign "${@}" 2> testlog.txt; then
+        cat testlog.txt
+        echo "Failure"
+        exit 1
+    fi
 }
 
 function diff() {
@@ -42,10 +41,13 @@ cat tests/phix.1.fastq | strobealign -x tests/phix.fasta - | tail -n 11 > phix.s
 diff tests/phix.se.paf phix.se.paf
 rm phix.se.paf
 
+# TODO
+# This test is disabled for now because we need to look closer into whether
+# the mapping qualities make sense.
 # Paired-end PAF
-strobealign -x tests/phix.fasta tests/phix.1.fastq tests/phix.2.fastq | tail -n 11 > phix.pe.paf
-diff tests/phix.pe.paf phix.pe.paf
-rm phix.pe.paf
+#strobealign -x tests/phix.fasta tests/phix.1.fastq tests/phix.2.fastq | tail -n 11 > phix.pe.paf
+#diff tests/phix.pe.paf phix.pe.paf
+#rm phix.pe.paf
 
 # Single-end abundance estimation
 strobealign --aemb tests/phix.fasta tests/phix.1.fastq > phix.abun.se.txt


marcelm:
I have improved runtime slightly by 1) sorting anchors with an unstable sort instead of a stable sort (which gives the same results since we compare all members of the `Anchor` struct anyway) and 2) compiling for CPU target `x86-64-v2`.

Here are the most recent runtime and accuracy measurements in mapping-only mode: [ends.pdf](https://github.com/user-attachments/files/24752866/ends.pdf)
I’ll add alignment mode plots once they are done.

Further changes:
- `--use-index` is now implemented
- Tests are run on macOS in CI
- Formatted the code with `cargo fmt`. This is enforced during CI.

PR #557, opened 2026-02-10
PAF files now report number of matching bases on the reference instead of number of matching seeds.

NicolasBuchin:
As mentioned to @ksahlin, [minimap2 reports the number of matching bases in PAF files at column 10](https://lh3.github.io/minimap2/minimap2.html#10), we were reporting the number of matching seeds.

This PR changes column 10 in PAF files to be the coverage on the reference so that strobealign's output is the same as minimap2.


NicolasBuchin:
Seems like the CI pipeline still uses the old PAF files in the tests?

PR #565, opened 2026-02-27
Replace the quadratic paired-mapping search with linear-time sweep

NicolasBuchin:
The previous paired-end mapping logic attempted to form pairs by testing all combinations of chains from read1 and read2, ordered by score, with a hard cap on the number of trials. This resulted in worst-case behavior close to O(MAX_PAIRS²) runtime and could not guarantee to find the best pair of chains.

This PR proposes to find pairs by splitting chains of each read into forward and revcomp sets, then sorting in O(N*log(N)) by ref_id then ref_start for each set, and doing a ~O(N+M) sweep on each forward/revcomp pair of sets using 2 pointers. Credit to @ksahlin for the idea.

We can guarantee to find the best scoring pair, that each chain is in one unique pair, but not that all valid pairs of unique chains will be returned.

For scoring pairs, we also introduce a bonus like what is done in paired extension to favor pairs that fit well in the distribution of known paired mappings.

Since returning sorted chains by score isn't needed when calling get_nams_by_chaining(), we extract the logic outside of this function.

Later I plan to introduce this new approach to paired extension, which still uses O(N²) complexity in this PR, but only after some evaluation has been done and we agree to use this new approach for pairing chains.

TODO:
- [ ] Evaluate accuracy and runtime compared to main

### DIFF
--- a/src/details.rs
+++ b/src/details.rs
@@ -23,6 +23,9 @@ pub struct NamDetails {
     pub time_chaining: f64,
     pub time_rescue: f64,
     pub time_sort_nams: f64,
+
+    /// Whether both orientations were tested
+    pub both_orientations: bool,
 }
 
 impl ops::AddAssign<NamDetails> for NamDetails {
--- a/src/maponly.rs
+++ b/src/maponly.rs
@@ -375,7 +375,7 @@ fn split_nams_by_orientation(nams: &mut [Nam]) -> (&mut [Nam], &mut [Nam]) {
     nams.split_at_mut(left)
 }
 
-/// Find most forward/revcomp pairs using a linear two-pointer scan.
+/// Find most forward/revcomp pairs using a two-pointer scan.
 /// Assumes both slices are sorted by (ref_id, projected_ref_start).
 fn find_pairs(fwd: &[Nam], rev: &[Nam], mu: f32, sigma: f32, swap_order: bool) -> Vec<NamPair> {
     let mut out = Vec::new();
--- a/src/maponly.rs
+++ b/src/maponly.rs
@@ -381,7 +381,7 @@ fn find_pairs(fwd: &[Nam], rev: &[Nam], mu: f32, sigma: f32, swap_order: bool) -
     let mut out = Vec::new();
     let max_dist = (mu + 10.0 * sigma).ceil() as usize; // distance cutoff from insert size distribution
     let mut rev_ptr = 0;
-    let mut last_paired: Option<(usize, f64)> = None;
+    let mut last_paired = None;
 
     for f in fwd {
         // Advance revcomp pointer to the first possible candidate
@@ -400,33 +400,33 @@ fn find_pairs(fwd: &[Nam], rev: &[Nam], mu: f32, sigma: f32, swap_order: bool) -
         }
 
         // Scan window of revcomp nams within distance limit.
-        let mut best: Option<(usize, f32)> = None;
+        let mut best = None;
         let mut i = rev_ptr;
         while i < rev.len()
             && rev[i].ref_id == f.ref_id
             && (rev[i].projected_ref_start() - f.projected_ref_start()) <= max_dist
         {
-            let s = rev[i].score;
-            if best.is_none_or(|(_, bs)| s > bs) {
-                best = Some((i, s));
+            let r = &rev[i];
+            // The pairing score gets a bonus based on the reference distance of the two chosen nam
+            // paired and from our current knowledge of the reference distance distribution
+            let x = f.ref_start.abs_diff(r.ref_start);
+            let score = f.score as f64
+                + r.score as f64
+                + 0.001f64.max((normal_pdf(x as f32, mu, sigma) + 1.0).ln() as f64);
+
+            if best.is_none_or(|(_, highest_score)| score > highest_score) {
+                best = Some((i, score));
             }
             i += 1;
         }
 
-        // Best scoring candidate
-        let Some((best_id, best_score)) = best else {
+        // Highest scoring candidate
+        let Some((best_id, score)) = best else {
             continue;
         };
         let r = &rev[best_id];
 
-        // The pairing score get's a bonus based on the reference distance of the two chosen nam
-        // paired and from our current knowledge of the reference distance distribution
-        let score = f.score as f64
-            + best_score as f64
-            + (-20.0f64 + 0.001)
-                .max(normal_pdf(f.ref_start.abs_diff(r.ref_start) as f32, mu, sigma).ln() as f64);
-
-        // If the same recomp nam was paired previously, keep only the better scoring pair.
+        // If the same revcomp nam was paired previously, keep only the better scoring pair.
         if let Some((last_id, prev_score)) = last_paired
             && last_id == best_id
         {
--- a/src/maponly.rs
+++ b/src/maponly.rs
@@ -1,15 +1,16 @@
 use fastrand::Rng;
 
 use crate::chainer::Chainer;
-use crate::details::Details;
+use crate::details::{Details, NamDetails};
 use crate::index::StrobemerIndex;
 use crate::insertsize::InsertSizeDistribution;
 use crate::io::fasta::RefSequence;
 use crate::io::paf::PafRecord;
 use crate::io::record::{End, SequenceRecord};
-use crate::mapper::{NamPair, get_best_scoring_nam_pairs, mapping_quality};
+use crate::mapper::mapping_quality;
+use crate::math::normal_pdf;
 use crate::mcsstrategy::McsStrategy;
-use crate::nam::{Nam, get_nams_by_chaining};
+use crate::nam::{Nam, get_nams_by_chaining, sort_nams};
 
 /// Map a single-end read to the reference and return PAF records
 ///
@@ -23,14 +24,14 @@ pub fn map_single_end_read(
     chainer: &Chainer,
     rng: &mut Rng,
 ) -> (Vec<PafRecord>, Details) {
-    let (nam_details, nams) = get_nams_by_chaining(
+    let (mut nam_details, mut nams) = get_nams_by_chaining(
         &record.sequence,
         index,
         chainer,
         rescue_distance,
         mcs_strategy,
-        rng,
     );
+    nam_details.time_sort_nams = sort_nams(&mut nams, rng);
 
     if nams.is_empty() {
         (vec![], nam_details.into())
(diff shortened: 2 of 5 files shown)


src/details.rs
```diff
@@ -35,6 +35,9 @@ pub struct NamDetails {
     pub time_chaining: f64,
     pub time_rescue: f64,
     pub time_sort_nams: f64,
+
+    /// If of orientations were tested
```
marcelm: Should this be "Whether both orientations were tested"?

src/nam.rs
marcelm: This comment has become incorrect

src/nam.rs
```diff
@@ -178,6 +172,13 @@ pub fn get_nams_by_chaining(
     (nam_details, nams)
 }
 
+pub fn sort_nams(nams: &mut [Nam], nam_details: &mut NamDetails, rng: &mut Rng) {
```
marcelm: I’m not so keen on having the `&mut NamDetails` parameter: It exposes too much. The function could do whatever it wants to do with the `nam_details` object, but the only thing it is supposed to do is to update `time_sort_nams`. If you remove the parameter and let the function return the elapsed time instead, the "surface" becomes much smaller.

src/maponly.rs
```diff
@@ -106,9 +108,6 @@ fn paf_record_from_nam(
     }
 }
 
-/// Map a paired-end read pair to the reference and return PAF records
-///
-/// This implements "mapping-only" mode in which no base-level alignments are computed
```
marcelm, line 111: Why did you remove the comment?

src/maponly.rs
```diff
@@ -120,76 +119,72 @@ pub fn map_paired_end_read(
...
         insert_size_distribution.sigma,
+        &nam_details1,
+        &nam_details2,
     );
-    let mapped_nam =
-        get_best_paired_map_location(&nam_pairs, &nams1, &nams2, insert_size_distribution);
-    let mut records = vec![];
 
-    match mapped_nam {
-        MappedNams::Individual(nam1, nam2) => {
```
marcelm, line 151: You swapped the `MappedNams::Individual` and `MappedNams::Pair` cases, which makes review a bit harder. It would be easier to do this type of refactoring work separately (separate commit or separate PR).

src/maponly.rs
```diff
@@ -270,66 +254,205 @@ pub fn abundances_paired_end_read(
...
     }
 }
 
-enum MappedNams {
-    Individual(Option<Nam>, Option<Nam>),
-    Pair(Nam, Nam),
-    Unmapped,
+/// Represents either:
+/// A proper paired mapping (nam1, nam2, pairing score)
+/// Two independent best mappings (one per read)
```
marcelm: This is just a copy of what the doc comments for the individual fields already say. Can be removed IMO.

src/maponly.rs
```diff
@@ -270,66 +254,205 @@ pub fn abundances_paired_end_read(
...
+enum MappedNams<'a> {
+    /// A proper paired mapping (nam1, nam2, pairing score)
```
marcelm: I wonder about the "mapping" terminology that you introduce here and use also further down a couple of times. We haven’t used that term before AFAIK, and it doesn’t sound quite right.

If I understand correctly, this is just a "mapped read", so you could use that term. Or you call it a NAM because that is what it is. So I think I’d write "Two properly-paired NAMs" here. When we switch from the "NAM" to "chain" terminology (PR #575), this will then become "Two properly-paired chains", which also reads quite ok.

src/maponly.rs
```diff
@@ -270,66 +254,205 @@ pub fn abundances_paired_end_read(
...
-    nam_pairs: &[NamPair],
-    nams1: &[Nam],
-    nams2: &[Nam],
+/// Choose between:
+/// - the best proper pair of mappings
+/// - the best individual mappings
+///
+/// Also updates the insert size distribution using confident pairs.
+///
+/// For paired mapping and abundance only
```
marcelm:
```suggestion
/// For paired-end mapping and abundance estimation modes only
```

src/maponly.rs
```diff
@@ -270,66 +254,205 @@ pub fn abundances_paired_end_read(
...
+pub struct NamPair {
+    pub nam1: Nam,
+    pub nam2: Nam,
+    pub score: f64,
+}
+
+/// Build all plausible forward/revcomp mapping pairings
+fn get_nam_pairs(
+    nams1: &mut [Nam],
+    nams2: &mut [Nam],
```
marcelm, line 316: I understand why you did it this way and think it’s ok to leave this as is for now, but it’s a bit annoying that these two arguments have to be mutable. It would be much nicer to have no side effects, especially given that the function name starts with `get_`.

Maybe the proper solution would be to change the chainer so that it produces two separate lists of NAMs/chains from the start (forward and revcomp). It seems this is necessary anyway if we switch to noncanonical syncmers. But that would be something for later.

src/maponly.rs
```diff
@@ -270,66 +254,205 @@ pub fn abundances_paired_end_read(
...
 
-        MappedNams::Pair(best.0.unwrap(), best.1.unwrap())
-    } else {
-        MappedNams::Individual(best_individual_nam1.cloned(), best_individual_nam2.cloned())
+        // Scan window of revcomp nams within distance limit.
+        let mut best: Option<(usize, f32)> = None;
+        let mut i = rev_ptr;
+        while i < rev.len()
+            && rev[i].ref_id == f.ref_id
+            && (rev[i].projected_ref_start() - f.projected_ref_start()) as f32 <= max_dist
```
marcelm: You could also convert `max_dist` to an integer once, then you don’t need to convert to `f32` here in every iteration.

marcelm (commented):
As discussed yesterday, I’m submitting my partial review, waiting for further updates before I continue.

src/maponly.rs
```diff
@@ -270,66 +240,219 @@ pub fn abundances_paired_end_read(
...
+        // Scan window of revcomp nams within distance limit.
+        let mut best: Option<(usize, f32)> = None;
+        let mut i = rev_ptr;
+        while i < rev.len()
+            && rev[i].ref_id == f.ref_id
+            && (rev[i].projected_ref_start() - f.projected_ref_start()) <= max_dist
+        {
+            let s = rev[i].score;
+            if best.is_none_or(|(_, bs)| s > bs) {
+                best = Some((i, s));
```
marcelm: I suggest to rename `s` to `score` and `bs` to `best_score`.

src/maponly.rs
```diff
@@ -270,66 +240,219 @@ pub fn abundances_paired_end_read(
...
+            }
+            i += 1;
+        }
+
+        // Best scoring candidate
```
marcelm:
```suggestion
        // Highest-scoring candidate
```

src/maponly.rs
```diff
@@ -270,66 +240,219 @@ pub fn abundances_paired_end_read(
...
-        MappedNams::Individual(best_individual_nam1.cloned(), best_individual_nam2.cloned())
+        // Scan window of revcomp nams within distance limit.
+        let mut best: Option<(usize, f32)> = None;
+        let mut i = rev_ptr;
+        while i < rev.len()
+            && rev[i].ref_id == f.ref_id
+            && (rev[i].projected_ref_start() - f.projected_ref_start()) <= max_dist
+        {
+            let s = rev[i].score;
+            if best.is_none_or(|(_, bs)| s > bs) {
```
marcelm: Hm, I would have expected the bonus to be added already here so that the decision of which partner is best is made based also on the distance between the NAMs. (Could help as a tie breaker if two candidates have the same score.)

src/maponly.rs
```diff
@@ -270,66 +240,219 @@ pub fn abundances_paired_end_read(
...
+            i += 1;
+        }
+
+        // Best scoring candidate
+        let Some((best_id, best_score)) = best else {
+            continue;
+        };
+        let r = &rev[best_id];
+
+        // The pairing score get's a bonus based on the reference distance of the two chosen nam
```
marcelm:
```suggestion
        // The pairing score gets a bonus based on the reference distance of the two chosen nam
```

marcelm (commented):
I like this pairing algorithm better than what we had before, but strictly speaking, if I am not mistaken, asymptotic runtime is still quadratic since there are two nested loops in `find_pairs`. If all $n$ NAMs happen to be within the $10\sigma$ window, this will still run for $O(n^2)$ iterations.

But still, runtime is a bit better (also accuracy) and the concept is nicer, so looks good to me, just a few comments.

src/maponly.rs
NicolasBuchin: you are 100% right

marcelm: I ran a small benchmark just now with that change and it’s not a big difference in accuracy, but probably still worth it.

NicolasBuchin: There is a bug in the current pairing score on this branch, where it can be negative if the distribution is too far from the norm, since it's passed through a log function, so any value less than 1 will be negative. I changed it so it will always be positive. That might affect the results in accuracy more.

src/maponly.rs
marcelm: Nitpick: You left out the hyphen in your fix (it should be "Highest-scoring"). (Sorry, I cannot unsee this once I’ve noticed it ...)

src/maponly.rs
```diff
@@ -270,66 +240,219 @@ pub fn abundances_paired_end_read(
...
-                best.0
-                    .as_ref()
-                    .unwrap()
-                    .ref_start
-                    .abs_diff(best.1.as_ref().unwrap().ref_start),
-            );
+    nams.split_at_mut(left)
+}
+
+/// Find most forward/revcomp pairs using a linear two-pointer scan.
```
marcelm: Unless you can convince me that it actually is linear, can you remove the word "linear" here?

marcelm:
@ksahlin Are you fine with this PR?

src/nam.rs
marcelm, line 172: One more thing: I noticed that the trace output has now changed because the NAMs at this point no longer are sorted by score. Would it work to move this section of the code into `sort_nams()`?

NicolasBuchin: yeah absolutely, I'll make the change

ksahlin:
@marcelm I am fine with the PR on a conceptual level. I have been on vacation this week and have had minimal time to work. Unless you'd like me to review a specific part of the implementation in detail, I'm fine with a merge. @NicolasBuchin and I have had offline meetings about this PR and discussed issues like potentially buggy pairing scoring, where some valid pairs could get a negative "bonus" score. But as far as I understand, these issues have been addressed.

PR #573, opened 2026-03-20
Attempt to measure runtime quicker

marcelm:
This should probably not be merged, but I wanted to report this somewhere.

When profiling strobealign on my laptop, I noticed that measuring elapsed time (by calling `std::time::Instant::now()` and `std::time::Instant::elapsed()`) took up about 5% of the total runtime. On x86, it is possible to measure elapsed time using a special machine instruction (RDTSC) that reads out the TSC (Time Stamp Counter) register, which is very fast.

The [fastant crate](https://crates.io/crates/fastant) provides a drop-in replacement for `std::time::Instant` that measures time in that way.

This PR replaces all uses of `std::time::Instant` with `fastant::time::Instant`.

Now the weird part: On my laptop, this made strobealign about 5% faster, but on my desktop PC, I measure no difference at all (both have similar CPUs). Even the profiler output is clearly different.

`std::time::Instant::now()` [is documented to use `clock_gettime`](https://doc.rust-lang.org/std/time/struct.Instant.html#underlying-system-calls), which is part of the C library. My hypothesis is that on my desktop PC, `clock_gettime` actually uses `RDTSC` and is therefore fast, but that the laptop uses a slower implementation. I’ll need to check this when I have access to the laptop again.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -106,6 +106,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "bumpalo"
+version = "3.20.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5d20789868f4b01b2f2caec9f5c4e0213b41e3e5702a50157d699ae31ced2fcb"
+
+[[package]]
 name = "cc"
 version = "1.2.56"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


marcelm:
The speed difference is real and on the dataset I tested, it was responsible for an increase in runtime of 20%.

It turns out that my desktop PC and laptop use different "clock sources". This is a Linux kernel configuration thing. To check it, on my laptop I ran this command:
```
$ cat /sys/devices/system/clocksource/clocksource0/current_clocksource
hpet
```
HPET stands for "High Precision Event Timer", which apparently is a slower way to measure time. On my desktop PC, the same command prints `tsc` instead.

Looking at the logs, apparently my laptop booted with `tsc` as well and then after some time, it somehow detected that tsc was "unstable" (whatever that means) and then switched to hpet instead.

I think we don’t need to fix this in strobealign. The correct fix is to configure the correct clocksource (which I have now done for my laptop). The one thing we could possibly do is to detect (in strobealign), if the clocksource is currently `hpet` and print a warning.

ksahlin:
I saw your other PR addressing this. Great!

PR #574, opened 2026-03-22
Determine anchor orientation prior to chaining

Itolstoganov:
This stores a bool in `Syncmer` which is true iff the syncmer is canonical. These canonicity bits are then included in the randstrobe hash and used to filter out hits with a different orientation than the reference. The new randstrobe hash layout 

<strobe 1 hash><strobe 1 canonicity bit><strobe 2 hash><strobe 2 canonicity bit><offset>

Accuracy is the same as in main, the runtime is also mostly the same (slightly faster for sim6).

[ends.pdf](https://github.com/user-attachments/files/26169966/ends.pdf)

The main benefit is the removal of spurious anchors and the resulting small chains.

[chain-stats.pdf](https://github.com/user-attachments/files/26169997/chain-stats.pdf)

### DIFF
--- a/src/chainer.rs
+++ b/src/chainer.rs
@@ -356,7 +356,7 @@ fn hits_to_anchors(hits: &Vec<Hit>, index: &StrobemerIndex) -> Vec<Anchor> {
                 hit.query_start,
                 hit.query_end,
                 index,
-                hit.position
+                hit.position,
             );
         }
     }
--- a/src/chainer.rs
+++ b/src/chainer.rs
@@ -322,7 +322,7 @@ fn add_to_anchors_partial(
 ) {
     let forward_hash = index.get_hash_partial_forward(position);
     for pos in position..index.randstrobes.len() {
-        //Filter out partial lookups with different orientation
+        // Filter out partial lookups with different orientation
         if index.get_hash_partial_forward(pos) != forward_hash {
             break;
         }
--- a/src/chainer.rs
+++ b/src/chainer.rs
@@ -290,7 +290,7 @@ fn add_to_anchors_full(
 ) {
     let mut min_length_diff = usize::MAX;
     let canonical_hash = index.randstrobes[position].canonical_hash();
-    for randstrobe in index.randstrobes[position..].iter() {
+    for randstrobe in &index.randstrobes[position..] {
         if randstrobe.canonical_hash() != canonical_hash {
             break;
         }
@@ -319,10 +319,9 @@ fn add_to_anchors_partial(
     query_start: usize,
     index: &StrobemerIndex,
     position: usize,
-    query_hash: u64,
+    query_canonicity: u8,
 ) {
     let hash = index.get_hash_partial(position);
-    let query_canonicity = index.query_canonicity_partial(query_hash);
     for pos in position..index.randstrobes.len() {
         if index.get_hash_partial(pos) != hash {
             break;
@@ -350,7 +349,7 @@ fn hits_to_anchors(hits: &Vec<Hit>, index: &StrobemerIndex) -> Vec<Anchor> {
             continue;
         }
         if hit.is_partial {
-            add_to_anchors_partial(&mut anchors, hit.query_start, index, hit.position, hit.hash);
+            add_to_anchors_partial(&mut anchors, hit.query_start, index, hit.position, hit.query_canonicity);
         } else {
             add_to_anchors_full(
                 &mut anchors,
--- a/src/chainer.rs
+++ b/src/chainer.rs
@@ -344,8 +344,8 @@ fn hits_to_anchors(hits: &Vec<Hit>, index: &StrobemerIndex) -> Vec<Anchor> {
             continue;
         }
         if hit.is_partial {
-            if let Some(position) = hit.forward_position {
-                add_to_anchors_partial(&mut anchors, hit.query_start, index, position);
+            if let Some(forward_position) = index.get_partial_forward_from(hit.hash, hit.position) {
+                add_to_anchors_partial(&mut anchors, hit.query_start, index, forward_position);
             }
         } else {
             add_to_anchors_full(
@@ -353,7 +353,7 @@ fn hits_to_anchors(hits: &Vec<Hit>, index: &StrobemerIndex) -> Vec<Anchor> {
                 hit.query_start,
                 hit.query_end,
                 index,
-                hit.forward_position.unwrap(),
+                hit.position,
             );
         }
     }
--- a/src/chainer.rs
+++ b/src/chainer.rs
@@ -291,30 +291,29 @@ fn add_to_anchors_full(
     query_canonicity: u8,
 ) {
     let mut min_length_diff = usize::MAX;
-    let canonical_query_hash = index.randstrobes[position].hash() | ((query_canonicity as u64) << 8);
-    let start_pos = index.find_canonical(canonical_query_hash, position, forward_count)
-        .unwrap_or(position);
-    let canonical_hash = index.randstrobes[start_pos].canonical_hash();
-    for randstrobe in &index.randstrobes[start_pos..] {
-        if randstrobe.canonical_hash() != canonical_hash {
-            break;
-        }
-        let ref_start = randstrobe.position();
-        let ref_end = ref_start + randstrobe.strobe2_offset() + index.k();
-        let length_diff = (query_end - query_start).abs_diff(ref_end - ref_start);
-        if length_diff <= min_length_diff {
-            let ref_id = randstrobe.reference_index();
-            anchors.push(Anchor {
-                ref_id,
-                ref_start,
-                query_start,
-            });
-            anchors.push(Anchor {
-                ref_id,
-                ref_start: ref_end - index.k(),
-                query_start: query_end - index.k(),
-            });
-            min_length_diff = length_diff;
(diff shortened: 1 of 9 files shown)


ksahlin:
Very nice. Looks like it could solve problems with inconsistent NAMs (chains) donwstream -- similar to the non-canonical syncmers idea. This approach still preserves the same syncmers being generated, so more consistency in scoring chains between different directions compared to non-canonical seeds that can be different in abundance and in quantity between strands.

About the commit: The description/code should be changed so that "canonical" is replaced with "forward" in most places. All the syncmers are canonical, but some are forward and some reverse w.r.t. the sequence. So the bit is really keeping whether the seed is forward or not.

Itolstoganov:
> About the commit: The description/code should be changed so that "canonical" is replaced with "forward" in most places. All the syncmers are canonical, but some are forward and some reverse w.r.t. the sequence. So the bit is really keeping whether the seed is forward or not.

Depends on how you look at it? My logic was that we only store forward syncmers from the sequence and hashes of their canonical versions, so all syncmers are forward and some of them are canonical. This seems more intuitive to me since the position field of the `Syncmer` always refers to the forward version.

This only concerns the syncmers code, in the context of the index, "forward"/"unoriented" is used instead of "canonical".

src/seeding/strobes.rs
```diff
@@ -81,8 +100,10 @@ impl Randstrobe {
     /// the main hash and the bottom bits to the bits of the auxiliary
     /// hash. Since entries in the index are sorted by randstrobe hash, this allows
     /// us to search for the main syncmer by masking out the lower bits.
-    pub fn hash(hash1: u64, hash2: u64, main_hash_mask: u64) -> u64 {
-        ((hash1 & main_hash_mask) | (hash2 & !main_hash_mask)) & REF_RANDSTROBE_HASH_MASK
+    /// The orientation bit position (between main and aux) is left as zero.
+    pub fn hash(hash1: u64, hash2: u64, parameters: &RandstrobeParameters) -> u64 {
+        ((hash1 & parameters.main_hash_mask) ^ (hash2 & !parameters.forward_main_hash_mask))
```
marcelm: I’m happy to be convinced otherwise, but in my opinion, setting bits should use OR. I was a bit confused initially because XOR is used here, which to me implies toggling bits. The result is of course the same if you start with 0, but the apparent intention is a bit different. Leaving `is_forward1 ^ 1` and `is_forward2 ^ 1` as-is and changing the other XOR to OR would make it slightly clearer that the value returned from `hash()` has the orientation bits set to zero.

src/seeding/strobes.rs
```diff
@@ -14,6 +14,12 @@ pub struct RandstrobeParameters {
 
     /// Mask for bits of the hash that represent the main hash
     pub main_hash_mask: u64,
+
+    /// Mask for the main hash including its orientation bit
+    pub forward_main_hash_mask: u64,
```
marcelm, line 19: Can you have a comment somewhere that explains the various masks and other bit-fiddling related values? I had to draw something on paper in order to understand.

Maybe something like in your PR description where you used the `<strobe1_hash> ...` notation, but also with `forward_main_hash_mask`, `main_hash_mask`, `partial_orientation_pos` and `STROBE2_OFFSET_BITS`.

src/seeding/strobes.rs
```diff
@@ -81,8 +100,10 @@ impl Randstrobe {
     /// the main hash and the bottom bits to the bits of the auxiliary
     /// hash. Since entries in the index are sorted by randstrobe hash, this allows
     /// us to search for the main syncmer by masking out the lower bits.
-    pub fn hash(hash1: u64, hash2: u64, main_hash_mask: u64) -> u64 {
-        ((hash1 & main_hash_mask) | (hash2 & !main_hash_mask)) & REF_RANDSTROBE_HASH_MASK
+    /// The orientation bit position (between main and aux) is left as zero.
+    pub fn hash(hash1: u64, hash2: u64, parameters: &RandstrobeParameters) -> u64 {
```
marcelm: Can you try changing the `hash()` function such that it gets the `is_forward1` and `is_forward2` bools as parameters? Then the `from_randstrobes` function could be simplified somewhat.

src/seeding/syncmers.rs
```diff
@@ -6,10 +6,25 @@ use super::hash::xxh64;
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct Syncmer {
-    pub hash: u64,
+    hash: u64,
+    forward: bool,
```
marcelm: It would probably make sense to name this `is_forward` (like the function).

src/chainer.rs
```diff
@@ -320,9 +320,10 @@ fn add_to_anchors_partial(
     index: &StrobemerIndex,
     position: usize,
 ) {
-    let hash = index.get_hash_partial(position);
+    let forward_hash = index.get_hash_partial_forward(position);
     for pos in position..index.randstrobes.len() {
-        if index.get_hash_partial(pos) != hash {
+        //Filter out partial lookups with different orientation
```
marcelm:
```suggestion
        // Filter out partial lookups with different orientation
```

src/hit.rs
```diff
@@ -127,40 +129,43 @@ fn find_all_hits(
...
                     index.is_too_frequent(position, filter_cutoff, randstrobe.hash_revcomp);
                 if is_filtered {
                     hits_details.full_filtered += 1;
                 } else {
                     hits_details.full_found += 1;
                 }
-
                 let hit = Hit {
-                    position,
+                    position: position,
```
marcelm:
```suggestion
                    position,
```

marcelm (commented):
Hi, I think I’ll have to have another look later, but I left a few initial comments. Can you ensure the tests pass? You need to add a commit with the `Is-new-baseline: yes` trailer.

Overall, this adds a bit of complexity, and we need to find a way to manage this a bit better. I’m getting confused myself with all the `get_` functions in the `Index` now. I may have an idea for refactoring, but that would be something for later. Right now, I think documenting in one place a bit better which different things are stored in `hash_offset` and how the masks relate to it would be good.

src/seeding/strobes.rs
Itolstoganov: Makes sense, I have added a comment before the `RefRandstrobe::hash_offset` field

Itolstoganov:
> Hi, I think I’ll have to have another look later, but I left a few initial comments. Can you ensure the tests pass? You need to add a commit with the `Is-new-baseline: yes` trailer.

Hi! I`ve updated the baseline and accepted the comments.

> Overall, this adds a bit of complexity, and we need to find a way to manage this a bit better. I’m getting confused myself with all the `get_` functions in the `Index` now. I may have an idea for refactoring, but that would be something for later. Right now, I think documenting in one place a bit better which different things are stored in `hash_offset` and how the masks relate to it would be good.

Added some documentation to the `RefRandstrobe`, hopefully the layout is clearer now.

src/seeding/strobes.rs
marcelm: Thanks, very nice!

marcelm:
Looks good to me now! @ksahlin Has your comment about the terminology (forward vs canonical) been addressed?

marcelm (approved):

ksahlin:
> Looks good to me now! @ksahlin Has your comment about the terminology (forward vs canonical) been addressed?

Yes, it has been addressed. All good on my end.

src/seeding/strobes.rs
```diff
@@ -81,8 +108,19 @@ impl Randstrobe {
...
+        hash2: u64,
+        is_forward1: u64,
+        is_forward2: u64,
+        parameters: &RandstrobeParameters,
+    ) -> u64 {
+        ((hash1 & parameters.main_hash_mask)
+            | (hash2 & !parameters.forward_main_hash_mask)
+            | (is_forward1 << parameters.partial_orientation_pos)
+            | (is_forward2 << STROBE2_OFFSET_BITS))
+            & (REF_RANDSTROBE_HASH_MASK << 1)
```
marcelm, line 123: Hi @Itolstoganov, I’m currently (for reasons not related to this PR) working on this code, and I was wondering whether I’m misunderstanding something. The comment says "The orientation bit position (between main and aux) is left as zero." Is this correct? Because the intention of the code seems to be to set all four elements of the hash:
- Forward hash without orientation
- Orientation bit of forward hash according to `is_forward1`
- Aux hash without orientation
- Orientation bit of aux hash

But then there’s the `& (REF_RANDSTROBE_HASH_MASK << 1)`, which *unsets* the aux hash orientation if I understand correctly.

Is this just buggy?

marcelm: Hm, maybe the `& (REF_RANDSTROBE_HASH_MASK << 1)` is just in the wrong spot? Maybe like this:

```suggestion
    /// The orientation bit position (between main and aux) is left as zero.
    pub fn hash(
        hash1: u64,
        hash2: u64,
        is_forward1: u64,
        is_forward2: u64,
        parameters: &RandstrobeParameters,
    ) -> u64 {
        (hash1 & parameters.main_hash_mask)
            | (hash2 & !parameters.forward_main_hash_mask & (REF_RANDSTROBE_HASH_MASK << 1))
            | (is_forward1 << parameters.partial_orientation_pos)
            | (is_forward2 << STROBE2_OFFSET_BITS)
```

PR #575, opened 2026-03-23
Rename NAM to chain

marcelm:
This is essentially a search-and-replace job that switchs from "NAM" to "chain" terminology. In particular, the `Nam` struct is now `Chain` (appropriate because it contains a list of anchors).

- [x] Wait until #565 is merged and rebase (to avoid conflicts)


marcelm:
I’ve rebased this onto main. This will lead to conflicts in the other branches, but it’s probably better to do this now than wait even longer.

PR #576, opened 2026-03-31
New pairing of chains for extension mode

NicolasBuchin:
An addition to the new pairing of chains method, this time used for the extension algorithm.

I rewrote a lot of the paired-end extension code, not just to include the new pairing, but also to have a easier to maintain code. These changes include a new `pairing.rs` file containing pairing of chains/alignments. But the logic stays the same.

I also did some small changes that I believe should be done but we could decide to do them in another PR:
- not more `n_matches` in the Nam struct, as it is redundant information we already have access to with `nam.anchors.len()`.
- a new `shuffle.rs`, with a function that can be used to "shuffle" the best items for any vector, since this is a recurring pattern in this paired-end mode.
- added a counter of best alignments obtained from rescued alignments, this is an interesting statistic and something we should try to minimize.
- fallback to unpaired chains with rescued alignments using the mate only as a last resort.

I have some results on the accuracy and runtime for this PR: [ends.pdf](https://github.com/user-attachments/files/26374530/ends.pdf). Since I ran these on my laptop, on different days, I don't think we should trust the runtime too much.

Even if the results don't show improvements, I am in favor of adopting this pairing method, and this refactored paired-end logic.

I believe the paired-end extension code is a bit nicer than it used to be, but I also think there's room for improvement, looking forward to your suggestions!

### DIFF
--- a/src/chainer.rs
+++ b/src/chainer.rs
@@ -392,7 +392,6 @@ fn extract_chains_from_dp(
         }
 
         let mut j = i;
-        let mut c = 1;
         let mut overlaps = false;
         let mut chain_anchors = vec![anchors[i]];
 
@@ -407,7 +406,6 @@ fn extract_chains_from_dp(
             }
             chain_anchors.push(anchors[j]);
             used[j] = true;
-            c += 1;
 
             matching_bases += ref_coverage.saturating_sub(anchors[j].ref_start).min(k);
             ref_coverage = anchors[j].ref_start;
@@ -426,10 +424,9 @@ fn extract_chains_from_dp(
             query_end: last.query_start + k,
             ref_start: first.ref_start,
             ref_end: last.ref_start + k,
-            n_matches: c,
             matching_bases,
             ref_id: last.ref_id,
-            score: score + c as f32 * chaining_parameters.matches_weight,
+            score: score + chain_anchors.len() as f32 * chaining_parameters.matches_weight,
             is_revcomp,
             anchors: chain_anchors,
         });
--- a/src/details.rs
+++ b/src/details.rs
@@ -53,6 +53,9 @@ pub struct Details {
     /// No. of times rescue by local alignment was attempted
     pub mate_rescue: usize,
 
+    /// No. of best alignments found using mate rescue
+    pub best_rescued: usize,
+
     /// No. of computed alignments (get_alignment or rescue_mate)
     pub tried_alignment: usize,
 
@@ -70,6 +73,7 @@ impl ops::AddAssign<Details> for Details {
         self.nam += rhs.nam;
         self.inconsistent_nams += rhs.inconsistent_nams;
         self.mate_rescue += rhs.mate_rescue;
+        self.best_rescued += rhs.best_rescued;
         self.tried_alignment += rhs.tried_alignment;
         self.gapped += rhs.gapped;
         self.best_alignments += rhs.best_alignments;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -16,4 +16,5 @@ pub mod piecewisealigner;
 pub mod read;
 pub mod revcomp;
 pub mod seeding;
+pub mod shuffle;
 pub mod ssw;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -11,6 +11,7 @@ pub mod mapper;
 pub mod math;
 pub mod mcsstrategy;
 pub mod nam;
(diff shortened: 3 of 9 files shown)


marcelm:
> I also did some small changes that I believe should be done but we could decide to do them in another PR:
> 
> * not more `n_matches` in the Nam struct, as it is redundant information we already have access to with `nam.anchors.len()`.
> 
> * a new `shuffle.rs`, with a function that can be used to "shuffle" the best items for any vector, since this is a recurring pattern in this paired-end mode.
> 
> * added a counter of best alignments obtained from rescued alignments, this is an interesting statistic and something we should try to minimize.
> 
> * fallback to unpaired chains with rescued alignments using the mate only as a last resort.

It’s fine to have this in this PR, but every bullet point above should be its own commit with the text of the bullet point being the commit message.

The baseline comparison commit fails at the moment. It seems that the output is no longer deterministic (that is, repeated runs don’t give the same output), can you look into that?

NicolasBuchin:
> It’s fine to have this in this PR, but every bullet point above should be its own commit with the text of the bullet point being the commit message.

You're right, I'll make those changes into their own commits.

> The baseline comparison commit fails at the moment. It seems that the output is no longer deterministic (that is, repeated runs don’t give the same output), can you look into that?

I fixed the non-deterministic issue, it happens that because hash tables in rust use random hash functions, collecting the alignments from the tables will be in a random order. Even if we sort, the best scoring ones will be in a random order. I think the best option is to use a fixed hash hash-table, that way we keep everything the same, we just need one extra dependency.

marcelm:
> I think the best option is to use a fixed hash hash-table, that way we keep everything the same, we just need one extra dependency.

We had the same problem in the old, now removed NAM code. This is what I used there. It does not require adding another dependency:
https://github.com/ksahlin/strobealign/blob/0ce640fe36eedf02f7e7a4855255871c5caf21f0/src/nam.rs#L66-L67

And a small note regarding the `Is-new-baseline: yes` line in the commit message: This is intended to be a [Git trailer](https://git-scm.com/docs/git-interpret-trailers) and for that, it needs to be preceded by a blank line. Our baseline comparison script happens to recognize the line even without the blank line, but we shouldn’t rely on that.

NicolasBuchin:
> We had the same problem in the old, now removed NAM code. This is what I used there. It does not require adding another dependency:
> 
> https://github.com/ksahlin/strobealign/blob/0ce640fe36eedf02f7e7a4855255871c5caf21f0/src/nam.rs#L66-L67

Good catch! But I also realized that the "caching" of alignments is useless now, we retrieve pairs of unique chains, we will never align the same nam twice. We can just use vectors to keep the alignments.

> And a small note regarding the `Is-new-baseline: yes` line in the commit message: This is intended to be a [Git trailer](https://git-scm.com/docs/git-interpret-trailers) and for that, it needs to be preceded by a blank line. Our baseline comparison script happens to recognize the line even without the blank line, but we shouldn’t rely on that.

I'll make sure to use it properly from now on.

I also separated the other changes of this PR into their own commit, these might not compile as I just edited theses changes into their own commits.

marcelm:
This is a bit hard to review because so much code was shuffled around. In particular, commit 80639d35a846fdb0ed2f6b5d10286d53265df524 mixes changing of the pairing algorithm with a refactoring where some code is moved to `pairing.rs`. For reviewing purposes, it really helps if a single commit only does one, limited thing, like the other commits in this PR. As it is now, there are >800 deleted lines in `mapper.rs` and >800 added lines in `pairing.rs`, and I cannot tell where the algorithmic changes are. (Try looking at the "Files changed" tab to get an impression.)

I know it is hard to make small, self-contained commits, at least I find it difficult because it is so easy to do something unrelated that I happen to notice while doing something else, but it is crucial for reviews.

Note that if refactoring moves a large chunk of code around, that is fine because I can trust that the tests will catch any accidental changes in behavior. But if a commit is large and has a `Is-new-baseline: yes` trailer, it’s hard to find the actual thing that changed.

I don’t think I want to ask you to split up this commit unless you think you could do it without too much effort. So I will probably, instead of reviewing the PR itself, just have a brief look through `pairing.rs` and run the evaluation.

NicolasBuchin:
I think it's probably best to just look at pairing.rs, and the benchmarks in this case. I agree with your points though, I will keep in mind for the future.

marcelm:
Let’s close this in favor of #584.

PR #583, opened 2026-04-21
Refactor: Move index creation to a new indexer module

marcelm:
The index.rs file was a bit large, so this splits up index creation and index querying.

Also, it was a bit sketchy that to create an empty StrobemerIndex, one had to do `index = StrobemerIndex::new()` followed by `index.populate()`, so in between those steps, you had a half-initialized index. Now, there is only a `make_index()` function that returns a fully populated index.

### DIFF
--- a/src/chainer.rs
+++ b/src/chainer.rs
@@ -165,7 +165,7 @@ impl Chainer {
                 &query_randstrobes[is_revcomp],
                 index,
                 mcs_strategy,
-                index.filter_cutoff,
+                index.filter_cutoff(),
                 rescue_distance,
             );
         }
--- a/src/index.rs
+++ b/src/index.rs
@@ -4,7 +4,6 @@ use std::path::Path;
 
 use thiserror::Error;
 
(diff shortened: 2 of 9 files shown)


ksahlin:
Makes sense, nice!

marcelm:
I added a few more refactoring commits that remove unused attributes of the `StrobemerIndex` struct.

One question: It turns out that the `partial_cutoff` attribute is never used. It appears that the `is_too_frequent_partial` function just gets called with the regular `filter_cutoff`. We have always set the `partial_cutoff` identical to `filter_cutoff`, so this doesn’t really matter. Is it ok if I just remove `partial_cutoff`? The original intention was to be able to have two different cutoffs for the full and partial cases, but we can always re-introduce this if we decide it makes sense. At the moment, it just makes the code a bit harder to follow than necessary.

ksahlin:
Sure, I am okay with this. I know @Itolstoganov is looking into this cutoff (or something related) currently, but he is on a different branch anyway. But just as an FYI (to both of you).

PR #586, opened 2026-05-11
Some refactoring

marcelm:
Since it’s a bit unclear whether we are going to merge PR #576/#584, I’ve extracted the first couple of commits from #584. These only do some refactoring that is I think an overall improvement and can be merged no matter what we decide on the main PRs.

### DIFF
--- a/src/chainer.rs
+++ b/src/chainer.rs
@@ -41,6 +41,13 @@ impl Default for ChainingParameters {
     }
 }
 
+#[derive(Debug, Default)]
+pub struct ChainingResult {
+    best_score: f32,
+    dp: Vec<f32>,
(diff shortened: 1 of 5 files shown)


marcelm:
I’ve added a couple of (refactoring) commits that I had in a different branch. These introduce a `ChainingResult` struct that is returned by the `Chainer`.

That is, instead of `colinear_chaining` returning a `(f32, Vec<f32>, Vec<usize>)` tuple, it returns a `ChainingResult`. Then the chains are computed with `ChainingResult::extract_chains` instead of a free-standing `extract_chains_from_dp` function.

The idea is that `ChainingResult` could become an iterator that yields chains one by one (but this is not yet implemented here).

PR #591, opened 2026-06-05
Do not expose the RefRandstrobe directly from IndexEntry

marcelm:
I forgot to push this as part of #588. No runtime and accuracy impact as before.


ksahlin:
👍

PR #592, opened 2026-06-05
Fix unmapped secondary alignments

marcelm:
When secondary alignments are enabled, we currently handle unmapped reads incorrectly. So for example, if we have a pair where R1 can be mapped and has multiple mappings but R2 cannot be mapped, we output these SAM records:
* R1 with primary mapping location
* R2 marked as unmapped
* R1 with alternative location, marked as secondary
* R2 as an unmapped read, *not marked as secondary* (this is the problematic record)

I first changed it so that the second time R2 is output it is also marked as secondary, but then we have a SAM record that is both unmapped and secondary, which `picard ValidateSamFile` rightfully complains about. The correct solution seems to be to just not output that last record, which is what this PR does.

Also, we now let strobealign generate secondary alignments within the baseline comparison script. They are ignored when actually comparing the records to the baseline version, but this allows us to run `ValidateSamFile` on a file with secondary alignments.

The last commit changes the baseline commit because running strobealign with `-N 5` and ignoring all secondary alignments does not give the same results as running it with `-N 0`.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -83,9 +83,10 @@ jobs:
         run: tests/compare-baseline.sh
       - name: Validate with Picard
         run: |
-          PicardCommandLine ValidateSamFile IGNORE=RECORD_MISSING_READ_GROUP IGNORE=MISSING_READ_GROUP I=head.bam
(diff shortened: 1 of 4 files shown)


ksahlin:
very good!

PR #595, opened 2026-06-08
Implement 2-bit representation of reference sequences.

### LINKED ISSUE #372: Use a 2-bit encoded reference
As discussed previously, we should try to switch to a 2-bit encoded reference to save memory. Any nucleotides that are not A, C, G, T would need to be mapped to a pseudo-randomly chosen nucleotide (as BWA does).

ksahlin:
Replaces Vec<u8> (1 byte/base) with PackedSeq (2 bits/base) for reference sequences. 

Memory savings as expected and runtime appears unaffected. 

1. FASTA is packed byte-by-byte during reading — no intermediate Vec<u8> per chromosome. 
2. SyncmerIterator is generic over a SeqAccess trait, so indexing iterates PackedSeq directly with no decode step -- base_bits() returns 2-bit values straight from the packed word. 
3. Alignment hot paths decode only the needed subrange on demand via PackedSeq::decode(start, end).

Currently, N/ambiguous bases map to A. Should we make this pseudo-random instead?

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -542,6 +542,7 @@ mod tests {
 
     use crate::indexer::make_index;
     use crate::io::fasta::{RefSequence, read_ref};
+    use crate::packed_seq::PackedSeq;
     use crate::revcomp::reverse_complement;
 
     #[test]
@@ -590,7 +591,7 @@ mod tests {
         let rc_seq = reverse_complement(&seq_decoded);
         let rc_references = vec![RefSequence {
             name: "phix_rc".to_string(),
-            sequence: crate::packed_seq::PackedSeq::from_slice(&rc_seq),
+            sequence: PackedSeq::from_slice(&rc_seq),
         }];
 
         let parameters = SeedingParameters::new(300);
--- a/src/index.rs
+++ b/src/index.rs
@@ -586,11 +586,11 @@ mod tests {
     #[test]
     fn test_partial_orientation() {
         let references = read_ref("tests/phix.fasta").unwrap();
-        let seq = &references[0].sequence;
-        let rc_seq = reverse_complement(seq);
+        let seq_decoded = references[0].sequence.decode_all();
+        let rc_seq = reverse_complement(&seq_decoded);
         let rc_references = vec![RefSequence {
             name: "phix_rc".to_string(),
-            sequence: rc_seq,
+            sequence: crate::packed_seq::PackedSeq::from_slice(&rc_seq),
         }];
 
         let parameters = SeedingParameters::new(300);
--- a/src/io/fasta.rs
+++ b/src/io/fasta.rs
@@ -126,7 +126,10 @@ pub fn read_fasta<R: BufRead>(reader: &mut R) -> Result<Vec<RefSequence>, Sequen
             Ok(_) => {
                 if line.starts_with('>') {
                     if let Some(name) = current_name.take() {
-                        records.push(RefSequence { name, sequence: current_seq });
+                        records.push(RefSequence {
+                            name,
+                            sequence: current_seq,
+                        });
                         current_seq = PackedSeq::new();
                     }
                     let (name, _comment) = split_header(&line[1..]);
@@ -145,7 +148,10 @@ pub fn read_fasta<R: BufRead>(reader: &mut R) -> Result<Vec<RefSequence>, Sequen
         }
     }
     if let Some(name) = current_name {
-        records.push(RefSequence { name, sequence: current_seq });
+        records.push(RefSequence {
+            name,
+            sequence: current_seq,
+        });
     }
 
     for record in &records {
--- a/src/io/fasta.rs
+++ b/src/io/fasta.rs
@@ -4,11 +4,12 @@ use std::path::Path;
 
 use crate::io::record::{End, SequenceRecord};
 use crate::io::{SequenceIOError, split_header};
+use crate::packed_seq::PackedSeq;
 
 #[derive(Debug, Clone)]
 pub struct RefSequence {
     pub name: String,
-    pub sequence: Vec<u8>,
+    pub sequence: PackedSeq,
 }
 
 /// Check whether a name is fine to use in SAM output.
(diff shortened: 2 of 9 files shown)


ksahlin:
This commit is now improved by mapping N to random nt. Memory savings as expected, and runtime appears unaffected. 


However, I got surprised by the quite large amount of reads that changed locations in our drosophila dataset:

```
      11558 reads were unmapped before and after
*         2 reads became mapped
*         1 reads became unmapped
     174109 reads were mapped to same locus before and after
*     10015 reads were multimapper before and after, same alignment score (AS)
*        73 reads were multimapper before and after, better alignment score (AS)
*        72 reads were multimapper before and after, worse alignment score (AS)
*      4170 reads changed in another way
     200000 total reads
```

Out of 4170 that "changed in another way" was:

3,940 (94.5%) — same AS, but mapped to a different but equally-scored location
120 (2.9%) — better AS in 2bit
110 (2.6%) — worse AS in 2bit

I also tried the maize genome with 3,807,886 N bases and again the output differ:

```
Before/after comparisons

        452 reads were unmapped before and after
          0 reads became mapped
*         1 reads became unmapped
    1953046 reads were mapped to same locus before and after
*     41630 reads were multimapper before and after, same alignment score (AS)
*       462 reads were multimapper before and after, better alignment score (AS)
*       469 reads were multimapper before and after, worse alignment score (AS)
*      3940 reads changed in another way
    2000000 total reads
```

However, when I map 2M PE reads to CHM13 genome (without Ns), the SAM files produced are *identical* between the two versions. I checked with both `samdiff.py` and `diff` from the terminal. 

So it seems that N -> random base yields different alignments, which is expected in theory, but I had no idea it would be this different. Are the RNG if a reads is multimapper not only using the read header as seed? 

That being said, the quality of the mappings does not appear either worse or better.

So if @marcelm is okay with the introduced code(?) i would be happy to try this. If the code structure does not meet standards, it could at least serve as inspiration.

________ 

On a related note: `samdiff.py` infers multimapper from the MAPQ score. strobealign has a known issue of not always reporting a reliable mapq score (e.g., in above probably the 3,940 who had same AS should have been a multimapper).

So I might look into this next.

src/io/fasta.rs
```diff
@@ -113,18 +114,42 @@ impl<B: BufRead> Iterator for FastaReader<B> {
...
-    let fasta_reader = FastaReader::new(reader);
-    let records = fasta_reader
-        .map(|result| {
-            result.map(|record| RefSequence {
-                name: record.name,
-                sequence: record.sequence,
-            })
-        })
-        .collect::<Result<Vec<RefSequence>, SequenceIOError>>()?;
+    let mut records: Vec<RefSequence> = Vec::new();
```
marcelm, line 117: Since `read_fasta` no longer is implemented in terms of `FastaReader`, it needs separate tests

src/index.rs
```diff
@@ -586,11 +586,11 @@ mod tests {
...
     fn test_partial_orientation() {
         let references = read_ref("tests/phix.fasta").unwrap();
-        let seq = &references[0].sequence;
-        let rc_seq = reverse_complement(seq);
+        let seq_decoded = references[0].sequence.decode_all();
+        let rc_seq = reverse_complement(&seq_decoded);
         let rc_references = vec![RefSequence {
             name: "phix_rc".to_string(),
-            sequence: rc_seq,
+            sequence: crate::packed_seq::PackedSeq::from_slice(&rc_seq),
```
marcelm: Should be `PackedSeq::from_slice` and then an appropriate `use` declaration within `mod test` should be used.

src/seeding/syncmers.rs
```diff
@@ -57,16 +82,16 @@ impl SyncmerParameters {
...
+    seq: S,
     k: usize,
     s: usize,
     t: usize,
     kmask: u64,
     smask: u64,
     kshift: usize,
     sshift: usize,
-    qs: VecDeque<u64>, // s-mer hashes
+    qs: VecDeque<u64>,
```
marcelm: Any idea why the comment was removed?

src/seeding/syncmers.rs
```diff
@@ -3,6 +3,31 @@ use std::collections::VecDeque;
...
 use super::hash::xxh64;
+use crate::packed_seq::PackedSeq;
+
+/// Trait for types that can serve as a sequence source for `SyncmerIterator`.
+///
+/// `base_bits` returns the nucleotide as a 2-bit value (0=A, 1=C, 2=G, 3=T,
+/// 4=N/ambiguous). This is what the syncmer hash machinery needs directly,
+/// so implementations should avoid going through ASCII as an intermediate.
+pub trait SeqAccess {
+    fn base_bits(&self, i: usize) -> u8;
```
marcelm: I was confused by the term `base` at first because I thought it had something to do with the base of a number system. I suggest to name this `nucleotide_bits`.

src/seeding/syncmers.rs
```diff
@@ -3,6 +3,31 @@ use std::collections::VecDeque;
...
+    fn seq_len(&self) -> usize;
```
marcelm: This should IMO just be called `len`.

src/packed_seq.rs
```diff
@@ -0,0 +1,118 @@
...
+    t[b'A' as usize] = 0; t[b'a' as usize] = 0;
+    t[b'C' as usize] = 0; t[b'c' as usize] = 0;
+    t[b'G' as usize] = 0; t[b'g' as usize] = 0;
+    t[b'T' as usize] = 0; t[b't' as usize] = 0;
+    t[b'U' as usize] = 0; t[b'u' as usize] = 0;
+    t
+}
+
+static PACK: [u8; 256] = make_pack_table();
+static IS_AMBIG: [u8; 256] = make_is_ambig_table();
```
marcelm: Should IMO be called `IS_AMBIGUOUS` (or `IS_AMBIGUITY`?). Or maybe `IS_WILDCARD` is simpler.

src/packed_seq.rs
```diff
@@ -0,0 +1,118 @@
...
+static DECODE: [u8; 4] = [b'A', b'C', b'G', b'T'];
+
+/// Deterministic pseudo-random 2-bit value for position `pos`.
+///
+/// Uses a Fibonacci/Knuth multiplicative hash so every ambiguous base at a
+/// distinct position maps to a distinct-looking 2-bit value, making syncmers
+/// that span N runs have effectively random hashes (following BWA's approach).
+#[inline]
+fn ambig_bits(pos: usize) -> u64 {
```
marcelm: Maybe `pack_wildcard`.

src/packed_seq.rs
```diff
@@ -0,0 +1,118 @@
...
+    let mut t = [1u8; 256];
+    t[b'A' as usize] = 0; t[b'a' as usize] = 0;
+    t[b'C' as usize] = 0; t[b'c' as usize] = 0;
+    t[b'G' as usize] = 0; t[b'g' as usize] = 0;
+    t[b'T' as usize] = 0; t[b't' as usize] = 0;
+    t[b'U' as usize] = 0; t[b'u' as usize] = 0;
+    t
+}
+
+static PACK: [u8; 256] = make_pack_table();
```
marcelm: I think this should be called `ENCODE` for consistency with `DECODE`.

src/packed_seq.rs
```diff
@@ -0,0 +1,118 @@
...
+static IS_AMBIG: [u8; 256] = make_is_ambig_table();
+static DECODE: [u8; 4] = [b'A', b'C', b'G', b'T'];
+
+/// Deterministic pseudo-random 2-bit value for position `pos`.
+///
+/// Uses a Fibonacci/Knuth multiplicative hash so every ambiguous base at a
+/// distinct position maps to a distinct-looking 2-bit value, making syncmers
```
marcelm:
```suggestion
/// distinct position maps to a random-looking 2-bit value, making syncmers
```

marcelm (commented):
This looks quite good! The slowdown is much less than what I had expected (hard to see a  slowdown actually).

One downside is that with this version, we generate randstrobes even for `N` regions, which means that the index will become bigger in many cases. For example, on GRCh38, the number of randstrobes goes up by 78 million (from 611409875 to 689205916) with this PR. This increases memory usage by ~1 GB. It’s still a memory reduction overall of course.

I think it would be relatively easy to just keep track of the `N` regions and ignore them when generating syncmers, but I think it’s ok to do this in a follow-up PR.

PR #596, opened 2026-06-09
Fixes bug in assigning MAPQ 60 to reads that are not proper pairs.

ksahlin:
Previously, the best non-proper pair would get a mapq of 60, regardless of whether it had multiple best scoring chains or not. This commit fixes the bug. 

If the best pair is not proper: check for each read if the second-best chain has the same score, if so, assign mapq 0, otherwise compute mapq from our formula. If there is no second chain, the read gets 60.

The fix is quite modular. But we may want to watch out for conflicts or to reintroduce the faulty behavior when merging PR
#584 (based on PR https://github.com/ksahlin/strobealign/pull/576) 

This issue was detected when I was running samdiff.py for the 2bit-ref-rs PR (see end of this comment https://github.com/ksahlin/strobealign/pull/595#issuecomment-4652957368), which had 4170 that had a MAPQ score of > 0 and logged under "changed in another way", but further assessment of those 4170 gave:

3,940 (94.5%) — same AS between versions, but mapped to a different but equally-scored location (should be MAPQ=0).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,8 @@
   select it. This is equivalent to `-k 16 -s 12 -l 2 -u 2 -m 100`.
   We found these settings to increase accuracy on error-prone reads at the cost
   of runtime.
+* #596: Fix not properly paired reads getting MAPQ 60 even if there were
+  alternative equally good mapping locations.
 
 ## v0.17.0 (2025-12-18)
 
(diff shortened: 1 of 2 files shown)


marcelm:
Nice, looks good! I have added a changelog entry and simplified the MAPQ computation by removing one nested `if`.

ksahlin:
great. I am ok to merge if you are.

PR #601, opened 2026-06-11
Dropoff based on chain score instead on nr of anchors.

ksahlin:
Makes more sense based on the approach of sorting NAMs (chains) based on score.

Some benchmarks on single-end reads:

## Single-end alignment mode (8 threads)

| Genome      | Len | Tool | Reads   | Mapped% | Acc%    | Δ Acc   | Runtime  |
|-------------|-----|------|--------:|--------:|--------:|--------:|------:|
| chrY        | 100 | main |  50,000 | 47.640% | 33.458% |         |  1.5s |
| chrY        | 100 | fix  |  50,000 | 47.944% | 33.572% | +0.114% |  1.5s |
| chrY        | 250 | main |  50,000 | 52.776% | 38.712% |         |  1.5s |
| chrY        | 250 | fix  |  50,000 | 52.144% | 39.084% | +0.372% |  1.5s |
| CHM13 full  | 100 | main | 100,000 | 92.157% | 89.965% |         | 17.1s |
| CHM13 full  | 100 | fix  | 100,000 | 92.133% | 89.984% | +0.019% | 17.1s |
| CHM13 full  | 250 | main | 100,000 | 93.926% | 92.653% |         | 18.7s |
| CHM13 full  | 250 | fix  | 100,000 | 93.877% | 92.664% | +0.011% | 18.7s |
| Maize       | 100 | main |  50,000 | 75.940% | 67.474% |         | 10.9s |
| Maize       | 100 | fix  |  50,000 | 75.860% | 67.574% | +0.100% | 10.9s |
| Maize       | 250 | main |  50,000 | 89.550% | 87.722% |         | 11.1s |
| Maize       | 250 | fix  |  50,000 | 89.454% | 87.722% | +0.000% | 11.2s |

## Single-end mapping mode (-x, 8 threads)

Identical results for main and fix on all datasets — the fix only applies
to the alignment code path and does not affect mapping mode.

PR #603, opened 2026-06-12
Also adds the score dropoff to PE path

ksahlin:
PR #601 fixed only for the SE path, plus for rescue function in PE mode. This PR completes the changes to also use score-based dropoff for the pairing of NAMs(/chains).

Neither runtime nor accuracy changes significantly (see accuracy experiment below). The fix is for consistency in approach.

## Paired-end alignment, 200k PE reads 

Genome | Len | Tool | Acc% | Δ vs main
-- | -- | -- | -- | --
chrY | 100 | main | 44.259% |  
chrY | 100 | score-dropoff | 44.279% | +0.020%
chrY | 250 | main | 55.867% |  
chrY | 250 | score-dropoff | 55.870% | +0.004%
chrY | 500 | main | 67.846% |  
chrY | 500 | score-dropoff | 67.842% | −0.004%
CHM13 | 100 | main | 93.525% |  
CHM13 | 100 | score-dropoff | 93.527% | +0.002%
CHM13 | 250 | main | 95.396% |  
CHM13 | 250 | score-dropoff | 95.396% | +0.000%
CHM13 | 500 | main | 96.919% |  
CHM13 | 500 | score-dropoff | 96.919% | +0.000%
Maize | 100 | main | 87.396% |  
Maize | 100 | score-dropoff | 87.378% | −0.018%
Maize | 250 | main | 95.942% |  
Maize | 250 | score-dropoff | 95.943% | +0.001%
Maize | 500 | main | 98.442% |  
Maize | 500 | score-dropoff | 98.444% | +0.002%

</body></html>

### DIFF
--- a/src/mapper.rs
+++ b/src/mapper.rs
@@ -1454,10 +1454,9 @@ where
 /// compute dropoff of the first (top) NAM
 fn top_dropoff(nams: &[Nam]) -> f32 {
     let n_max = &nams[0];
-    if n_max.anchors.len() <= 2 {
-        1.0
-    } else if nams.len() > 1 {
-        nams[1].anchors.len() as f32 / n_max.anchors.len() as f32
+    if nams.len() > 1 {
+        debug_assert!(n_max.score > 0.0);
+        nams[1].score / n_max.score
     } else {
         0.0
     }


src/mapper.rs
```diff
@@ -1433,10 +1433,10 @@ where
 /// compute dropoff of the first (top) NAM
 fn top_dropoff(nams: &[Nam]) -> f32 {
     let n_max = &nams[0];
-    if n_max.anchors.len() <= 2 {
+    if n_max.score <= 0.0 {
```
marcelm: As discussed just now: Does this ever apply? Probably remove - or use a  more sensible threshold.

ksahlin: A chain can never have a score below k (@NicolasBuchin ), so this conditional should be safe to remove.

marcelm:
I confirm that there is nearly no difference in accuracy, and no discernible difference in runtime.

PR #604, opened 2026-06-15
Introduce a threshold for local rescue (fixes maize T2T memory usage)

marcelm:
For the maize T2T genome, we have the problem of extremely repetitive hits, for example in regions with TAG repeats. The strobemers generated in these regions occur ~500000 times in the genome. Currently, when running local rescue, we just rescue everything, even this type of hit, which makes chaining of reads from these regions very slow and memory intensive (due to the number of anchors produced).

This PR makes it so that the threshold of 1000 we currently use for global rescue is also used for local rescue.

I am running our evaluation to see how this changes accuracy.

Issue found by @mahmudsami

### DIFF
--- a/src/hit.rs
+++ b/src/hit.rs
@@ -11,6 +11,10 @@ use crate::seeding::QueryRandstrobe;
 /// Number of hits to rescue globally if we have too few nonrepetitive hits
 const GLOBAL_RESCUE_COUNT: usize = 5;
 
+/// Hits with a count above this are always ignored (and never rescued)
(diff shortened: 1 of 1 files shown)


ksahlin:
Very good. I am okay with merging

marcelm:
Just got the benchmark results: accuracy drops quite a bit on chrY with this change. I’m testing a different threshold now.

marcelm:
I changed it so that there are two different thresholds: 1000 for global rescue (as before) and 10000 for local rescue. This no longer changes accuracy on our benchmark datasets (but still fixes the T2T maize issue.)

PR #605, opened 2026-06-16
Make max_ref_gap dynamic and add max_diagonal_ratio parameter

NicolasBuchin:
Previously, `max_ref_gap` was a fixed value (1000bp), which is too restrictive for long reads. It now defaults to the read length and can be overridden via --sg.

A new `max_diagonal_ratio` parameter (--mr, default 10.0) is added to prevent chaining a child anchor to a parent anchor far off the main diagonal, which would imply an unreasonably large indel relative to the distance between them. I discovered this through a blockaligner crash on global alignment of ~10×10000bp, which should never have been attempted in the first place. Because the `max_ref_gap` is now dynamic, it opens the door to very long gaps between anchors for long reads. Beyond fixing the crash, it also restricts the chaining search to a diagonal cone, avoiding unreasonable parent anchors even when the ref gap window is large.

Overall, accuracy on long reads in repetitive regions such as chrY improves significantly.
[ends-se-accuracy-table.pdf](https://github.com/user-attachments/files/29012489/ends-se-accuracy-table.pdf)

I am re-running my benchmark, so that we can examine runtime accurately. Will report when it's done.

### DIFF
--- a/src/chainer.rs
+++ b/src/chainer.rs
@@ -24,8 +24,9 @@ pub struct ChainingParameters {
     pub diag_diff_penalty: f32,
     pub gap_length_penalty: f32,
     pub valid_score_threshold: f32,
-    pub max_ref_gap: usize,
+    pub max_ref_gap: Option<usize>,
     pub matches_weight: f32,
+    pub max_diagonal_ratio: f32,
 }
 
 impl Default for ChainingParameters {
@@ -35,8 +36,9 @@ impl Default for ChainingParameters {
             diag_diff_penalty: 0.1,
             gap_length_penalty: 0.05,
             valid_score_threshold: 0.7,
-            max_ref_gap: 1000,
+            max_ref_gap: None,
             matches_weight: 0.01,
+            max_diagonal_ratio: 10.0,
         }
     }
 }
@@ -79,11 +81,14 @@ impl Chainer {
         }
     }
 
-    fn collinear_chaining(&self, anchors: Vec<Anchor>) -> ChainingResult {
+    fn collinear_chaining(&self, anchors: Vec<Anchor>, read_len: usize) -> ChainingResult {
         let n = anchors.len();
         if n == 0 {
             return ChainingResult::default();
         }
+
+        // dynamic maximum allowed ref gap based on read length
+        let max_ref_gap = self.parameters.max_ref_gap.map_or(read_len, |max| max);
 
         let mut dp = vec![self.k as f32; n];
         let mut predecessors = vec![usize::MAX; n];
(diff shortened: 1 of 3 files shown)


NicolasBuchin:
Here are the result from my benchark.
[ends.pdf](https://github.com/user-attachments/files/29021752/ends.pdf)

Looks like mapping might take longer for longer reads, but extension is same or faster. My guess is we can now create chains with more anchors for longer reads, therefore less time doing alignments with block aligner.

We can also see our previous chaining dropped in accuracy for very long reads (20kbps+) on chromosome Y.

ksahlin:
The mapping runtime increase only looks significant on chrY, and then the accompanying accuracy increase is worth it. The other datasets looks nearly unaffected, but this chaining change makes sense conceptually. I am okay to merge if @marcelm is.

src/chainer.rs
```diff
@@ -79,12 +81,15 @@ impl Chainer {
...
 
-    fn collinear_chaining(&self, anchors: Vec<Anchor>) -> ChainingResult {
+    fn collinear_chaining(&self, anchors: Vec<Anchor>, read_len: usize) -> ChainingResult {
         let n = anchors.len();
         if n == 0 {
             return ChainingResult::default();
         }
 
+        // dynamic maximum allowed ref gap based on read length
+        let max_ref_gap = self.parameters.max_ref_gap.map_or(read_len, |max| max);
```
marcelm:
```suggestion
        let max_ref_gap = self.parameters.max_ref_gap.unwrap_or(read_len);
```

src/main.rs
```diff
@@ -227,14 +227,19 @@ struct Args {
     #[arg(long = "vp", default_value_t = ChainingParameters::default().valid_score_threshold, help_heading = "Collinear chaining")]
     valid_score_threshold: f32,
 
-    /// Collinear chaining skip distance, how far on the reference do we allow anchors to chain
-    #[arg(long = "sg", default_value_t = ChainingParameters::default().max_ref_gap, help_heading = "Collinear chaining")]
-    max_ref_gap: usize,
+    /// Collinear chaining skip distance, how far on the reference do we allow anchors to chain. By
+    /// default will be the size of the read, but can be overriden if need be.
```
marcelm:
```suggestion
    /// Collinear chaining skip distance, how far on the reference do we allow anchors to chain [default: same as length of read]
```

src/chainer.rs
```diff
@@ -447,7 +462,8 @@ mod test {
             Anchor { ref_id: 0, ref_start: 95, query_start: 35, },
         ];
 
-        let chaining_result = chainer.collinear_chaining(anchors);
+        let chaining_result = chainer.collinear_chaining(anchors, 2000);
+        println!("{:?}", chaining_result);
```
marcelm:
```suggestion
```

marcelm (commented):
Looks good to me! I’ll apply some minor suggestions I have and then merge.

Issue #597, opened 2026-06-10
Ideas for reducing memory usage


marcelm:
We have discussed (variations of) this before (#227, #27 and others).

An entry in the index (`RefRandstrobe`) currently takes up 16 bytes:

what | bits
-|-:
hash | 56
strobe 2 offset | 8
reference index | 32
reference start | 32

There are a couple of ways in which this could be reduced:

- The first bits of the hash are also available in the `bucket_starts` vector (a "bucket" is a slice of the randstrobes table where all hashes share the same prefix). The number of bits used for the `bucket_starts` is determined automatically and can currently be as low as 9, but we could force it to be at least 24 (requires 128 MiB) and then always save a nice round three bytes.

- Reference start and reference index could be combined into a single value (concatenating all chromosomes and using a single position) as discussed in #227. Maybe 40 bits is enough for that? That corresponds to a maximum genome size of 1100 Gbp. This would save another three bytes.

- Also from #227: We could reduce the size of the hash. Is it realistic to reduce the size to 40 bits? We currently use 56 bits in total, 39 bits for the main hash and 17 for the secondary. It’s probably too many collisions if we go from 39 to 23 for the main hash.

- If we are desperate, we could omit the strobe 2 offset and re-compute from the reference after looking up the randstrobe. Hm, or maybe this could be done better: We could assume that the offset is the same as in the query, test whether that matches with the actual reference and only re-compute if it doesn’t match. Or maybe we could even use the fact that the the second syncmer is often the first syncmer in the next randstrobe. This needs more thought ...
 
One problem here is that the current data structure is nicely aligned on 64 bits. If we reduce the size from 16 to 12 bytes, for example, we need some extra code to ensure we don’t have unaligned memory access, which supposedly is slow – but maybe it’s not so bad. If we could somehow get down to 8 bytes, that problem disappears, but that is going to be hard. It would cut the index size into half, though ...

ksahlin:
Nice summary. 

It seems nearly "trivial" to go to 12 bytes if we do the reference start and reference index combination and take a few bits from the hash (secondary doesn't need many bits when conditioned on the main hash). I don't have good insight into the memory alignment overhead, though.

It also seems doable to go down to 8 bytes, because we have the extra B bits for hashes. Thus, a layout could be: 

what | bits
-|-:
hash | B+16 (B ~ 24-30)
ref idx+pos | 40
strobe-2-offset | 8

[...]

PR #607, opened 2026-06-30
Concatenate the reference contigs into one "flat" reference

### LINKED ISSUE #227: A potential way to save space for genomes smaller than 2^32nt.
Consider the following genome of three 20nt long chromosomes:

```
>chr1
AGACGCAGCGAGACGCAGCG
>chr2
TGCGTTAGTACGATTGCATC
>chr3
CCTTTAGTACGATTGCATCT
```

Instead of representing the seed location as e.g. (chr3, 10) for position 10 on chr3, we could concatenate the strings and represent it as position 50. The only time this needs to be 'converted back' to (chr3, 10) is for the sam alignments.  Such an offset vector could be implemented [0,20,40], storing the offset positions of each chromosome in the concatenated string. One can do binary search over the very small vector to find the right chr_id. The index in the offset vector is the index of our chromosome id (and we already have an index-> chromosome name vector.

How about our hash table (call it `H`)? Currently, `H` contains seed hash value as keys and two different values: 

If unique seed: stores reference chr_id and packed seed offset in 4 bytes and chromosome position in 4bytes.
If repetitive: store position in a vector with repetitive seeds (4 bytes), and the count of the seed (4bytes).

I think we can use only 4-bytes in `H` for both cases.

If unique seed: If one is willing to discard the packed strobe 2 position (taking 8 bits) we could go from 8 bytes down to 4 bytes. Here, the 4 bytes would simply be the position in the concatenated string. The strobe 2 position is, from what I remember, only used for checking if distance between the two strobes are the same on query and reference. If yes, first try hamming distance. otherwise go directly to alignment. We could instead always try hamming distance first as nearly no cost. 

If repetitive seed: We could store only the offset position in the vector in 4 bytes in `H`. The count could instead be fetched by having a (new) hash map `M : {hash_value -> hash_value_next}` which is obtained after sorting the vector on hash values. `M` should be fairly succinct as is only need to store one key-value pair for each repetitive seed. Getting the count (i.e., number of positions to iterate over in the vector) for a seed with the i:th smallest hash value `h_i`  would require four operations:

1. Obtain offset `o_i` of `h_i` from querying `H`
2. Obtain the next hash value `h_{i+1}` through`M`
3. Obtain offset `o_{i+1}`of `h_i+1` in `H`
4.  `count = o_{i+1} - o_i` 

The operations 1-3 of course comes at a cost in runtime compared to only have one lookup. I wonder how much if e.g. computation of prefetched data sa suggested in #203.

_Edit:  The packed strobe 2 position is also used for merging hits into 'merged hits' (called NAMs in the code). Two seed hits are merged if they overlap on both the query and the reference in the same order. To detect overlap, we use the length of the seed, where the position of the second strobe is needed. The extent to which the exact length of seeds is needed (or whether an approximate seed-length can be used) remains to be explored._

marcelm:
Instead of addressing a position on the genome with a (reference_id, reference_start) tuple, this uses a single 64-bit `ref_start` integer.

Advantages:
- Slightly simpler code in some places
- It appears to be slightly faster
- No more separate limit on the number of contigs and the length of each contig. Instead, the limit is on the total genome length (currently 64 bit).
- This should allow us to later reduce the number of bits needed for the reference coordinate.

Recovering the chromosome a position is on requires a binary search, but
this does not need to be done that often.

~~However, there are a couple of places in the code (currently marked with a `TODO` comment) where we iterate over anchors or chains from left to right and where we would previously do something when transitioning from one contig to another. Now, we can theoretically 1) chain from one contig to another and 2) pair one read with one on another contig. We’ll have to think a bit more about how much we need to prevent this.~~

The randstrobes generated in N regions change slightly because the pseudorandom nucleotide that the N wildcards are replaced with depends on the position, which is now relative to the start of the genome and not the start of the contig.

Closes #227
See #597

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,11 +10,15 @@
   Please report any other regressions!
 
   If you need one of the above features, use version 0.17.0.
-* #550: Introduce piecewise extension for single-end reads. The previous 
+* #550: Introduce piecewise extension alignment for single-end reads. The previous
   SSW extension can be enabled with `--ssw`.
 * #595: The in-memory representation of the reference sequence now uses two bits
   per nucleotide, reducing memory usage of the reference to 25% of its previous
   size. This change does not affect runtime.
+* #607: The reference is now stored as a single concatentade string of all
+  contigs. Reference coordinates now use a single 64-bit value. This means that
+  there are no longer any (relevant) limitations on the number and length of
+  contigs.
 * #539: Add a "noisy" read profile. Use `-P noisy` on the command line to
   select it. This is equivalent to `-k 16 -s 12 -l 2 -u 2 -m 100`.
   We found these settings to increase accuracy on error-prone reads at the cost
--- a/src/chain.rs
+++ b/src/chain.rs
@@ -19,12 +19,13 @@ use crate::shuffle::shuffle_best;
 #[derive(Clone, Debug, Default)]
 pub struct Chain {
     pub id: usize,
+    /// Start coordinate of the contig containing `ref_start`.
+    pub ref_contig_start: usize,
     pub ref_start: usize,
     pub ref_end: usize,
     pub query_start: usize,
     pub query_end: usize,
     pub matching_bases: usize,
-    pub ref_id: usize,
     pub score: f32,
     pub is_revcomp: bool,
     pub anchors: Vec<Anchor>,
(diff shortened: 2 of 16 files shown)


marcelm:
I have now fixed the remaining issues where alignments or pairs could cross contig boundaries. From my side, this is ready to be reviewed and merged.

I’ve done a full evaluation run and accuracy does not change by more than 0.01 percentage points. As alluded, there’s indeed a consistent speedup of a couple of percent.

ksahlin:
While one could do a deep dive into why accuracy changes before merging, the PR has been evaluated empirically on several datasets, and I approve the merge.

Do you have an idea of why there is a small difference in output locations?

marcelm:
> Do you have an idea of why there is a small difference in output locations?

These differences should come from the different assignment of pseudorandom nucleotides to `N` wildcards, which was based on the contig-relative coordinate, but is now based on the reference-relative coordinate.

PR #610, opened 2026-07-03
Make RefSequence a container for all contigs

marcelm:
This contains all of the purely refactoring commits of PR #607. That is, we can just merge this without having to wait for the other minor things that need to be done in that PR.

Instead of a single `PackedSeq`, the change is to store a `Vec<PackedSeq>` in `RefSequence`. So this means that what gets passed to a function that needs to know about the reference is no longer a `&[RefSequence]`, but a `&RefSequence`.

This lets us have a single data structure that stores the full genome sequence.

Also, this moves the `RefSequence` to a new `refseq` module (`refseq.rs`).

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -542,8 +542,9 @@ mod tests {
     use super::*;
 
     use crate::indexer::make_index;
-    use crate::io::fasta::{RefSequence, read_ref};
+    use crate::io::fasta::read_ref;
     use crate::packed_seq::PackedSeq;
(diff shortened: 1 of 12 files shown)


marcelm:
By the way, to avoid getting confused, I’m now trying to consistently use the term *reference* for the full genome and the term *contig* for the parts it is made of (a FASTA record, so a single chromosome or contig).

ksahlin:
Okay, approved!

PR #613, opened 2026-07-18
Pin the Rust toolchain and fix new clippy 1.97 lints

NicolasBuchin:
A fix to match our toolchain version and the CI's. It fixes the version in `rust-toolchain.toml`, and I set it to the latest 1.97 (currently in the CI).

I also went ahead and fixed main to match clippy 1.97 rules. Feel free to use an older version of the toolchain if you don't want the small changes in the code.

Since the clippy warnings are found in main, no commit will pass the CI until this is resolved.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,3 @@
+[toolchain]
+channel = "1.97.0"
+components = ["clippy", "rustfmt"]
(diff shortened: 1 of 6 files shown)


marcelm:
Nice, good improvement. The implication is that we from now on need to manually bump to newer Rust versions, but [this is the whole point](https://swatinem.de/blog/rust-toolchain/).

PR #614, opened 2026-07-30
Replace block-aligner with a custom SIMD aligner

### LINKED ISSUE #26: Base level alignment

NicolasBuchin:
Replace block-aligner with an in-house SIMD aligner in `src/simdaligner`, for both places the piecewise path used it: the global alignments between anchors and the extensions off the first and last anchor. Single-end only, since paired-end still extends with SSW.

I wanted this for two reasons. It is faster, and it is custom, so we can add the alignment types we need instead of working around what our dependency offers. `split_reference_alignment` is the first of those, aligning one query across two references with a single jump, which is what I will need to tell a tandem duplication from an insertion. strobealign calls three of the six modes the module exposes, `split_reference_alignment` and the two `local_reference_*` modes are unused for now.

This has 3000+ lines of code, written with Claude... I made sure to read ALL of them before making this PR. I don't claim that every line is perfect, or that the code is written like we would normally have written it, for example about 40% of it is documentation, and uses rust compile-time boolean const generics that we never use in the codebase before. I would treat the module more like a separate internal component than an ordinary extension of the existing codebase. That said, I am comfortable guaranteeing the high-level logic and design, and I have reviewed the implementation closely enough to be confident of it's correctness. The backend tests, end-to-end equivalence tests, and alignment level checks are also intended to put some guardrails around the parts that are easiest to get wrong. This is not a claim that the codebase is flawless, it is a deliberate tradeoff to get a custom, substantially faster aligner into strobealign while keeping the implementation self-contained and testable in a short amount of time.

## The kernel

Affine-gap alignment on the difference recurrence of Suzuki & Kasahara ([BMC Bioinformatics, 2018](https://link.springer.com/article/10.1186/s12859-018-2014-8)). Storing differences instead of absolute cell values bounds every stored value by the scoring scheme alone, with no length term, so 8-bit lanes hold at any sequence length and an AVX2 register holds 32 cells. The problem is that not every scoring scheme fits the bound.

It is written once and generic over a `Backend` of 21 operations, each a single intrinsic, so an instruction set costs a transcription rather than an aligner: AVX2 at 32 lanes, SSE4.1, NEON and wasm simd128 at 16, and a portable fallback for anything else. The pick happens at runtime on x86_64 and never panics for want of an instruction set. CI covers AVX2 and SSE4.1 on `ubuntu-latest` and NEON on `macos-latest`, but no wasm target is built, so treat simd128 as untested.

Vector code adds 93 `unsafe` blocks, all of them under `src/simdaligner`. 80 are in the backends, and 72 of those are a single intrinsic call, unsafe only because the intrinsic needs a target feature the CPU may not have. Of the remaining 8, five forward to the kernel and three build a movemask out of a shift and a reduction. That leaves 9 blocks in the shared kernel doing the pointer work, unchecked loads and stores into the DP and traceback buffers, and 4 in tests. `backend/tests.rs` checks all 21 operations of every backend and then checks the backends against each other end to end.

## What changes

Between anchors the call is unbanded and exact. block-aligner capped its block size (8192), so a long anchor gap could come back as a heuristic alignment and not an exact one. Among equally scoring alignments the CIGAR is the one minimising the edit distance, and it is built straight into strobealign's `Cigar`, so the three block-aligner translation helpers are gone.

Off the ends we now do a local start or end alignment, with no x-drop and no dynamic block sizes. We do use a heuristic, a bandwidth: confined to `w` diagonals either side of the anchor diagonal, optimal inside the band, and an alignment needing more than `w` bases of net indel out to the end of the read is unreachable. `--xdrop` is removed, `--bw` sets `w` and defaults to 1024. The band is there for runtime: an unbanded extension costs the read length times the reference extent, and is what makes very long reads with low chain coverage expensive (e.g. sim6 chrY at 20kb+ ).

The second commit checks `-A/-B/-O/-E/-L` at startup, because the kernel cannot represent every scoring scheme and previously found out by panicking after the whole index build. The third narrows that check to the piecewise path so `--ssw` does not reject schemes SSW handles fine. It is separable and droppable if we want.

`src/piecewisealigner.rs` loses 829 lines net, including the 23 block-aligner unit tests. `src/simdaligner` has 28 tests covering the six modes, re-scoring every returned CIGAR against its reported score and spans, plus the 11 backend-equivalence tests.

## Results

Accuracy vs runtime on 8 threads [ends.pdf](https://github.com/user-attachments/files/30547647/ends.pdf).

Table [ends-se-accuracy-table.pdf](https://github.com/user-attachments/files/30547649/ends-se-accuracy-table.pdf).

I evaluated single-end `align`, fruitfly/maize/chrY/CHM13 crossed with sim0/sim4/sim6, 50 to 30000 bp. Total wall clock over the sweep went from 2780 s to 882 s, a 3.15x. Replacing the global alignment alone takes it to 1207 s, the banded ends account for the rest. Per configuration the median speedup is 1.3-1.9x at 1 kb and below and 1.9-3.7x from 2.5 kb up, peaking at 3.7x at 20 kb. `map` mode is identical for all three tools, since nothing there aligns.

Replacing the global alignment changes nothing: accuracy is identical to main in all configurations, to the last digit reported, so that half is purely a runtime win. Only the end extensions change output. 52 of the 156 configurations move, 32 up and 20 down. The gains concentrate at 30 kb and go up to +0.40 points; every loss is at most 0.05 points. The fraction of reads aligned is unchanged everywhere.

Paired-end output is unchanged.

Closes #26

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -281,9 +281,9 @@ struct Args {
     #[arg(long = "ssw", help_heading = "Alignment")]
     use_ssw: bool,
 
-    /// X-drop threshold for piecewise extension
-    #[arg(long = "xdrop", default_value_t = 500, value_name = "N", help_heading = "Alignment")]
-    xdrop: i32,
+    /// Bandwidth of the piecewise start/end extensions, in diagonals
+    #[arg(long = "bw", default_value_t = 1024, value_name = "N", help_heading = "Alignment")]
+    bandwidth: usize,
 
 
     /// Path to input reference (in FASTA format)
@@ -559,7 +559,7 @@ fn run() -> Result<(), CliError> {
     debug!("{:?}", scores);
 
     let chainer = Chainer::new(index.k(), chaining_parameters);
-    let aligner = Aligner::new(scores, index.k(), args.xdrop);
+    let aligner = Aligner::new(scores, index.k(), args.bandwidth);
 
     let cmd_line = env::args().skip(1).collect::<Vec<_>>().join(" ");
     let rg_id = match args.rg_id {
--- a/src/main.rs
+++ b/src/main.rs
@@ -357,7 +357,10 @@ fn run() -> Result<(), CliError> {
         gap_extend: args.gap_extension_penalty,
         end_bonus: args.end_bonus,
     };
-    check_scores(scores)?;
+    // Only the piecewise path has a bound on the scheme, and --ssw does not take it.
+    if !args.use_ssw {
+        check_scores(scores)?;
+    }
 
     if cfg!(target_os = "linux") {
         warn_clocksource();
@@ -566,7 +569,11 @@ fn run() -> Result<(), CliError> {
     debug!("{:?}", scores);
 
     let chainer = Chainer::new(index.k(), chaining_parameters);
-    let aligner = Aligner::new(scores, index.k(), args.bandwidth);
+    let aligner = if args.use_ssw {
+        Aligner::new_ssw_only(scores)
+    } else {
+        Aligner::new(scores, index.k(), args.bandwidth)
+    };
 
     let cmd_line = env::args().skip(1).collect::<Vec<_>>().join(" ");
     let rg_id = match args.rg_id {
--- a/src/main.rs
+++ b/src/main.rs
@@ -40,6 +40,7 @@ use strobealign::mapper::{
 };
 use strobealign::mcsstrategy::McsStrategy;
 use strobealign::seeding::{DEFAULT_AUX_LEN, InvalidSeedingParameter, SeedingParameters};
+use strobealign::simdaligner::{InvalidScores, check_scores};
 
 mod logger;
 
@@ -318,6 +319,9 @@ enum CliError {
     #[error("No sequences found in the reference FASTA")]
     NoReference,
 
+    #[error("Invalid scoring scheme: {0}")]
+    InvalidScores(#[from] InvalidScores),
+
     #[error("When opening '{0}': {1}", .path, .e)]
     IoWithPath { e: io::Error, path: String },
 }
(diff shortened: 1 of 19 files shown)


src/main.rs
```diff
@@ -265,9 +266,9 @@ struct Args {
     #[arg(long = "ssw", help_heading = "Alignment")]
     use_ssw: bool,
 
-    /// X-drop threshold for piecewise extension
-    #[arg(long = "xdrop", default_value_t = 500, value_name = "N", help_heading = "Alignment")]
-    xdrop: i32,
+    /// Bandwidth of the piecewise start/end extensions, in diagonals
+    #[arg(long = "bw", default_value_t = 1024, value_name = "N", help_heading = "Alignment")]
```
marcelm, line 286: The changelog should mention that `--xdrop` has been removed and `--bw` added.

marcelm (commented):
I have reviewed the parts that I can review. I suggested earlier that we should maybe move this into a separate crate, but I no longer think that is necessary. Having multiple crates makes everything a bit more complicated. One argument was that the code can then be used in other projects, but strobealign itself is now on crates.io and can be added as a dependency by anyone who wants it, so this is already possible. So let’s keep it in this crate.

ksahlin (approved):

NicolasBuchin:
Updated the Changes.md to mention the new `--bw` parameter and removed `--xdrop`. Also rebased it on main.

PR #250, opened 2023-02-28
Introduce an end bonus to penalize soft-clipping

marcelm:
An alternative to #247 that is independent of the WFA2/ksw2 alignment restructuring in #242.

Closes #54

~~I’m not happy with how single-end accuracy (using the simpler "1/0" overlap metric) decreases at 50 bp (-0.1 percentage points, in bold in the table) and will need to investigate, but everything else looks ok or actually quite good.~~ This was a bug, see comment below.

**Tables updated on 2023-03-01 for commit ee34340**

## Single ends, Jaccard overlap

Average difference: 0.14565

dataset | main | this PR | diff
-|-|-|-|
drosophila/1x50-1M | 81.66 | 81.969 | 0.309
drosophila/1x100-1M | 88.882 | 89.064 | 0.182
drosophila/1x150-1M | 90.657 | 90.781 | 0.124
drosophila/1x200-1M | 91.743 | 91.837 | 0.094
drosophila/1x300-1M | 93.063 | 93.126 | 0.063
CHM13/1x50-1M | 80.771 | 81.092 | 0.321
CHM13/1x100-1M | 90.236 | 90.419 | 0.183
CHM13/1x150-1M | 92.118 | 92.244 | 0.126
CHM13/1x200-1M | 92.995 | 93.09 | 0.095
CHM13/1x300-1M | 93.988 | 94.051 | 0.063
rye/1x50-1M | 44.156 | 44.381 | 0.225
rye/1x100-1M | 68.961 | 69.13 | 0.169
rye/1x150-1M | 80.102 | 80.224 | 0.122
rye/1x200-1M | 85.579 | 85.669 | 0.09
rye/1x300-1M | 90.173 | 90.232 | 0.059
maize/1x50-1M | 46.91 | 47.144 | 0.234
maize/1x100-1M | 70.222 | 70.399 | 0.177
maize/1x150-1M | 80.936 | 81.059 | 0.123
maize/1x200-1M | 86.497 | 86.59 | 0.093
maize/1x300-1M | 91.706 | 91.767 | 0.061

## Paired ends, Jaccard overlap

Average difference: 0.1553

dataset | main | this PR | diff
-|-|-|-|
drosophila/2x50-1M | 89.603 | 89.927 | 0.324
drosophila/2x100-1M | 92.05 | 92.24 | 0.19
drosophila/2x150-1M | 92.961 | 93.084 | 0.123
drosophila/2x200-1M | 93.293 | 93.392 | 0.099
[...]

PR #615, opened 2026-07-31
Suggestion: Change -L description to bonus and not penality

NicolasBuchin:
`-L` is described as "Soft-clipping penalty", but it sets the end bonus, which is added to the score when the alignment reaches the end of the read. Nothing is subtracted for clipped bases, and the value does not scale with how many there are.

I propose we change the description to explicitly say it's a bonus, which is also how CHANGES.md describes it (PR #250).

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -257,7 +257,7 @@ struct Args {
     #[arg(short = 'E', default_value_t = Scores::default().gap_extend, value_name = "N", help_heading = "Alignment")]
     gap_extension_penalty: u8,
 
-    /// Soft-clipping penalty
+    /// Bonus added to the alignment score when no soft-clipping occurs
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -259,7 +259,7 @@ struct Args {
     #[arg(short = 'E', default_value_t = Scores::default().gap_extend, value_name = "N", help_heading = "Alignment")]
     gap_extension_penalty: u8,
 
-    /// Soft-clipping penalty
+    /// Bonus added to the alignment score when it reaches a read end
```
marcelm: The term "soft-clipping" should still be mentioned because the option tunes when the CIGAR soft-clipping operator appears. I suggest: "Bonus added to the alignment score when no soft-clipping occurs"

PR #616, opened 2026-07-31
Scale the chaining lookback with read length, and precompute its log2 penalty

NicolasBuchin:
Two independent changes in `chainer.rs`.

`max_lookback` (`-H`) is a number of anchors, default 50. The number of anchors on a read grows with the read length, so 50 is too small for all long reads. At 30 kb the correct previous anchor is often further back than 50 positions in the sorted array, so the DP never looks at it, and the chain either breaks or comes out biased towards insertions. `-H` is now a floor and the lookback scales above it, `max_lookback.max(read_len / 200)`, so nothing changes at or below 10 kb.

The second commit precomputes log2. `compute_score` calls `log2f` for the diagonal penalty on the innermost line of the DP. It is now a 1024 entry table built in `Chainer::new`, and we still call libm for `dd >= 1024`. The table holds exactly what the call returns so the output is bit identical. This helps cut a little bit of the runtime at no cost to the outputs.

## Results

The `max_lookback` change is the only one affecting accuracy, while runtime is affected my the log2 change.

Accuracy vs runtime: [ends.pdf](https://github.com/user-attachments/files/30585951/ends.pdf)

Accuracy table: [ends-se-accuracy-table.pdf](https://github.com/user-attachments/files/30585952/ends-se-accuracy-table.pdf)

I evaluated single-end only, as the accuracy will only change for long reads. Accuracy is identical to main in all 12 cells at every length from 50 bp to 10 kb, which is expected since the lookback only grows past 50 above 10 kb. It moves in 19 of the 156 cells, all of them at 20 and 30 kb, and every one of the 19 is an improvement.

This is more of a "fix" for long reads above 10kbp, I have not evaluated if we could use a fully fine-tuned dynamic look-back window based on read length for all datasets.

### DIFF
--- a/src/chainer.rs
+++ b/src/chainer.rs
@@ -57,19 +57,26 @@ pub struct Chainer {
     k: usize,
     parameters: ChainingParameters,
     precomputed_scores: [f32; N_PRECOMPUTED],
+    precomputed_log2: [f32; N_PRECOMPUTED],
 }
 
 impl Chainer {
     pub fn new(k: usize, parameters: ChainingParameters) -> Self {
+        let mut precomputed_log2 = [0f32; N_PRECOMPUTED];
+        #[allow(clippy::needless_range_loop)]
+        for d in 0..N_PRECOMPUTED {
+            precomputed_log2[d] = (d as f32 + 1f32).log2();
+        }
         let mut precomputed_scores = [0f32; N_PRECOMPUTED];
         #[allow(clippy::needless_range_loop)]
         for i in 0..N_PRECOMPUTED {
-            precomputed_scores[i] = compute_score(i, i, k, &parameters);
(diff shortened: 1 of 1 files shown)


NicolasBuchin:
Only the lookback change is going forward. It is now #620.

I dropped the log2 precompute. Its output is bit identical to main, so runtime was its whole justification, and I measured no gain and possibly a small loss. The table only pays off when `dq != dr`, since the `dq == dr` path is already precomputed and never called `log2f` in the first place. What it saves there is one `log2f` against an L1 load and a bounds check, and it adds a second 1024 entry table next to `precomputed_scores` for the DP to walk.

The runtime numbers above were measured on a laptop and I am not confident in them. The accuracy numbers are unaffected and carry over to #620.

PR #618, opened 2026-08-04
Single-end: Remove is_consistent and verify anchor k-mers before the piecewise extension

NicolasBuchin:
`is_consistent` checks the first and last k-mer of a chain against the reference and throws the whole chain away if either one is a hash collision. It comes from NAMs, where the start and the end were the only positions we had. A chain gives us every anchor, so the check now looks at two of them and discards what the rest say. When the collision is on one end and that chain was the only one we found, the read comes out unmapped even though the chain pointed at the right place.

`retain_verified_anchors` compares each anchor's k-mer base by base between query and reference and drops only the ones that do not match. It runs in `extend_seed` right before the extension, on the decoded reference, and after the other anchor pruning so it works on as few anchors as possible. If none of them survive we return no alignment for that chain and move on to the next one. The `is_consistent` gate now applies only when we extend with SSW.

This is a fix, and ideally we would catch these at seeding instead, which would save the N * k base comparisons that cleaning a chain costs here.

### DIFF
--- a/src/mapper.rs
+++ b/src/mapper.rs
@@ -21,7 +21,7 @@ use crate::io::sam::{
 };
 use crate::math::normal_pdf;
 use crate::mcsstrategy::McsStrategy;
-use crate::piecewisealigner::remove_spurious_anchors;
+use crate::piecewisealigner::{remove_spurious_anchors, retain_verified_anchors};
 use crate::read::Read;
 use crate::refseq::RefSequence;
 use crate::revcomp::reverse_complement;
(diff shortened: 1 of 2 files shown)


ksahlin:
Great, and we need a similar fix in paired-end mode.

NicolasBuchin:
Superseded by #619, which fixes this upstream in the chainer instead. Anchors are verified as they are built, so no collision reaches a chain and `is_consistent` goes away entirely. That also covers paired-end, as you asked.

PR #619, opened 2026-08-04
Verify anchor k-mers before chaining and remove is_consistent checks

NicolasBuchin:
In single-end reads, `is_consistent` checks the first and last k-mer of a chain against the reference and throws the whole chain away if either one is a hash collision. It comes from NAMs, where the start and the end were the only positions we had. When the collision is on one end and that chain was the only one we found, the read comes out unmapped even though the chain pointed at the right place.

We now check every anchor as it is built in `hits_to_anchors` and keep only the ones whose k-mer really matches, so a collision never becomes an anchor. No collisions means no more `is_consistent` check, and no more `reverse_chain_if_needed` either. Both are removed.

The reference is already 2-bit packed and we pack the query once per read, so checking a k-mer is one `u64` comparison, O(1) instead of O(k).

This fixes paired-end too.

## Results
Ran with 8 threads:
Runtime and accuracy plots: [ends.pdf](https://github.com/user-attachments/files/30710702/ends.pdf)
Accuracy table+: [ends-se-accuracy-table.pdf](https://github.com/user-attachments/files/30710706/ends-se-accuracy-table.pdf)

Accuracy is up in 64 of the 156 cells and down in 29, and the gains are the bigger ones, 0.019 points on average against 0.014.

I have not checked paired-end results.

### DIFF
--- a/src/chain.rs
+++ b/src/chain.rs
@@ -45,24 +45,6 @@ impl Chain {
             .saturating_sub(self.query_start)
             .max(self.ref_contig_start)
     }
-
-    /// Returns whether a chain represents a consistent match between read and
-    /// reference by comparing the nucleotide sequences of the first and last
-    /// strobe (taking orientation into account).
-    pub fn is_consistent(&self, read: &Read, refseq: &RefSequence, k: usize) -> bool {
-        let ref_start_kmer = refseq.decode(self.ref_start, self.ref_start + k);
-        let ref_end_kmer = refseq.decode(self.ref_end - k, self.ref_end);
-
-        let seq = if self.is_revcomp {
-            read.rc()
-        } else {
-            read.seq()
-        };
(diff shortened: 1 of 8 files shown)


marcelm:
I’ve rebased on top of main and added some code to show hash collision statistics. Not sure whether we actually want this in case we merge it, but it’s good to have temporarily.

NicolasBuchin:
I think it would be good to remove hash collisions, but this has a non negligible cost. Accuracy won't improve much, but the alignments with anchors from hash collisions are quite annoying, and a pruning check would also have a cost... Ideally we would be able to find a way to keep the full k-mers encoded on 2*k bits hash, so we don't have any collisions. I'll try and see but I don't expect it to be able to fit all the encoding. we use at most k=23 (profile read len at 400), so a full hit would be 46*2=92bits?

marcelm:
Closing in favor of #628.

PR #621, opened 2026-08-06
Trim a piecewise alignment to its best-scoring segment

NicolasBuchin:
In single-end reads, `extend_piecewise` builds one alignment across the whole chain and returns it only if the total score is positive. That costs us two kinds of reads: when the chain scores zero or less overall we return nothing and the read comes out unmapped, even though a stretch inside it aligns well. When it scores positive but the extension has run past the point where it still pays, the bad flank stays in the output.

We now scan the finished CIGAR and keep only its best-scoring run of operations. The rest becomes soft clip. It is Kadane's algorithm over the operations.

The segment collects the end bonus when it reaches the start or the end of the read. So a flank that costs a few mismatches is kept when reaching the end of the read pays for them, and trimmed when it does not.

## Results
Benchmarked with 16 threads:
[ends.pdf](https://github.com/user-attachments/files/30780339/ends.pdf)
[ends-se-accuracy-table.pdf](https://github.com/user-attachments/files/30780341/ends-se-accuracy-table.pdf)
[ends-se-aligned.pdf](https://github.com/user-attachments/files/30780364/ends-se-aligned.pdf)

I evaluated single-end only. Accuracy is up in 41 of the 156 cells and down in 20.

Paired-end output is unchanged, it still extends with SSW.

### DIFF
--- a/src/cigar.rs
+++ b/src/cigar.rs
@@ -1,4 +1,5 @@
 use std::fmt::{Debug, Display, Formatter};
+use std::ops::Range;
 use std::str::FromStr;
 
 #[derive(Debug, PartialEq, Eq, Clone, Copy)]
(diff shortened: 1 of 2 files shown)


marcelm:
I think this looks good. It’s a bit unfortunate our test data don’t really show a clearer improvement, so it’s a bit of a judgement call.

When I ran the evaluation, I thought that maybe it is helpful to look at the Jaccard accuracy, but this one actually goes down a little bit (-0.2 in the worst case, avg. -0.02), which doesn’t quite match the expectation that we would throw away fewer alignments. I’ll investigate a little bit.

PR #622, opened 2026-08-13
Add anchors from filtered hits to the chains after chaining

NicolasBuchin:
We drop hits whose hash is too frequent before chaining. It keeps runtime and memory low, but in a repetitive genome those are most of the seeds: on chrY up to ~40% of a read's hits never make it in. The chains we build there come out sparse or broken and the read lands in the wrong place, even though the hits that would have placed it right were found and then thrown away.

I propose to bring that information back after chaining: The filter stays, chaining runs as it did, and once it's done we know where the chains are, so we go back for the filtered hits and try and add them back into the chains.

A chain that gets anchors back covers more of the read and scores higher. In map only mode the top scoring chain is the output, this makes map only much more accurate, and the matching-bases count we report gets closer to what the read really covers, even when the region does not change. In alignment mode the score is what decides which chains we extend and in what order, and the chains that gain the most are the ones on the read's true region, this means we extend it earlier and are more likely to report the alignment on the right location. 

Extension is also cheaper on a fuller chain, shorter gaps to close between anchors and less to extend off the ends. It does not cover the cost though, chaining gets more expensive than extension gets cheaper, so we are paying runtime for accuracy.

## How we do anchor recovery

Recovery runs once per read, after both orientations are chained and the chains extracted. For each chain we take the filtered anchors near it, try and add them into the gaps between the chain's own anchors plus one gap at each end, for that we run a small DP per gap that keeps the best scoring run of them. A run only goes in if it beats the 2 anchors link it replaces. A run of anchors use the same gates as `collinear_chaining`, `max_ref_gap` and `max_diagonal_ratio` and strictly increasing in query and reference, so we can't add a set of anchors the chainer would have refused. We only add filtered hits but we do not remove or replace the original anchors of a chain. After recovery, we recompute the score, matching bases and endpoints for each chain if it had recovered anchors added to it.

We only keep hits that can give an anchor near a chain's diagonal. For a given hit we know where such an anchor would have to sit on the reference, so we ask the index for those ranges instead of pulling in every entry the hit points at.

All the chains get recovery, and the best chains post-recovery are not in the same order as post-chaining. So feeding more chains to recovery helps accuracy, especially in mapping only. I lowered the score threshold to `0.5 * best_score` and set a limit of 200 chains per orientation.

## Results

These are the results ran on 16 threads:
[ends.pdf](https://github.com/user-attachments/files/31068840/ends.pdf)
[ends-se-accuracy-table.pdf](https://github.com/user-attachments/files/31068842/ends-se-accuracy-table.pdf)
[ends-se-memory.pdf](https://github.com/user-attachments/files/31068848/ends-se-memory.pdf)
(These results are a bit outdated now, since new things have been merged into main)

I would prefer until PR #614 to be merged before considering this idea, and to do a full benchmark again once it's done. Not only would it help cut some of the runtime that we add here, but it also fixes problems with big gaps between anchors.

Recovery has a lot of moving pieces and I tried to adjust them as best I could, but it's quite a big change, and it means that we now have 1 more step in our pipeline that can be affected by future changes: seeding -> chaining -> recovery -> extension. I think the results are great and the fact that we seem to be very close, if not better than minimap2 on most datasets make it worth while.


marcelm:
I’ve already started benchmarks for aa88665, but you force-pushed just now, should I re-start the benchmarks?

marcelm:
Hm, ok, I can see that there is a code change, will restart.

NicolasBuchin:
The CI said no, I checked why it failed, and then I refactored to make more sens.

Issue #598, opened 2026-06-10
Almost perfect secondary matches not found


tprodanov:
Good afternoon,

I encountered a problem (present in the latest commit) where almost perfect secondary alignments don't get reported no matter the `-S` and `-M` values.

Specifically, using the following flies:
[haplotypes.fa.gz](https://github.com/user-attachments/files/28795101/haplotypes.fa.gz)
[read.fq.gz](https://github.com/user-attachments/files/28795122/read.fq.gz)
```sh
strobealign --no-progress -M 10000 -N 10000 -S 0.1 -f 0.0001 -k 15 --eqx -r 150 -t 8 \
    db/loci/PARVB/haplotypes.fa.gz read.fq.gz
```
produces 116 alignments but I know for sure that there should be more. For example, the given read aligns to haplotypes HG02451.1 or NA19240.1 with just a single mismatch, so I hoped they would be reported. I tried different values of `-M`, `-N`, `-S` and `-f` but could not make it work. Could you please look into it and perhaps provide a flag to disable whatever heuristics that stops the read alignment when a perfect alignment is already found?

Best wishes,
Timofey

ksahlin:
Hi Timofey,

Thanks for the detailed report and for providing the test files.

Is there a clean definition of "enough alignments" in a haplotype-aware context like yours, where many sequences are nearly identical by design? The current heuristics were tuned for reference genome mapping, not multi-haplotype panels.

What is the edit distance / alignment score of the 116 alignments that are reported -- are they all perfect?

On the question of a flag to disable early stopping: this is something we can consider, but it requires some thought on what would then be the stopping criterion - edit distance, alignment score dropoff to the best match, or similar? We want to make sure such a flag is clearly scoped, e.g. "report all alignments within -S of the best" rather than unbounded.

tprodanov:
Thank you for looking into this!

I understand that my use-case is rather unusual, so no worries if this is hard to solve. I am writing some code right now that would transfer alignments from one haplotype to another by combining CIGARs, but my worry is that I may still miss some alignments so it would be more reliable if Strobealign could generate them as well.

[...]

PR #624, opened 2026-08-24
Use a relative score threshold for secondary alignment output

Itolstoganov:
This changes the filter for secondary alignment outputs. Currently, the filter is hardcoded as

```
if alignment.score.saturating_sub(best_score)
    > 2 * aligner.scores.mismatch as u32 + aligner.scores.gap_open as u32
```
This seems bugged, since `alignment.score` never exceeds `best_score`. 

The hardcoded threshold was replaced with user-defined `secondary_threshold` (0.95 by default), i.e. the score should satisfy

`alignment.score >= secondary_threshold * best_score`. User-defined threshold was requested in [#598 ](https://github.com/ksahlin/strobealign/issues/598)

Additionally, the `-N` upper bound for the number of secondary alignments can be turned off by using a negative value.

Sanity check results (accuracy should be the same, since secondary alignments are not used in the strobealign-evaluation benchmark)
[ends.pdf](https://github.com/user-attachments/files/31373862/ends.pdf)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -143,6 +143,15 @@ struct Args {
     #[arg(short = 'N', default_value_t = 0, value_name = "N", help_heading = "SAM output")]
     max_secondary: usize,
 
+    /// Output secondary alignments with alignment score at least st of primary alignment score
+    #[arg(
+        long = "st",
+        default_value_t = MappingParameters::default().secondary_threshold,
+        value_name = "SECONDARY_THRESHOLD",
+        help_heading = "SAM output"
+    )]
+    secondary_threshold: f32,
+
     // Seeding arguments
 
     /// Mean read length. Default: estimated from the first 500 records in the input file
@@ -499,6 +508,7 @@ fn run() -> Result<(), CliError> {
     let timer = Instant::now();
     let mapping_parameters = MappingParameters {
         max_secondary: args.max_secondary,
+        secondary_threshold: args.secondary_threshold,
         max_tries: args.max_tries,
         dropoff_threshold: args.dropoff_threshold,
         rescue_distance: args.rescue_distance,
--- a/src/main.rs
+++ b/src/main.rs
@@ -149,11 +149,11 @@ struct Args {
     )]
     max_secondary: i64,
 
-    /// Output secondary alignments with alignment score at least st of primary alignment score
+    /// Secondary threshold. Output only secondary alignments with score at least FRACTION of primary alignment score
     #[arg(
         long = "st",
         default_value_t = MappingParameters::default().secondary_threshold,
-        value_name = "SECONDARY_THRESHOLD",
+        value_name = "FRACTION",
         help_heading = "SAM output"
     )]
     secondary_threshold: f32,
@@ -513,7 +513,7 @@ fn run() -> Result<(), CliError> {
 
     let timer = Instant::now();
     let mapping_parameters = MappingParameters {
-        // Replaced with usize::MAX to turn off the max_secondary threshold
+        // Negative -N argument disables limit
         max_secondary: if args.max_secondary < 0 {
             usize::MAX
         } else {
--- a/src/main.rs
+++ b/src/main.rs
@@ -139,9 +139,15 @@ struct Args {
     #[arg(short = 'C', help_heading = "SAM output")]
     fastq_comments: bool,
 
-    /// Retain at most N secondary alignments (is upper bounded by -M and depends on -S)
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -139,9 +139,24 @@ struct Args {
...
+        allow_negative_numbers = true,
+        help_heading = "SAM output"
+    )]
+    max_secondary: i64,
+
+    /// Output secondary alignments with alignment score at least st of primary alignment score
+    #[arg(
+        long = "st",
+        default_value_t = MappingParameters::default().secondary_threshold,
+        value_name = "SECONDARY_THRESHOLD",
```
marcelm: Written like this, one may misunderstand that setting this enables outputting secondary alignments. Maybe like this:
```suggestion
    /// Secondary threshold. Output only secondary alignments with score at least FRACTION of primary alignment score
    #[arg(
        long = "st",
        default_value_t = MappingParameters::default().secondary_threshold,
        value_name = "FRACTION",
```

src/main.rs
```diff
@@ -498,7 +513,13 @@ fn run() -> Result<(), CliError> {
 
     let timer = Instant::now();
     let mapping_parameters = MappingParameters {
-        max_secondary: args.max_secondary,
+        // Replaced with usize::MAX to turn off the max_secondary threshold
```
marcelm: I was confused by this comment for a moment because I thought it told me something important, but then I realized it just re-states what the code does. I think I’d just remove it or make it shorter
```suggestion
        // Negative -N argument disables limit
```

src/mapper.rs
```diff
@@ -1425,11 +1425,16 @@ fn aligned_pairs_to_sam(
     } else {
         let mut alignment_status = AlignmentStatus::Primary;
         let s_max = best_aln_pair.score;
-        for aln_pair in high_scores.iter().take(max_secondary) {
+        // Expressed as a maximum allowed drop from the best score rather than as
+        // `s_score >= secondary_threshold * s_max`, since `s_max` can be negative.
+        let max_drop = ((1.0 - secondary_threshold) * s_max.abs()).max(0.0);
```
marcelm: I don’t quite understand this.
- shouldn’t `1.0 - secondary_threshold` always be nonnegative? Then multiplying it with `s_max.abs()` should also always be nonnegative and then the `max(0.0)` isn’t necessary. I guess this guards against someone using a secondary_threshold that is larger than 1.0, but I think this should better be caught when parsing the command-line arguments and give an error.
- IMO, when `s_max` is already negative, it would make sense to just not output any secondary alignments at all. That is, the alternative expression with `s_score >= secondary_threshold * s_max` is IMO fine to use. (But I don’t know under which circumstances the best score is negative.)

marcelm (commented):
Good catch, this never worked it seems (it was already like this in C++, so not an issue from the Rust port).

ksahlin (approved):

src/mapper.rs
Itolstoganov: I agree with the first point; added a guard in main.rs instead (it seems a bit out of place, though, since there are no other checks for the parameters from the same group)

Itolstoganov: > IMO, when s_max is already negative, it would make sense to just not output any secondary alignments at all. That is, the alternative expression with s_score >= secondary_threshold * s_max is IMO fine to use. (But I don’t know under which circumstances the best score is negative.)

`s_max` can be negative if the sum of mate alignment scores is less than the log-likelihood of the predicted insert size (or less than 20 if the pair orientation is not proper):

https://github.com/ksahlin/strobealign/blob/98eb6301291ecfa6593cb4ae03db132597d139d2/src/mapper.rs#L992-L997

This probably never happens in practice, since a single read alignment score should be at least 2 * (seed length). But I do not see why we should treat the positive-score pairs differently from the negative-score ones.

src/main.rs
```diff
@@ -333,6 +348,14 @@ fn run() -> Result<(), CliError> {
     }
     debug!("RUSTFLAGS='{}'", RUSTFLAGS.unwrap_or(""));
 
+    if !(0.0..=1.0).contains(&args.secondary_threshold) {
+        error!(
+            "Secondary threshold (--st) must be between 0 and 1, got {}",
+            args.secondary_threshold
+        );
+        exit(1);
```
marcelm: The exit code should be 2 for errors in how command-line parameters are given for consistency with how clap handles it. This was incorrect in two other places in the code. I’ll push a commit to fix that.

src/mapper.rs
marcelm: Ah alright, then let’s leave it like it is.

PR #628, opened 2026-09-03
Remove invalid anchors after chaining

marcelm:
This is a modified version of PR #619 and an alternative to it.

This removes anchors for which the reference sequence and the query sequence don’t match from each chain before alignment. These invalid anchors come from hash collisions.

This has no measurable performance impact.

The alternative (PR #619) is to discard invalid anchors *before* chaining, which can be slightly more accurate (because the optimal chain may change with removed anchors), but incurs a significant slowdown.

Invalid anchors are currently not removed in mapping-only mode.

I’ll post benchmark results later.

### DIFF
--- a/src/mapper.rs
+++ b/src/mapper.rs
@@ -406,9 +406,14 @@ pub fn align_single_end_read(
         {
             break;
         }
-        let Some(alignment) =
-            extend_seed(aligner, chain, refseq, &read, mapping_parameters.use_ssw)
-        else {
+        let Some(alignment) = extend_seed(
+            aligner,
+            chain,
+            refseq,
+            &read,
+            mapping_parameters.use_ssw,
+            index.k(),
+        ) else {
             continue;
         };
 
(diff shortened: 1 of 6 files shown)


src/mapper.rs
```diff
@@ -595,11 +601,21 @@ fn extend_seed(
             let adjusted_anchors: Vec<Anchor> = chain
                 .anchors
                 .iter()
-                .map(|a| Anchor {
-                    ref_start: a.ref_start - decode_start - refseq.starts.0[contig_id],
-                    query_start: a.query_start,
+                .filter_map(|a| {
+                    let ref_start = a.ref_start - decode_start - refseq.starts.0[contig_id];
+                    // Skip anchors resulting from hash collisions
+                    if query[a.query_start..a.query_start + k] == segment[ref_start..ref_start + k]
```
NicolasBuchin: Not sure if it would be a visible difference in runtime, but instead of comparing two &[u8], we could retrieve the encoded k-mers directly in u64 from the encoded sequences, then it should be a lesser cost to compare them. It would also mean any k size will result in the same comparison cost. I had done this in the pre-chaining anchor verification.

ksahlin (approved):

PR #633, opened 2026-09-11
Some refactoring

marcelm:
This is purely refactoring without any (intentional) functional changes, with one exception: In abundance estimation mode, we previously had all-zero statistics, which is now fixed.

The two major changes are:
- Chaining and sorting the chains was done within the various alignment/mapping/abundance functions, but this is actually the same for all of them, so we can do this before and pass the sorted chains to the functions. This allows us to remove quite a few parameters from these functions and some duplicated code. (This was suggested by Anton.)
- The code that implements the various modes (extension alignment, mapping-only and abundance estimation) was moved to a separate `modes` module with one submodule for each mode. The name `mapper` for the module that implements extension alignment always bothered me (and calling it `aligner` also doesn’t work because we use that for base-level alignment), so moving everything to a `modes` module resolves that.

@NicolasBuchin, can you have a look at b6653f92493d47c3477030576a0e617d94aa99e7? This removes the `split_chains_by_orientation_checked` function because the `split_chains_by_orientation` function already handles the special case fine that all chains are forward or all revcomp. I’m not sure whether there was something else to it or whether this was just a bit of overly defensive programming.


marcelm:
Ah, so there was a reason. I have done some quick measurements now and this seems not worth it. For the dataset I tested, we would avoid counting 16% of the chains. If this were linear in the number of anchors, then I would get it, but there are only a handful of chains per read on average, which is dwarfed by everything else.

PR #634, opened 2026-09-15
Refactor alternative arguments

Alfonsson:
This is meant to only refactor without any functional changes. The goals are two: 

1. To allow the hits to be found per strobemer and not just for a full a query.
2. To allow for finding chains using a hits-vector already computed for a query.


src/hit.rs
```diff
@@ -198,32 +165,78 @@ fn find_all_hits(
...
+        let hit = Hit {
+            position: undirected_entry.position,
+            query_start: randstrobe.start,
+            query_end: randstrobe.start + index.k(),
+            is_partial: true,
+            is_filtered,
+            hash: randstrobe.hash,
+            hash_revcomp: randstrobe.hash_revcomp,
+        };
+        return Some(hit);
```
marcelm, line 233:
```suggestion
        
        Some(hit)
```

src/hit.rs
```diff
@@ -198,32 +165,78 @@ fn find_all_hits(
...
+    } else {
+        hits_details.partial_not_found += 1;
+        return None;
```
marcelm, line 236:
```suggestion
        
        None
```

src/chainer.rs
```diff
@@ -213,8 +213,24 @@ impl Chainer {
...
             );
         }
-        let mut time_find_hits = hits_timer.elapsed().as_secs_f64();
+        let time_find_hits = hits_timer.elapsed().as_secs_f64();
 
+        let (mut chain_details, chains) =
+            self.get_chains_from_hits(index, read_len, hits, hits_details);
+
+        chain_details.n_anchors = query_randstrobes[0].len() + query_randstrobes[1].len();
+        chain_details.time_find_hits += time_find_hits; //Anton: Why is this part of chain_details instead of hit details?
```
marcelm, line 222: Please remove the comment. Questions like that should be handled outside the code (message on Slack, GitHub issue, etc.).

marcelm (commented):
Looks good to me; just some minor changes. This can be merged when you re-open it (or submit a new PR).

PR #636, opened 2026-09-16
Reduce memory usage and speed up indexing for repetitive genomes

marcelm:
TL; DR: We save 1 GB RAM when indexing rye.

Instead of keeping one possibly huge `strobemer_counts` vector with one entry for each unique strobemer that has a count of two or larger, this now uses only one 1000-element long vector that keeps directly track of the frequencies of strobemers (up to 1000).

Rye has about ~110 million strobemers that have a count of two or larger, so this saves us allocating roughly 1 GB (8 bytes for entry). For CHM13, there are only ~12 million strobemers, so we save much less (~100 MB).

The runtime improves also quite noticably on rye, where indexing time goes from 90 s to 85 s.
(This may only be on my system where RAM is quite full when indexing rye.)

Plots: [ends-se-memory.pdf](https://github.com/user-attachments/files/32294306/ends-se-memory.pdf)


src/indexer.rs
```diff
@@ -89,6 +87,11 @@ pub fn make_index(
     if !randstrobes.is_empty() {
         randstrobe_start_indices.push(0);
     }
+
+    // strobemer_counts[i] is how many strobemers occur i times. Except:
+    // - `strobemer_counts[1]` is always zero
```
ksahlin: comment says it's 0 but below there is: "strobemer_counts[1] = unique_mers;"

ksahlin (commented):
Co

src/indexer.rs
marcelm: Fixed

ksahlin (approved):

